PR #1, opened 2018-02-01
Fix incompatibility with fair logger

awegrzyn:


Barthelemy:
testing it

awegrzyn:
@dberzano `ALICE Builder` did not launch :(

Barthelemy:
It will take a few hours, it is rebuilding the full stack on my machine

PR #2, opened 2018-02-07
Fix Configuration and Monitoring incompatibilities

awegrzyn:


awegrzyn:
To be clarified before merging:
• Which backend should be used
• Whether process monitoring needs to be enabled

PR #9, opened 2018-04-19
Configuration lowercase namespace

awegrzyn:
Compatible with Configuration `v1.4.0`


awegrzyn:
@Barthelemy is it okay with you?

PR #10, opened 2018-04-26
Fix build on GCC 7.3.1.

teo:


Barthelemy:
@knopers8 Any idea why the test fails ?

awegrzyn:
@Barthelemy Is it OK to merge? This also fixes build on CC7 devtoolset-7 gcc 7.2.1

Barthelemy:
sorry forgot about that one

PR #15, opened 2018-05-04
Fix instantiating empty framework::OutputSpec

knopers8:
The build was failing because default constructor of OutputSpec was recently deleted in O2.


knopers8:
The check shows that PR fixes failing build, merging.

PR #41, opened 2018-07-25
Rename TaskDPL

Barthelemy:
As it is not a class it should not have a capital first letter.


Barthelemy:
@knopers8 what do you think ?

knopers8:
@Barthelemy Sure, I didn't realise before that there is such a convention.

knopers8:
But change the cmake file

knopers8:
And update the link in README.md line 283

PR #47, opened 2018-08-21
Add arrow dependency

Barthelemy:
Needed because O2 has a reference to it in a header.


Barthelemy:
wait for https://github.com/alisw/alidist/pull/1254

PR #49, opened 2018-08-24
fix for mac (error loading lib)

Barthelemy:
Also a bit of cleanup in the code

### DIFF
--- a/Framework/include/QualityControl/CheckerDataProcessor.h
+++ b/Framework/include/QualityControl/CheckerDataProcessor.h
@@ -35,71 +35,94 @@ namespace checker {
 /// TODO Evaluate whether we should have a dedicated device to store in the database.
 ///
 /// \author Barthélémy von Haller
-class CheckerDataProcessor : public framework::Task {
-public:
-  /// Constructor
-  CheckerDataProcessor(std::string checkerName, std::string taskName, std::string configurationSource);
-  /// Destructor
-  ~CheckerDataProcessor() override;
+class CheckerDataProcessor : public framework::Task
+{
+  public:
+    /// Constructor
+    CheckerDataProcessor(std::string checkerName, std::string taskName, std::string configurationSource);
 
-  /// \brief Checker init callback
-  void init(framework::InitContext& ctx) override;
-  /// \brief Checker process callback
-  void run(framework::ProcessingContext& ctx) override;
+    /// Destructor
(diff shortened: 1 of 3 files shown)


Barthelemy:
I should have kept the code refactoring apart but I guess it is ok to have it in this PR.

Framework/include/QualityControl/CheckerDataProcessor.h
```diff
@@ -35,71 +35,94 @@ namespace checker {
 /// TODO Evaluate whether we should have a dedicated device to store in the database.
 ///
 /// \author Barthélémy von Haller
-class CheckerDataProcessor : public framework::Task {
```
knopers8: There is actually discrepancy between clang-format in O2 and QualityControl (i think QC is closer to Coding Guidelines). Maybe you could mention that in WP3?

Framework/src/CheckerDataProcessor.cxx
```diff
@@ -31,6 +31,7 @@ using namespace repository;
 namespace checker {
 
 // TODO do we need a CheckFactory ? here it is embedded in the Checker
+// TODO maybe we could use the CheckerDataProcessorFactory
```
knopers8: I am not sure if I understand - `CheckerDataProcessorFactory` for both checkers and checks?

Framework/src/CheckerDataProcessor.cxx
```diff
@@ -62,7 +63,9 @@ void CheckerDataProcessor::init(framework::InitContext&)
   // configuration
   try {
     std::unique_ptr<ConfigurationInterface> config = ConfigurationFactory::getConfiguration(mConfigurationSource);
-    populateConfig(config);
+    // configuration of the database
+    mDatabase = DatabaseFactory::create(config->get<std::string>("qc/config/database/implementation").value());
+    mDatabase->connect(config);
```
knopers8: Thank you, it has more sense now.

Framework/include/QualityControl/CheckerDataProcessor.h
Barthelemy: I will, thank you.

Framework/src/CheckerDataProcessor.cxx
Barthelemy: yes, with a rename :)

PR #51, opened 2018-09-05
Add helper script for generation of new QC modules

knopers8:
[QC-75]


Modules/modulesHelper.sh
```diff
@@ -0,0 +1,135 @@
+#!/usr/bin/env bash
+set -e ;# exit on error
+set -u ;# exit when using undeclared variable
+#set -x ;# debugging
+
+DONOR=SkeletonDPL
+DONOR_LC=skeleton_dpl
+DONOR_TASK=SkeletonTaskDPL
+DONOR_TASK_INCLUDE_GUARD=QC_MODULE_EXAMPLE_EXAMPLETASKDPL_H
```
Barthelemy: DONOR_TASK_INCLUDE_GUARD could be generated automatically with the DONOR_TASK : `QC_MODULE_${DONOR}_${DONOR_TASK}_H`

Barthelemy: And of course, the guard should be updated in SkeletonTaskDPL.h to match this... (and it will make more sense).

Modules/modulesHelper.sh
```diff
@@ -0,0 +1,135 @@
...
+DONOR_CHECK=SkeletonCheckDPL
+DONOR_CHECK_INCLUDE_GUARD=QUALITYCONTROL_LIBS_CHECKER_SkeletonCheckDPL_H_
```
Barthelemy: Same remark as above

Modules/modulesHelper.sh
```diff
@@ -0,0 +1,135 @@
...
+
+Options:
+ -h               print this message
+ -m=MODULE_NAME   create module named MODULE_NAME or add there some task/checker
+ -t=TASK_NAME     create task named TASK_NAME
+ -c=CHECK_NAME    create check named CHECK_NAME
+"
+}
+
+while getopts 'hm:t:c:' option; do
```
Barthelemy: It should handle the case when no parameters is passed (and show the help in such case)

Modules/modulesHelper.sh
```diff
@@ -0,0 +1,135 @@
...
+
+Example:
+# create new module and some task
+sh modulesHelper -m MyModule -t SuperTask
+# add one task and two checks
+sh modulesHelper -m MyModule -t EvenBetterTask -c HistoUniformityCheck -c MeanTest
+
+Options:
+ -h               print this message
+ -m=MODULE_NAME   create module named MODULE_NAME or add there some task/checker
```
Barthelemy: Above there is no equal sign and here there is. I think that it should be removed.

Modules/modulesHelper.sh
```diff
@@ -0,0 +1,135 @@
...
+function print_usage {
+  echo "Usage: sh modulesHelper -m MODULE_NAME [OPTION]
+
+Generate template QC module and/or tasks, checks.
+If a module with specified name already exists, new tasks and checks are inserted to the existing one.
+Please follow UpperCamelCase convention for modules', tasks' and checks' names. 
+
+Example:
+# create new module and some task
+sh modulesHelper -m MyModule -t SuperTask
```
Barthelemy: missing `.sh` in the name of the script

Modules/modulesHelper.sh
```diff
@@ -0,0 +1,135 @@
...
+# add one task and two checks
+sh modulesHelper -m MyModule -t EvenBetterTask -c HistoUniformityCheck -c MeanTest
```
Barthelemy: same

knopers8:
Thanks for the feedback, I will apply the changes!

Modules/modulesHelper.sh
knopers8: I can see that include guards could be fixed also in other modules by the occasion, so I will do it in a separate PR.

Modules/modulesHelper.sh
knopers8: Is the convention `QC_MODULE_${DONOR}_${DONOR_CHECK}_H` fine with you? It resembles more the file structure.

Modules/modulesHelper.sh
knopers8: Actually, it needs to be ```QC_MODULE_`echo ${DONOR} | tr a-z A-Z`_`echo ${DONOR_TASK} | tr a-z A-Z`_H``` if we want to preserve uppercase convention

knopers8:
I introduced all the changes you required. I used `git update-index --chmod=+x modulesHelper.sh` to change the permission, but I am not sure if that worked

Modules/modulesHelper.sh
Barthelemy: > I can see that include guards could be fixed also in other modules by the occasion, so I will do it in a separate PR.

Good, thank you.

Barthelemy: > Actually, it needs to be `` QC_MODULE_`echo ${DONOR} | tr a-z A-Z`_`echo ${DONOR_TASK} | tr a-z A-Z`_H `` if we want to preserve uppercase convention

Correct

Barthelemy:
> I introduced all the changes you required. I used `git update-index --chmod=+x modulesHelper.sh` to change the permission, but I am not sure if that worked

I will test it.

Modules/modulesHelper.sh
Barthelemy: yes, it is.

Barthelemy:
Please let me know when I can test it on Mac (i.e. when sed is fixed for mac).

knopers8:
@Barthelemy It should work now, though I didn't manage to come up with exactly the same command that would work both on BSD and GNU sed.

Modules/CMakeLists.txt
```diff
@@ -82,3 +82,5 @@ add_subdirectory(Skeleton)
 add_subdirectory(Daq)
 #add_subdirectory(ITS)
 add_subdirectory(SkeletonDPL)
+add_subdirectory(MyModule)
+add_subdirectory(Barth)
```
knopers8: We probably don't want these two in here

Barthelemy: indeed, sorry, I messed up

PR #53, opened 2018-09-11
Possibility to reset task's monitor objects after publishing

knopers8:
This will provide possibility to use mergers as "keepers" of full monitor objects, while distributed tasks would send only the updates. Of course, when there is no need for merging, tasks would keep the full MOs.


Barthelemy:
Nothing to say, good

PR #54, opened 2018-09-12
[WIP] Histogram merger for development purposes

knopers8:


Framework/include/QualityControl/HistoMerger.h
```diff
@@ -0,0 +1,65 @@
...
+#include "Common/Timer.h"
+
+#include "QualityControl/MonitorObject.h"
+
+namespace o2 {
+namespace quality_control {
+namespace core {
+
+
+/// \brief A crude histogram merger for development purposes.
```
Barthelemy: Maybe just add the limitations. Something like "It handles a single histogram and it expects to receive updates as opposed to full objects".

Framework/include/QualityControl/HistoMerger.h
```diff
@@ -0,0 +1,65 @@
...
+  framework::OutputSpec getOutputSpec() { return mOutputSpec; };
+
+  void setPublicationRate(int us);
+
+private:
+  // General state
+  std::string mMergerName;
+  std::shared_ptr<MonitorObject> mMonitorObject;
+  AliceO2::Common::Timer mPublicationTimer;
+  int mPublicationRate;
```
Barthelemy: Do you need this variable at all ? You could just get a parameter in the constructor and use it to initialize `mPublicationTimer`.

Barthelemy: Also it is not a rate, right ?

Framework/include/QualityControl/HistoMerger.h
```diff
@@ -0,0 +1,65 @@
...
+  /// Destructor
+  ~HistoMerger() override;
+
+  /// \brief Checker init callback
+  void init(framework::InitContext& ctx) override;
+
+  /// \brief Checker process callback
+  void run(framework::ProcessingContext& ctx) override;
+
+  void configureEdges(
```
Barthelemy: I am not sure about this name. Not very important though. "configureInputsOutputs" ? "configureTransputs" ? The latter is a new word in my vocabulary :)

Framework/include/QualityControl/HistoMerger.h
knopers8: Aah right, it is more like 'period' or 'interval'. But yes, making it a constructor parameter should be enough.

Framework/src/HistoMerger.cxx
```diff
@@ -0,0 +1,74 @@
...
+
+void HistoMerger::run(framework::ProcessingContext& ctx)
+{
+  for (const auto& input : ctx.inputs()) {
+    if (input.header != nullptr && input.spec != nullptr) {
+
+      if (!mMonitorObject) {
+        mMonitorObject.reset(DataRefUtils::as<MonitorObject>(input).release());
+      }
+      else {
```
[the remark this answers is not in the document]
knopers8: .clang-format in QC repo does impose that, I will do it manually

Framework/include/QualityControl/HistoMerger.h
knopers8: 'Edges' because of the graph theory. I am fine with `configureIO` or `configureInputsOutputs` though

PR #61, opened 2018-10-01
All libraries shall go to lib64

Barthelemy:
Some were in lib and some were in lib64 and some were... nowhere !


Barthelemy:
@awegrzyn for your information. Something to check in Monitoring as well.

awegrzyn:
Why to `lib64` and to `lib`?

Barthelemy:
because ${CMAKE_INSTALL_LIBDIR} defaults to lib64 (for me)

Barthelemy:
actually, I do not hard-code it anywhere any more, I just use this variable

awegrzyn:
So on Mac it's `lib` on CC7 `lib64` but eg. on SLC6 in my case it's not set at all... I think it's another point to investigate and decide what we shall use

PR #67, opened 2018-10-05
Remove unused classes (QC-83)

Barthelemy:


knopers8:
I cannot see any other reference to `SpyMainFrame`,  can it just be a problem with cached CMake files on the buildmachine?

knopers8:
OK, the tests passed this time, merging

PR #69, opened 2018-10-08
Remove old classes and rename DPL ones (QC-93)

Barthelemy:


Framework/include/QualityControl/Checker.h
```diff
@@ -1,31 +1,25 @@
 ///
 /// \file   Checker.h
```
knopers8: There is a little bit of inconsistency with naming of new `Checker` and `TaskDataProcessor`. They both share `framework::Task` interface and have associated factory, but have differerent naming convention. Maybe we should stick either to `Checker`/`Task`/`HistoMerger` or accordingly `*DataProcessor`.

Framework/include/QualityControl/Checker.h
```diff
@@ -1,31 +1,25 @@
...
 /// \author Barthelemy von Haller
+/// \author Piotr Konopka
 ///
 
-#ifndef QC_CHECKER_CHECKER_H
-#define QC_CHECKER_CHECKER_H
+#ifndef QC_CHECKER_CHECKERDATAPROCESSOR_H
```
knopers8: Please change to QC_CHECKER_CHECKER_H if you prefer to keep `Checker` name.

Modules/Daq/src/DaqTask.cxx
```diff
@@ -100,41 +100,42 @@ void DaqTask::startOfCycle()
...
 {
-  if (dataSet->empty()) {
-    cout << "Empty vector!" << endl;
-    return;
-  }
+  QcInfoLogger::GetInstance() << "initialize SkeletonTaskDPL" << AliceO2::InfoLogger::InfoLogger::endm;
+
+
+//what does it mean to have several inputs ? is it that we defined several in the config file ?
+// If I am connected to the readout can I ever receive several inputs ?
```
knopers8: Theoretically, one could also receive several inputs along with readout. However, they should be in the same timebase (`timesliceID`) , so DPL can match them. Then, they would be accesible by their binding:
```
auto a = ctx.inputs().get("readout");
auto b = ctx.inputs().get("something-after-readout");

File: Modules/Daq/src/DaqTask.cxx [hunk 4]
```diff
@@ -100,41 +100,42 @@ void DaqTask::startOfCycle()
...
-  }
-  TDatime now;
-  if ((now.Get() - mTimeLastRecord) >= 1) {
-    mIds->SetPoint(mNPoints, now.Convert(), dataSet->at(0)->getData()->header.id);
-    mNPoints++;
-    mTimeLastRecord = now.Get();
-  }
+  mNumberSubblocks->Fill(ctx.inputs().size());
+
+  // TODO if data has an id (like event id), we should plot it
```
knopers8: It should be in `DataProcessingHeader`, but after I make the dispatcher send it

Modules/Daq/test/testQcDaq.cxx
```diff
@@ -36,7 +37,7 @@ BOOST_AUTO_TEST_CASE(instantiate_task)
   task.startOfCycle();
   auto producer = AliceO2::DataSampling::DataBlockProducer(false, 1024);
   DataSetReference dataSet = producer.getDataSet();
-  task.monitorDataBlock(dataSet);
+//  task.monitorDataBlock(dataSet);// TODO
```
knopers8: The problem is that we still do not have a feature to test the algorithms in separation. The only way would be to create actual workflow executable, send some test data to tested data processor, and exit with 0 if the results are correct. This method is actually used in the main O2 repo.

Framework/CMakeLists.txt
```diff
@@ -39,16 +35,16 @@ set(
 
 set(
   SRCS_TOBJECT2JSON
-  src/TObject2Json.cxx
-  src/TObject2JsonServer.cxx
-  src/TObject2JsonWorker.cxx
-  src/TObject2JsonBackendFactory.cxx
-  src/TObject2JsonCcdb.cxx
+  src/tobject2json/TObject2Json.cxx
```
knopers8: It seems like this file was lost in the process of moving

Barthelemy:
I have renamed SkeletonDPL into Skeleton.

Framework/CMakeLists.txt
Barthelemy: For me it is there. Do you see it missing ?

Barthelemy: ok, fixed, problem with the gitignore

Barthelemy:
@knopers8 I have fixed what you raised. Thank you. For the rest, in particular the name of the Task class I propose to rediscuss when you are back. If you are ok with the current state I will merge.

Framework/include/QualityControl/Checker.h
knopers8: Aah, indeed. It does not inherit from framework::Task, though it has similar interface. And indeed, I wanted to have some common structure of these DPL devices. Then we could have one factory for them all.
I am fine with leaving it like it is, but when we discuss, we can consider also `TaskRunner`.

knopers8:
I still see a failing build because of ```
make[2]: *** No rule to make target `/mnt/mesos/sandbox/sandbox/sw/SOURCES/QualityControl/build_QualityControl_o2/0/Framework/include/QualityControl/CheckerConfig.h', needed by `Framework/QualityControlDict.cxx'.  Stop.
```, but aside from that, I am ok with the changes

Framework/include/QualityControl/Checker.h
Barthelemy: Funny you mention `TaskRunner`, that also crossed my mind. We got a winner :)

Barthelemy:
Yeah, I would say that it is a spurrious error. When I had that I ended up deleting the build directory and it fixed it. 

I will merge and if it appears in other builds, I will ask Dario to clear it.

PR #71, opened 2018-10-11
Better handle mysql 

Barthelemy:
Check not only MySQL but also the corresponding class in ROOT.


Barthelemy:
https://github.com/alisw/alidist/issues/1324

PR #74, opened 2018-10-15
Reformat the whole QC code (QC-100)

Barthelemy:
QC-100

### DIFF
--- a/Framework/include/QualityControl/TaskDataProcessorFactory.h
+++ b/Framework/include/QualityControl/TaskDataProcessorFactory.h
@@ -8,9 +8,12 @@
 
 #include "Framework/DataProcessorSpec.h"
 
-namespace o2 {
-namespace quality_control {
-namespace core {
+namespace o2
+{
+namespace quality_control
+{
+namespace core
+{
 
 /// \brief Factory in charge of creating DataProcessorSpec of QC task
 class TaskDataProcessorFactory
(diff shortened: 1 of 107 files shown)


Modules/Skeleton/src/SkeletonCheck.cxx
```diff
@@ -6,75 +6,68 @@
...
 
-namespace o2 {
-namespace quality_control_modules {
-namespace skeleton {
-
-SkeletonCheck::SkeletonCheck()
+  namespace o2
 {
-}
+  namespace quality_control_modules
```
knopers8: something bad happened with the namespace indents

Framework/include/QualityControl/TaskDataProcessorFactory.h
```diff
@@ -8,9 +8,12 @@
...
 
-namespace o2 {
-namespace quality_control {
-namespace core {
+namespace o2
+{
+namespace quality_control
+{
+namespace core
+{
```
knopers8: How about changing nested namespaces to c++17 syntax - `namespace o2::quality_control::core {`? Not now, but when we have common agreement on that.

Barthelemy: I think that it is a good idea. I was not aware of it. Could you create a ticket in JIRA ? 
O2 General

Barthelemy:
@knopers8 I have fixed the formatting

PR #76, opened 2018-10-16
Adopt to new Configuration

awegrzyn:
This should be merged together with https://github.com/alisw/alidist/pull/1341
...also QC and Readout tags should be bumped there.


Barthelemy:
>  ...also QC and Readout tags should be bumped there.
why ? I think that they are good as they are

PR #78, opened 2018-10-18
[WIP] Use getRecursive to optimise Configuration usage

awegrzyn:


awegrzyn:
@Barthelemy Let me know this is fine with you. I've only done a small part.

Barthelemy:
ok, I see, I will do the rest. I will already merge that.

PR #82, opened 2018-10-24
Use new Data Sampling API

knopers8:
It will need to be merged together with https://github.com/AliceO2Group/AliceO2/pull/1428 , but after I fix a few more things.


knopers8:
@Barthelemy Could you have a look?

PR #83, opened 2018-10-25
Update InputSpec usage to changes in O2

knopers8:
Now `InputSpec` can be constructed only with the full set of members.


knopers8:
I can see already FLP proto build failing, this should fix it. Merging

PR #87, opened 2018-11-01
Reworked QC infrastructure generation

knopers8:
This commit introduces a main qc factory which can be used to generate a full topology with just one function call. As the `TaskRunner` can be placed either locally on FLP or EPN machine or remotely on dedicated QC servers, a distinction between this two has been introduced. When a task is 'local', the `TaskRunners` will be generated by `generateLocalInfrastructure()`, but mergers and checkers by `generateRemoteInfrastructure()`. If a task is 'remote', both `taskRunner` and `checker` will be generated as the remote infrastructure.

I also deleted some deprecated .ini config files.

I will update the readme when we agree about the code and names.


Framework/basic.json
```diff
@@ -13,18 +13,18 @@
...
         "className": "o2::quality_control_modules::skeleton::SkeletonTask",
         "moduleName": "QcSkeleton",
-        "outputDataOrigin": "ITS",
         "cycleDurationSeconds": "10",
         "maxNumberCycles": "-1",
-        "outputDataDescription": "HIST_SKLT_TASK",
-        "dataSamplingPolicy": "its-raw"
+        "dataSamplingPolicy": "its-raw",
+        "taskParameters": {
+          "nothing": "rien"
```
Barthelemy: merci

Framework/example-default.json
```diff
@@ -16,53 +13,71 @@
...
+        "location": "local",
+        "machines": [
+          "o2flp1",
+          "o2flp2"
+        ]
       }
+    }
+  },
+  "dataSamplingPoliciesFile_comment": "In case that policies are stored in different file, specify its path below. When both dataSamplingPolicies and dataSamplingPoliciesFile are specified, the latter has higher priority",
+  "dataSamplingPoliciesFile": "json:///home//alice/QualityControl/dataSamplingConfig.json",
```
Barthelemy: spurrious slash ?

Framework/example-default.json
```diff
@@ -16,53 +13,71 @@
...
+        {
+          "binding": "mft-clusters",
+          "dataOrigin": "MFT",
+          "dataDescription": "CLUSTERS"
+        }
+      ],
+      "subSpec": "0",
+      "samplingConditions": [
+        {
+          "condition": "payloadSize",
```
Barthelemy: Could we add a "fraction" parameter here ?

Framework/include/QualityControl/QualityControlFactory.h
```diff
@@ -0,0 +1,58 @@
...
+///
+/// A factory class which can generate QC topologies given a configuration file (example in Framework/basic.json and
+/// Framework/example-default.json). As QC topologies will be spread on both processing chain machines and dedicated
+/// QC servers, a _local_ vs. _remote_ distinction was introduced. Tasks which are _local_ should have taskRunners
+/// placed on FLP or EPN machines and their results should be merged and checked on QC servers. The 'remote' option
+/// means, that full QC chain should be located on remote (QC) machines. For the laptop development, use 'remote' tasks
+/// and generateRemoteInfrastructure() to obtain the full topology in one go.
+///
+/// \author Piotr Konopka
+class QualityControlFactory
```
Barthelemy: I am not (yet) convinced by the name of the class because looking at it I would expect to be able to do getQualityControl(...) , whatever that means. Think of a BankAccountFactory, or a TaskFactory or a CheckerFactory. 

What about TopologyGenerator ?

Framework/include/QualityControl/TaskFactory.h
```diff
@@ -55,7 +55,7 @@ class TaskFactory
     std::string library = "lib" + taskConfig.moduleName;
     logger << "Loading library " << library << AliceO2::InfoLogger::InfoLogger::endm;
     int libLoaded = gSystem->Load(library.c_str(), "", true);
-    if (libLoaded) {
+    if (libLoaded < 0) {
```
Barthelemy: Good catch. How did it ever work ?

Framework/include/QualityControl/TaskRunner.h
```diff
@@ -65,7 +65,7 @@ using namespace std::chrono;
 class TaskRunner
 {
  public:
-  TaskRunner(std::string taskName, std::string configurationSource);
+  TaskRunner(std::string taskName, std::string configurationSource, size_t id = 0);
```
Barthelemy: Could you add doxygen ?

Barthelemy: Just because I don't know what id refer to

Framework/src/QualityControlFactory.cxx
```diff
@@ -0,0 +1,104 @@
...
+using boost::property_tree::ptree;
+
+namespace o2
+{
+namespace quality_control
+{
+namespace core
+{
+
+QualityControlFactory::QualityControlFactory()
```
Barthelemy: you could use "=default"

Framework/example-default.json
knopers8: I am not sure I understand what for - this is a condition which accepts data which has its payload size within specified range. If you wanted to add another, random condition, you would do it like this:
```
"samplingConditions": [
  {
    "condition": "payloadSize",
    "upperLimit": "5000",
    "lowerLimit": "1000"
  },
  {
    "condition": "random",
    "fraction": "0.1",
    "seed": "1441"
  }
],
```

Framework/include/QualityControl/QualityControlFactory.h
knopers8: `TopologyGenerator` works for me, I wasn't attached to the original name neither

Framework/include/QualityControl/TaskFactory.h
knopers8: Before, during the topology generation, we happened to load each library only once. It was no longer the case when I started to work on that complex topology example, then the bug revealed itself.

Framework/example-default.json
knopers8: indeed

Framework/src/QualityControlFactory.cxx
knopers8: Since the methods are static anyway, then how about deleting constructors?

Framework/include/QualityControl/QualityControlFactory.h
knopers8: Though we would use then two names for the same thing. Look at: `TopologyGenerator::generateLocalInfrastructure()`. I am not sure which one, infrastructure or topology, is better.

Barthelemy: `InfrastructureGenerator` sounds good as well !

Framework/src/QualityControlFactory.cxx
Barthelemy: yes, you are right.

Framework/example-default.json
Barthelemy: but what if I want 1% of the data between 1 and 4 kB ?

Barthelemy: for me "fraction" is meaningful in most conditions.

Barthelemy: oh.... provided that I did not get it, maybe this should be documented in the end user guide (whatever that is). 
Thanks !

knopers8: I would say that it is in the second sentence [there](https://github.com/AliceO2Group/AliceO2/tree/dev/Framework/Core#data-sampling-conditions), but I can add something like "it means that one can get 1% of data between 1 and 4 kB by placing two conditions in the list"

Barthelemy: you are right it is alraedy there.

knopers8:
I amended all the changes to the last commit.
Though, I am still not very enthusiastic about the interface name. Look at this:
```
auto qcInfrastructure = InfrastructureGenerator::generateRemoteInfrastructure(qcConfigurationSource);
```
The information, that it is QC infrastructure is only indicated in the result variable name. We could do:
```
auto qcInfrastructure = quality_control::core::InfrastructureGenerator::generateRemoteInfrastructure(qcConfigurationSource);
``` 
but this is quite long and involves a lot of nested namespaces. Maybe we could do something better than that?

I really like Data Sampling interface, which is just:
```
  DataSampling::GenerateInfrastructure(specs, qcConfigurationSource);
```
Maybe we could have something similar in QC?

Barthelemy:
I agree, it is neater. Do you want to open a new PR with a new naming scheme and we merge this PR ?

knopers8:
We can do it like that

Barthelemy:
there is a build failure though

knopers8:
I believe this is because the dictionary was not rebuilt in the build after I pushed the changes.

PR #88, opened 2018-11-06
An advanced example of QC topology

knopers8:
This is an executable showing a more complicated QC topology. It pretends to spawn 4 separate topologies - 3 of them consist of some dummy processing chain, a dispatcher and a local QC task. The last one represents the remote QC servers topology, which has a merger (joining the results from local QC tasks), a checker (checks the result of the previous) and a different, remote QC task with associated checker. Here they are joined into one big topology just to present the concept. The tasks do not provide any valid results, this is just a topology example.

It is still quite messy and includes some workarounds to make it work.


knopers8:
Screenshot

Barthelemy:
should I review or is it a WIP ?

knopers8:
I would like to work on it a bit more, no need to review it yet

knopers8:
For now, it is as good as it can get. Some workaround and boiler plate code is inevitable now, but it will disappear when we have proper mergers and DPL wildcards matching.

Barthelemy:
I think that it is a very good use case.

Framework/src/runAdvanced.cxx
```diff
@@ -0,0 +1,152 @@
...
+/// If you have glfw installed, you should see a window with the workflow visualization and sub-windows for each Data
+/// Processor where their logs can be seen. The processing will continue until the main window it is closed. Regardless
+/// of glfw being installed or not, in the terminal all the logs will be shown as well.
+
+#include <Framework/CompletionPolicyHelpers.h>
+#include "Framework/DataSampling.h"
+
+using namespace o2::framework;
+
+void customize(std::vector<CompletionPolicy>& policies)
```
Barthelemy: what is this ?

Framework/src/runAdvanced.cxx
```diff
@@ -0,0 +1,152 @@
...
+/// \file    runAdvanced.cxx
+/// \author  Piotr Konopka
+///
+/// \brief This is an executable showing a more complicated QC topology.
+///
+/// This is an executable showing a more complicated QC topology. It pretends to spawn 4 separate topologies - 3 of them
+/// consist of some dummy processing chain, a dispatcher and a local QC task. The last one represents the remote
+/// QC servers topology, which has a merger (joining the results from local QC tasks), a checker (checks the result of
+/// the previous) and a different, remote QC task with associated checker. Here they are joined into one big topology
+/// just to present the concept.
```
Barthelemy: As you have a screenshot of the resulting topology, I would add it here.

Framework/src/InfrastructureGenerator.cxx
```diff
@@ -39,7 +39,7 @@ WorkflowSpec InfrastructureGenerator::generateLocalInfrastructure(std::string co
 
         if (machine.second.get<std::string>("") == host) {
           // todo: optimize it by using the same ptree?
-          workflow.emplace_back(taskRunnerFactory.create(task.first, configurationSource, id));
+          workflow.emplace_back(taskRunnerFactory.create(task.first, configurationSource, id, true));
```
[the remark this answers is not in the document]
knopers8: Yes, it is because the `HistoMerger` expects histogram differences which are accumulated to its inner cache. I forgot about it earlier.

Framework/include/QualityControl/TaskRunnerFactory.h
```diff
@@ -22,7 +22,8 @@ class TaskRunnerFactory
   TaskRunnerFactory();
   virtual ~TaskRunnerFactory();
 
-  o2::framework::DataProcessorSpec create(std::string taskName, std::string configurationSource, size_t id = 0);
+  o2::framework::DataProcessorSpec
```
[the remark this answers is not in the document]
knopers8: will do

Framework/src/ObjectsManager.cxx
```diff
@@ -30,7 +30,7 @@ namespace core
 
 ObjectsManager::ObjectsManager(TaskConfig& taskConfig) : mTaskName(taskConfig.taskName)
 {
-  startPublishing(&mObjectsList, MonitorObject::SYSTEM_OBJECT_PUBLICATION_LIST);
+//  startPublishing(&mObjectsList, MonitorObject::SYSTEM_OBJECT_PUBLICATION_LIST);
```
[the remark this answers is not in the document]
knopers8: It should be completely redundant when we send MOs in a vector or list. In this case, it was destroying the merging, hence I commented it out.

Framework/src/runAdvanced.cxx
knopers8: This is some additional configuration of the topology, which is done by implementing `customize` functions and placing them **above**  `runDataProcessing.h` header. Personally, I am not a fan of this interface :)
In this case, both Dispatcher and Merger are configured to accept incoming messages without waiting for the rest of inputs.

Framework/src/runAdvanced.cxx
knopers8: A link? Where can I upload it so it does not disappear?

knopers8: Aaa, I can just add the image to the repo...

Framework/src/runAdvanced.cxx
```diff
@@ -0,0 +1,152 @@
...
+/// \image html qcRunAdvanced.png
```
knopers8: Is this how to do it? I also placed `IMAGE_PATH = images` in `doxygen.in`.
I failed with building the documentation on my laptop, I have no way to check if it works.

Framework/src/ObjectsManager.cxx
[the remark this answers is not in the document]
Barthelemy: I think that there is a misunderstanding on what is this mObjectsList. If I am not mistaken it was meant for the GUIs to know what are the objects being published by a certain Task.

Framework/src/runAdvanced.cxx
Barthelemy: I see. Could you comment it ? no need to make it long.

Framework/src/runAdvanced.cxx
Barthelemy: yes, I think that's the way. I should try the documentation generation. let's do it this way for the time being.

Framework/src/ObjectsManager.cxx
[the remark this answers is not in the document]
knopers8: According to the commit:
_Add support for the publication of the list of all published objects._

_This is needed by the CCDB backend as we have no ways of asking for the list of objects. It introduces the (first) so-called system object, i.e. an object that is transmitted by tasks behind the scene the same way the explicit MonitorObjects are. At the moment, the list of objects is a comma separated list of names. It is interpreted by the CCDB backend to provide the list of published objects. (QC-53)_

So I will try to find another way to make this work with current merger.

Framework/src/runAdvanced.cxx
knopers8: OK

Framework/src/ObjectsManager.cxx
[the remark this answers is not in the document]
knopers8: I added a check in `HistoMerger` to accept classes which are TH1, so that object list string does not interfere.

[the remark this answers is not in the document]
Barthelemy: Thank you very much for checking this for me. It saved me quite some effort. 

At the time it was needed, but I am not sure it is still needed. The GUI use an Information Service (the stupid one I wrote or the future official one) to know what objects are being published by a task (https://alice.its.cern.ch/jira/browse/OGUI-147). For the case when we explore all the objects published (in the past) the CCDB now provides an API for it. 

My take is that it can stay commented out.

knopers8:
All suggestions applied, checks are passed - merging

PR #91, opened 2018-11-14
[QC-88] Exit TaskRunner when cycles number reaches the specified maximum

knopers8:
Before, when task reached its maximum number of cycles, it just didn't accept any new data, but still published the latest MOs.
Now it can quit instead, which shuts down the whole topology. I am not sure if this is the best possible behaviour, but at least we have some starting point to discuss it.

### DIFF
--- a/Framework/src/TaskRunner.cxx
+++ b/Framework/src/TaskRunner.cxx
@@ -23,6 +23,7 @@
 #include "Configuration/ConfigurationFactory.h"
 #include "Framework/RawDeviceService.h"
 #include "Framework/DataSampling.h"
+#include "Framework/ControlService.h"
 #include "Monitoring/MonitoringFactory.h"
 #include "QualityControl/QcInfoLogger.h"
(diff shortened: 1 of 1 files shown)


Barthelemy:
Thank you. I see, as you noticed, a number of possible issues. They are not technical. 
The use case for this is usually to run locally, very much for debug, and to read a given number of events to treat them and exit. So, typically it should be a "basic" workflow. In such a case it works. 

If we now imagine a workflow with 2 tasks, each with a different max number of samples, what happens ?

knopers8:
The first task which reaches max number of cycles shuts down everything (confirmed on qcRunAdvanced).
Now I also discovered that one can use `readyToQuit(false)` (instead of `true`), which informs the driver, that only this process is ready to quit... but it does not quit it. I guess that in this case the driver waits for all processes to declare readiness to quit and only after that it shuts them down.

knopers8:
A message that maximum cycles number is reached will be the best option for now. Closing this in favour of new PR.

PR #93, opened 2018-11-16
[QC-114] Use a container to send all MOs at once

knopers8:
This just a proof that both `TMap` and `TObjArray` can be used as such container. As I expected, using `vector<MonitorObject>` didn't work even at the compilation level.
To do:
- [x] choose a container, 
- [x] consider using a messageable container already in `ObjectManager`, so less preparation before sending is required
- [x] adapt `HistoMerger` to accept a container

### DIFF
--- a/Framework/src/ObjectsManager.cxx
+++ b/Framework/src/ObjectsManager.cxx
@@ -31,14 +31,11 @@ namespace core
 ObjectsManager::ObjectsManager(TaskConfig& taskConfig) : mTaskName(taskConfig.taskName)
 {
   startPublishing(&mObjectsList, MonitorObject::SYSTEM_OBJECT_PUBLICATION_LIST);
+  mMonitorObjects.SetOwner(true);
 }
 
 ObjectsManager::~ObjectsManager()
 {
-  for (auto& mMonitorObject : mMonitorObjects) {
-    delete mMonitorObject.second;
-  }
-  mMonitorObjects.clear();
 }
 
 void ObjectsManager::startPublishing(TObject* object, std::string objectName)
@@ -46,7 +43,7 @@ void ObjectsManager::startPublishing(TObject* object, std::string objectName)
   std::string nonEmptyName = objectName.empty() ? object->GetName() : objectName;
   auto* newObject = new MonitorObject(nonEmptyName, object, mTaskName);
   newObject->setIsOwner(false);
-  mMonitorObjects[nonEmptyName] = newObject;
+  mMonitorObjects.Add(newObject);
 
   // update index
   if (objectName != MonitorObject::SYSTEM_OBJECT_PUBLICATION_LIST) {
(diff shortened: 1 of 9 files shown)


Barthelemy:
on it

Framework/src/ObjectsManager.cxx
```diff
@@ -31,22 +31,19 @@ namespace core
 ObjectsManager::ObjectsManager(TaskConfig& taskConfig) : mTaskName(taskConfig.taskName)
 {
   startPublishing(&mObjectsList, MonitorObject::SYSTEM_OBJECT_PUBLICATION_LIST);
+  mMonitorObjects.SetOwner(true);
 }
 
 ObjectsManager::~ObjectsManager()
 {
-  for (auto& mMonitorObject : mMonitorObjects) {
```
Barthelemy: No deletion of the MonitorObjects any more ?

knopers8: `TObjArray mMonitorObjects` is the explicit owner of MOs now (a few lines above).

knopers8:
Again, the o2 build failure is unrelated, merging
```
[  2%] Generating G__O2TPCCAGPUTrackingDict.cxx, G__O2TPCCAGPUTrackingDict_rdict.pcm, ../../../../lib/libO2TPCCAGPUTracking.rootmap
[  2%] Building CXX object Common/MathUtils/CMakeFiles/MathUtils.dir/src/Cartesian2D.cxx.o
input_line_12:3656:10: fatal error: 'TPCFastTransform.h' file not found
#include "TPCFastTransform.h"
         ^~~~~~~~~~~~~~~~~~~~
Error: /mnt/mesos/sandbox/sandbox/sw/slc7_x86-64/ROOT/v6-14-04-8/bin/rootcint: compilation failure (/mnt/mesos/sandbox/sandbox/sw/BUILD/2c1da1efa8448ee35395e7f49a3b0c976238ee6d/O2/Detectors/TPC/reconstruction/TPCCAGPUTracking/G__O2TPCCAGPUTrackingDict50ae978b44_dictUmbrella.h)
gmake[2]: *** [Detectors/TPC/reconstruction/TPCCAGPUTracking/G__O2TPCCAGPUTrackingDict.cxx] Error 1
gmake[1]: *** [Detectors/TPC/reconstruction/TPCCAGPUTracking/CMakeFiles/O2TPCCAGPUTracking.dir/all] Error 2
```

Framework/src/ObjectsManager.cxx
Barthelemy: I saw it but did not realize that the deletion of a TObjArray destroys the pointers it contains.

knopers8: I think it does - [docu](https://root.cern.ch/doc/master/classTObjArray.html#a30ea50379bcd56377448529f781dfb6a)

TObjArray::~TObjArray()
Delete an array.
Objects are not deleted unless the TObjArray is the owner (set via SetOwner()).

Barthelemy: You are correct.

PR #95, opened 2018-11-19
TObject2Json: Fix Ccdb backend creation

awegrzyn:


Barthelemy:
For reference, CCDB ignores most of the parameters : https://github.com/AliceO2Group/QualityControl/blob/290b110cd969fe105880b601efa3d008caf842ee/Framework/src/CcdbDatabase.cxx#L38

PR #98, opened 2018-11-23
Do not use find-modules for Common and Configuration

awegrzyn:


awegrzyn:
@Barthelemy let's wait for checks to pass

PR #103, opened 2018-12-10
Remove MO name (QC-123)

Barthelemy:
* Remove the name attribute from MonitorObject
* Preserve the interface
* RepositoryBenchmark : update to create multiple objects instead of sending the same.

### DIFF
--- a/Framework/include/QualityControl/MonitorObject.h
+++ b/Framework/include/QualityControl/MonitorObject.h
@@ -39,7 +39,7 @@ class MonitorObject : public TObject
  public:
   /// Default constructor
   MonitorObject();
-  MonitorObject(const std::string& name, TObject* object, const std::string& taskName);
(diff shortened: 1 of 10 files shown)


Framework/include/QualityControl/MonitorObject.h
```diff
@@ -134,7 +133,7 @@ class MonitorObject : public TObject
   // TODO : maybe we should always be the owner ?
   bool mIsOwner;
 
-  ClassDefOverride(MonitorObject, 2);
+  ClassDefOverride(MonitorObject, 3);
```
knopers8: What is this for?

Barthelemy: the version of the class must be incremented if its definition changes. It enables ROOT to read types that evolve over time.

PR #104, opened 2018-12-12
Fix compilation of generated module under CC7 (QC-126)

Barthelemy:


knopers8:
We can do it like that, but now the corrected lines do nothing, so we could as well delete them completely

Barthelemy:
indeed :) I will remove them then

PR #109, opened 2019-01-07
Adapt to new DPL API for InputSpec

ktf:


ktf:
This makes things compile for me on macOS. It's unrelated to the other issue about variants, which I am still trying to understand fully (right now, I suspect it's a compiler / runtime bug).

Notice that on Mac there are quite a number of genuine warnings which should probably be fixed, but I leave this to the maintainers of this project.

Barthelemy:
Compiles on my Mac

PR #110, opened 2019-01-08
Fix MySQL database connect function header

knopers8:
There was a quite simple error introduced [here](https://github.com/AliceO2Group/QualityControl/commit/f16756b42c4d4b08cddece6a2089d6def1ea3d78), which got in unnoticed because of larger build problems.

I am still compiling it on my machine with Giulio's fixes to O2, to make sure that it fixes the problem


knopers8:
It builds properly with https://github.com/AliceO2Group/AliceO2/pull/1574 , merging

Barthelemy:
thanks !

PR #113, opened 2019-02-01
[O2-471] Fix the printer in runReadout 

knopers8:
Changes in MO's format were followed in runBasic and runAdvanced, but this exec was forgotten.


knopers8:
build/QualityControl/o2 fails because of failing tests in O2. Tested compilation o2 defaults on my machine - it compiles. Merging.

PR #115, opened 2019-02-14
Allow for an alternative path to the config file

Barthelemy:
the environment variable QUALITYCONTROL_CONFIG_PATH can be set to point to a different config file to be used when running qcRunReadout.


Barthelemy:
I agree for --option, but can we do that ? I might just not be knowledgeable about DPL...

knopers8:
https://alice.its.cern.ch/jira/browse/QC-138?focusedCommentId=226470&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-226470

PR #116, opened 2019-02-14
More output and process monitoring

Barthelemy:


Framework/src/Checker.cxx
```diff
@@ -69,6 +69,9 @@ void Checker::init(framework::InitContext&)
     // configuration of the database
     mDatabase = DatabaseFactory::create(config->get<std::string>("qc.config.database.implementation"));
     mDatabase->connect(config->getRecursiveMap("qc.config.database"));
+    std::cout << "Database that is going to be used : " << std::endl;
```
knopers8: Why not LOG(INFO) or LOG(DEBUG) instead of std::cout?

knopers8: Anyway, we should unify all the usages of `std::cout`, `LOG()` and `InfoLogger << "sth"`

Barthelemy: You are correct, twice :)
I will fix this one.

PR #127, opened 2019-04-01
(QC-150) Allow tasks to run without DS

Barthelemy:
- Config file : introduction of "dataSource" whose "type" can be either "direct" or "dataSamplingPolicy".
- At the moment, one should call qcRunBasic with "--no-data-sampling" AND update the config file (see above)
- only implemented for qcRunBasic

### DIFF
--- a/Framework/src/TaskRunner.cxx
+++ b/Framework/src/TaskRunner.cxx
@@ -194,7 +194,6 @@ void TaskRunner::populateConfig(std::string taskName)
     } else if (type == "direct") {
 
       auto subSpecString = dataSourceTree.get<std::string>("subSpec");
-      LOG(INFO) << "subSpecString : " << subSpecString;
       auto subSpec = std::strtoull(subSpecString.c_str(), nullptr, 10);
 
       header::DataOrigin origin;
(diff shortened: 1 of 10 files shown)


Barthelemy:
In the end, if someone wants to have a direct connection, they should use "direct" dataSource and remove all the dataSamplingPolicies related to the Dispatcher. 

runBasic has an option to switch between two different config files that illustrate the two options.

Barthelemy:
ready for review

Framework/src/TaskRunner.cxx
```diff
@@ -182,11 +181,37 @@ void TaskRunner::populateConfig(std::string taskName)
...
+    std::string type = dataSourceTree.get<std::string>("type");
+
+    if (type == "dataSamplingPolicy") {
+      auto policyName = dataSourceTree.get<std::string>("name");
+      LOG(INFO) << "policyName : " << policyName;
+      mInputSpecs = framework::DataSampling::InputSpecsForPolicy(config, policyName);
+    } else if (type == "direct") {
+
+      auto subSpecString = dataSourceTree.get<std::string>("subSpec");
+      LOG(INFO) << "subSpecString : " << subSpecString;
```
knopers8: I guess this specific log is a leftover from the development. If you want it, the full InputSpec has the operator<< overloaded.

Barthelemy: Totally, I will remove it now.

PR #135, opened 2019-04-15
Fix and publish Doxygen (QC-148)

Barthelemy:


awegrzyn:
You could drop doxygen config file: https://github.com/AliceO2Group/Monitoring/blob/dev/doc/CMakeLists.txt

Barthelemy:
Indeed, I realized that meanwhile (https://aliceo2group.github.io/advanced/doxygen.html). Thanks !

PR #136, opened 2019-04-16
Extract the printer (QC-78)

Barthelemy:


Framework/src/ExamplePrinterSpec.h
```diff
@@ -0,0 +1,52 @@
+//
+// Created by bvonhall on 16/04/19.
+//
```
knopers8: Please add the license notice.

Framework/src/ExamplePrinterSpec.h
```diff
@@ -0,0 +1,52 @@
...
+
+#ifndef QUALITYCONTROL_EXAMPLEPRINTERSPEC_H
+#define QUALITYCONTROL_EXAMPLEPRINTERSPEC_H
+
+namespace o2::quality_control::core
```
knopers8: Could you put it in `examples` namespace instead of `core`? Alternatively, it could be just in the skeleton module, but I do not have a strong opinion on that.

PR #145, opened 2019-05-07
Add Service Discovery classes

awegrzyn:
```
o2::quality_control::core::ServiceDiscovery server{"http://consul:8500", "agent123"};
server._register("task1,task2,task3");
```

PR #146, opened 2019-05-09
Add service discovery calls (QC-176)

Barthelemy:
needs https://github.com/AliceO2Group/QualityControl/pull/145

### DIFF
--- a/Framework/include/QualityControl/ObjectsManager.h
+++ b/Framework/include/QualityControl/ObjectsManager.h
@@ -20,6 +20,7 @@
 #include "QualityControl/MonitorObject.h"
 #include "QualityControl/Quality.h"
 #include "QualityControl/TaskConfig.h"
+#include "ServiceDiscovery.h"
 // ROOT
 #include <TObjArray.h>
 #include <TObjString.h>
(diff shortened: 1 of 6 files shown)


Barthelemy:
I will keep only a few commits that are not part of the other PR : 1d5e3ce and 44b01c4.

Barthelemy:
It seems that the problem of not exiting is back. Also note that we have to use a shared pointer in TaskRunner because unique_ptr are incompatible with tasks in dpl. Maybe we could move the servicediscovery object to objectsManager altogether.

Barthelemy:
We must also get the address to consul from the configuration that is itself in Consul ¯\_(ツ)_/¯

Barthelemy:
commits to keep when PR https://github.com/AliceO2Group/QualityControl/pull/162 is merged.
3be5403 
bdfc850 
d9cec32 
620e808

Barthelemy:
I have retested it. It is fine and works. After #162  is merged we can merge this one, not before.

awegrzyn:
@Barthelemy could you let us know when you merge this PR?
https://github.com/AliceO2Group/WebUi/pull/307

Barthelemy:
Sure, There is a complex web of PRs to merge in the right order. I am sorting it out.

Framework/include/QualityControl/ObjectsManager.h
```diff
@@ -20,6 +20,7 @@
 #include "QualityControl/MonitorObject.h"
 #include "QualityControl/Quality.h"
 #include "QualityControl/TaskConfig.h"
+#include "ServiceDiscovery.h"
```
knopers8: Why not "QualityControl/ServiceDiscovery.h", since it is inside QualityControl?

Barthelemy: it was a mistake. Originally the header was not distributed.

PR #147, opened 2019-05-10
Add feature to let users set custom metadata on MonitorObjects (QC-169)

Barthelemy:


Barthelemy:
@knopers8 The version number should be increased. I left it on purpose because if I increase it, then we break the qcg. As we add data, the qcg is able to deserialize the new MO. It is not clean but better than breaking qcg for everyone. There is another issue with a fix. We can delay this PR as well. 

I will have a look at the check failures on Monday. It is weird indeed.

knopers8:
OK, I thought that not incrementing the version would break the deserialization as well.

PR #149, opened 2019-05-13
Use DPL Task API in TaskRunner

knopers8:


Barthelemy:
@ktf @dberzano any idea what is this error in the o2 build ? 
```c++: error: vfork: Cannot allocate memory```

Barthelemy:
@knopers8 nice, it simplifies considerably the code. I will test it but I have to rebuild something else first.

dberzano:
@Barthelemy it's an error due to the lack of resources on the running node. I am restarting the worker just to be sure.

ktf:
I think there is actually something which is exhausting PIDs on the builder nodes. I've just done the same for the O2 builders.

PR #153, opened 2019-05-20
Rename binaries according to new rules (QC-164)

Barthelemy:


Barthelemy:
I tried to update all references, including in the doc. Old names are still available via symlinks, just to avoid a surge in user support. It should be advertised to the users in a dedicated email.

Framework/CMakeLists.txt
```diff
@@ -139,29 +139,53 @@ set(
...
+  o2-qc-run-readout
+  o2-qc-run-merger-test
+  o2-qc-run-readout-for-data-dump
+  o2-qc-repository-benchmark
 )
 
+# These were the original names before the convention changed.
+# We will get rid of them but for the time being we want to create symlinks to avoid confusion.
+set(
+        EXE_OLD_NAMES
```
knopers8: Could you modify it to have the two-space indentation, as in the rest of the file?

Barthelemy:
agreed and fixed !

knopers8:
Thank you!

PR #158, opened 2019-05-21
Forbid publishing twice the same object (QC-178)

Barthelemy:


Framework/src/ObjectsManager.cxx
```diff
@@ -35,6 +35,11 @@ ObjectsManager::~ObjectsManager()
 
 void ObjectsManager::startPublishing(TObject* object)
 {
+  if (mMonitorObjects.FindObject(object->GetName()) != 0) {
+    QcInfoLogger::GetInstance() << "Object already published (" << object->GetName() << "), we won't republish"
```
knopers8: Maybe 'being published' instead of 'published'? I think this might be interpreted as "Object have been already published", though my understanding might be wrong here.

Framework/test/testObjectsManager.cxx
```diff
@@ -0,0 +1,33 @@
+///
```
knopers8: Copyright notice ;)

Framework/src/ObjectsManager.cxx
Barthelemy: correct, fixing it now.

PR #159, opened 2019-05-21
Add a feature to unpublish (QC-180)

Barthelemy:


Barthelemy:
when https://github.com/AliceO2Group/QualityControl/pull/158 is ready, remove the two extra commits

PR #162, opened 2019-05-24
Fix mo serialization (QC-183)

Barthelemy:


Barthelemy:
Remember to remove the two extra commits.

Framework/test/testMonitorObject.cxx
```diff
@@ -61,4 +67,48 @@ BOOST_AUTO_TEST_CASE(mo_check)
...
+  cout << "quality : " << obj.getQuality() << endl;
+  cout << "check numbers : " << obj.getChecks().size() << endl;
+  CheckDefinition c = obj.getCheck("name2");
+  cout << "check2 libraryName : " << c.libraryName << endl;
+  TFile file("/tmp/test.root", "RECREATE");
+  obj.Write(obj.getName().data());
+  file.Close();
+
+  cout << "***" << endl;
+  TFile file2("/tmp/test.root");
```
knopers8: If there are several builds/test running on the same CI machine, wouldn't that cause some race conditions? Though I don't have an idea how to solve it and anyway they would be very rare probably. Maybe a random number in the name?

Barthelemy: A random number would do it. I will create a separate PR so that this one can go ahead

Barthelemy: Thank you , this is a good remark.

PR #164, opened 2019-05-27
remove docs folder

Barthelemy:


knopers8:
Travis' clang-format complains that one of the deleted file is missing. I guess it tries to compare changes with something that doesn't exist anymore. I think it is fine to merge, do you agree @Barthelemy ?
```
fatal: ambiguous argument 'docs/README.md': unknown revision or path not in the working tree.
```

Barthelemy:
I will make one more check before I am uncomfortable about this error.

Barthelemy:
clang-format works on my computer with this branch

PR #165, opened 2019-05-27
Fix most of the warnings

knopers8:
The rest of the warnings are inside the imgui source files. Is there a way to ignore them?

### DIFF
--- a/Framework/include/QualityControl/ServiceDiscovery.h
+++ b/Framework/include/QualityControl/ServiceDiscovery.h
@@ -10,6 +10,7 @@
 #include <memory>
 #include <string>
 #include <thread>
+#include <boost/asio/ip/host_name.hpp>
 
 namespace o2::quality_control::core
 {
@@ -58,7 +59,11 @@ class ServiceDiscovery
   /// Health check thread loop
   void runHealthServer(unsigned int port);
 
-  static inline std::string GetDefaultUrl(); ///< Provides default health check URL
+  static inline std::string GetDefaultUrl() ///< Provides default health check URL
+  {
+    return boost::asio::ip::host_name() + ":" + std::to_string(7777);
+  }
+
 };
 
 } // namespace o2::quality_control::core
--- a/Framework/include/QualityControl/ServiceDiscovery.h
+++ b/Framework/include/QualityControl/ServiceDiscovery.h
@@ -63,7 +63,6 @@ class ServiceDiscovery
   {
     return boost::asio::ip::host_name() + ":" + std::to_string(7777);
   }
-
 };
 
 } // namespace o2::quality_control::core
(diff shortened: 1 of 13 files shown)


Framework/include/QualityControl/ServiceDiscovery.h
```diff
@@ -58,7 +59,11 @@ class ServiceDiscovery
   /// Health check thread loop
   void runHealthServer(unsigned int port);
 
-  static inline std::string GetDefaultUrl(); ///< Provides default health check URL
+  static inline std::string GetDefaultUrl() ///< Provides default health check URL
```
knopers8: To be inlined, a function must actually have its definition by the declaration.

Framework/src/DataDumpGui.cxx
```diff
@@ -139,7 +139,6 @@ void updatePayloadGui()
     // header row
     ImGui::Separator();
     resizeColumns(representation /*, old_representation*/);
-    ImGui::Text("");
```
knopers8: I confirmed that it doesn't break the GUI

Framework/src/InformationService.h
```diff
@@ -38,7 +38,7 @@ namespace pt = boost::property_tree;
 /// See runInformationService.cxx for the steering code.
 ///
 /// Example usage :
-///      qcInfoService -c /absolute/path/to/InformationService.json -n information_service \\
+///      qcInfoService -c /absolute/path/to/InformationService.json -n information_service
```
knopers8: I couldn't think of another way to get rid of that warning, any ideas?

Barthelemy: ANy ways, this is going to disappear as it is replaced by the ServiceDiscovery of Adam and George.

knopers8:
The 'missing initializer for members' warnings will fixed in https://github.com/AliceO2Group/AliceO2/pull/2016 , if it is accepted

Barthelemy:
Thank you very much. 
I get the following as well : 
```
In file included from /Users/bvonhall/dev/alice/sw/SOURCES/QualityControl/knopers8-fix-warnings/0/Framework/src/runBasic.cxx:66:
/Users/bvonhall/dev/alice/sw/SOURCES/QualityControl/knopers8-fix-warnings/0/Framework/src/ExamplePrinterSpec.h:34:41: warning: moving a temporary object prevents copy elision [-Wpessimizing-move]
    std::shared_ptr<TObjArray> moArray{ std::move(DataRefUtils::as<TObjArray>(*processingContext.inputs().begin())) };
                                        ^
```
Is it only on mac ?

knopers8:
I haven't seen this one on CC7, I will fix it as well.

knopers8:
The o2-dataflow build is stuck...

PR #174, opened 2019-05-31
Use dedicated function to set -w flag to imgui files

knopers8:
Actually I feel a little bit more sure with this function, as it is made exactly to do this thing


Barthelemy:
is build/QualityControl/o2 stuck ?

knopers8:
This time it was just slow

PR #179, opened 2019-06-06
Adapt to new names

ktf:


ktf:
@Barthelemy could you please verify and merge this? This is breaking the O2Suite package (and the AliceO2 CI with it).

Barthelemy:
ok, on it

PR #180, opened 2019-06-06
Use workflow merging feature

knopers8:


Framework/src/runQC.cxx
```diff
@@ -0,0 +1,82 @@
...
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+///
+/// \file    runQC.cxx
+/// \author  Piotr Konopka
+///
+
```
Barthelemy: Could you add a few lines to explain what it is ? like you did for runDataProducer

Barthelemy:
This is an excellent development !
I will test it locally before putting the approval.

Barthelemy:
Tested, works very well, instructions clear and working. Congrats !

knopers8:
It looks like there is a bug in the CI - it is being shown that the PR is WIP. When entering into the details, it actually refers to another PR (The title "[WIP] Reduce compilation time of the QC project" contains "WIP".)

Framework/src/runQC.cxx
knopers8: will do

PR #181, opened 2019-06-11
Extend CCDB interface to get past data (QC-186)

Barthelemy:


Barthelemy:
I will fix the compilation errors tomorrow morning

PR #182, opened 2019-06-12
Use queries for describing desired data

knopers8:
To work, it needs to be merged after Data Sampling uses queries, but it is already ready for discussion.

### DIFF
--- a/Framework/advanced.json
+++ b/Framework/advanced.json
@@ -56,19 +56,7 @@
       "id": "tst1",
       "active": "true",
       "machines": [],
-      "dataHeaders": [
-        {
-          "binding": "sum",
-          "dataOrigin": "TST",
-          "dataDescription": "SUM"
-        },
-        {
-          "binding": "param",
-          "dataOrigin": "TST",
-          "dataDescription": "PARAM"
-        }
-      ],
-      "subSpec": "2",
+      "query" : "sum:TST/SUM/2;param:TST/PARAM/2",
       "samplingConditions": [
         {
           "condition": "random",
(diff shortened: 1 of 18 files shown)


Framework/advanced.json
```diff
@@ -76,14 +64,7 @@
...
       "machines": [],
-      "dataHeaders": [
-        {
-          "binding": "data",
-          "dataOrigin": "TST",
-          "dataDescription": "DATA"
-        }
-      ],
-      "subSpec": "*",
+      "query" : "data:TST/RAWDATA",
```
Barthelemy: I am wondering if it would not be better to be explicit here : 
```
"query" : "data:TST/RAWDATA/*",
```
What do you think ?

Barthelemy: Or maybe it is imposed by upstream ?

knopers8: In fact the version with * is not properly parsed and I forgot to check out if this is a bug or a lack of feature. @ktf Does the query syntax expect * for subSpec, like in `"data:TST/RAWDATA/*"`?

ktf: I would need to check but I am on holidays. Maybe the syntax is simply “data:TST/RAWDATA”?

ktf: Or: “data:TST/RAWDATA/$1”

knopers8:
@Barthelemy There were some additional changes here since the last time you have reviewed. Do you have any additional comments?
I will trigger the build with some dull commit to make sure it compiles after https://github.com/AliceO2Group/AliceO2/pull/2095 has been merged.

Barthelemy:
@knopers8 you can trigger the build and I am reviewing

Framework/advanced.json
Barthelemy: I maintain that it would be better to be explicit here. However, we can proceed like that.

knopers8: So neither of `data:TST/RAWDATA/*` and `data:TST/RAWDATA/$1` work, the only way now is to leave it unspecified.

knopers8:
I have added the format of the "query" nearby its first occurence in the documentation. As for more extensive description, I would put it to the "Advanced/Configuration files details" section, which I would like to rewrite soon in another PR. Unless you disagree?

Barthelemy:
I agree !

Framework/advanced.json
ktf: You are right, I will try to add it.

PR #184, opened 2019-06-13
Move service discovery to objectsmanager and some clean up (QC-193, QC-189)

Barthelemy:


knopers8:
The application is not blocked anymore, but I can see "[30270]: bind: Address already in use" in the logs.

Barthelemy:
@knopers8 Are you sure it is linked to the Service Discovery ? Do you have something running on port 7777 ? (I think that it is the default)

knopers8:
I have checked, nothing is running on that port.

knopers8:
I am pretty sure it is connected somehow to Service Discovery. When it was getting stuck before, it was always stuck in that place.

Barthelemy:
Should I merge nevertheless ? the test failure is unrelated.

awegrzyn:
I can take a look at it today.

knopers8:
It improves the situation, so I am fine with merging this.

Framework/src/ObjectsManager.cxx
```diff
@@ -24,9 +24,12 @@ using namespace std;
 namespace o2::quality_control::core
 {
 
-ObjectsManager::ObjectsManager(TaskConfig& taskConfig, std::shared_ptr<ServiceDiscovery> serviceDiscovery) : mTaskName(taskConfig.taskName), mServiceDiscovery(serviceDiscovery), mUpdateServiceDiscovery(false)
+ObjectsManager::ObjectsManager(TaskConfig& taskConfig) : mTaskName(taskConfig.taskName), mUpdateServiceDiscovery(false)
 {
   mMonitorObjects.SetOwner(true);
+
+  // register with the discovery service
+  mServiceDiscovery = std::make_unique<ServiceDiscovery>("http://consul-test.cern.ch:8500", mTaskName);
```
awegrzyn: Maybe it's better to move the URL to config file?

Barthelemy: yes, let me do that

Barthelemy: Done

Barthelemy:
We duplicate monitoring and consul url in all json files. We should see how to do it better. For now it will do. 
@awegrzyn The test testObjectsManager::invalid_url_test shows that the ServiceDiscovery hangs if I provide a wrong url. It hangs in the destructor of ServiceDiscovery (during join) when ObjectsManager exits.

awegrzyn:
Sure but when I just run `o2-qc-run-basic` it does not happen, what argument shall I pass in order to set URL?

Barthelemy:
@awegrzyn I think that it would be better to work with the test because it excludes the DPL. 
I'll pass by your office.

Barthelemy:
@awegrzyn To run the test do 
```
cd alice/sw/BUILD/QualityControl-latest/QualityControl
./bin/testObjectsManager
```
It hangs (even after your commit).

Barthelemy:
@awegrzyn works well! Thank you
I have fixed the formatting and wait for the tests. If everything is green, I will merge it.

awegrzyn:
Ok, but we need to verify whether it works well in the real scenario.

Barthelemy:
Yes, but the plan is to merge , then release, then let you play with it.

PR #185, opened 2019-06-13
Use a different filename in different instances of the test (QC-188)

Barthelemy:


knopers8:
I can still make it break by running 4 instances of this test at the same time. One or two instances fail:
```
Error in <TFile::TFile>: file /tmp/test1560438163943.root does not exist
/home/pkonopka/alice/sw/SOURCES/QualityControl/o2-output-spec-wildcards/0/Framework/test/testMonitorObject.cxx(103): error: in "mo_save": check mo != nullptr has failed [0 == nullptr]
mo : 0

 *** Break *** segmentation violation
```

knopers8:
Switching milliseconds to nanoseconds fixes it.

PR #186, opened 2019-06-13
Fix unused argument warning

knopers8:
It was actually introduced in https://github.com/AliceO2Group/QualityControl/pull/181 which I shamefully approved.
@Barthelemy Can we merge https://github.com/alisw/alidist/pull/1640 after this is fixed?


Barthelemy:
@knopers8 sure

Barthelemy:
when this is merged I will retrigger a build on 1640 to see that we are able to build.

PR #187, opened 2019-06-17
Install tests under tests

Barthelemy:


knopers8:
In testInfrastructureGenerator.cxx the config file path is .../**test**/testQCFactory.json, could you change it to **tests**?

Barthelemy:
1. The CI should use `make test` which is unchanged. 
2. yes, I am going to do it now.

Barthelemy:
2. Done but testInfrastructureGenerator fails for other reasons now. I'll let you check, probably in a different PR, what the problems are.

knopers8:
OK, I will have a look, thanks.

PR #188, opened 2019-06-18
Fixing tests and improving coverage

knopers8:


Framework/test/testWorkflow.cxx
```diff
@@ -0,0 +1,103 @@
+// Copyright CERN and copyright holders of ALICE O2. This software is
```
knopers8: This test runs a DPL workflow with full QC chain and returns 0 if it works correctly (they do it in a similar way in O2). To make it behave properly, we need a timeout of around 15-30 seconds - it might not exit automatically when something is wrong. Also, it needs to be ran on machine with glfw, without ssh and docker, or with '-b' flag. Is this the case already for QC?

Framework/test/testChecker.cxx
```diff
@@ -0,0 +1,71 @@
...
+
+using namespace o2::quality_control::checker;
+using namespace std;
+using namespace o2::framework;
+using namespace o2::header;
+
+BOOST_AUTO_TEST_CASE(test_checker_factory)
+{
+  std::string configFilePath =
+    std::string("json:/") + getenv("QUALITYCONTROL_ROOT") + "/tests/testInfrastructureGenerator.json";
```
knopers8: This test, `testTaskRunner` and `testInfrastructureGenerator` share the same config file, but if needed, I can create separate config files for them. Advantage: maybe clearer project, disadvantage: more files to mantain. What do you think?

Barthelemy: I suggest to keep a single one for the time being. Just rename it so that we understand it is not only for testInfrastructureGenerator.

Framework/test/testChecker.cxx
```diff
@@ -0,0 +1,71 @@
...
+
+  Checker checker{ "abcChecker", "abcTask", configFilePath };
+
+  BOOST_CHECK_EQUAL(checker.getInputSpec(), (InputSpec{ { "mo" }, "QC", "abcTask-mo", 0 }));
+  BOOST_CHECK_EQUAL(checker.getOutputSpec(), (OutputSpec{ "QC", "abcTask-chk", 0 }));
+
+  // This is maximum that we can do until we are able to test the DPL algorithms in isolation.
+  // When it is possible, we should try calling run() and init()
```
Barthelemy: Make it a "TODO" so that we can collect all the todos easily.

Framework/test/testWorkflow.cxx
Barthelemy: I am not sure to understand the uqestion. Let's discuss Monday.

Barthelemy: The test testWorkflow took 1.69 seconds on my Mac.

knopers8:
All comments should be addressed now.

knopers8:
Then maybe it is the time to enable the tests in CI again?

Barthelemy:
yes
I am not sure how to do that yet. Will take care of it.

PR #189, opened 2019-06-18
Add option to specify the config file in basic workflow

Barthelemy:


Barthelemy:
This is needed by Adam for the demo tomorrow morning. thus I will force push it. Sorry. (it is the same as for the readout workflow). @knopers8  you can review it nevertheless as it is better to catch a problem later than never.

knopers8:
It looks correct.

Framework/src/runBasic.cxx
```diff
@@ -124,3 +127,17 @@ WorkflowSpec defineDataProcessing(const ConfigContext& config)
 
   return specs;
 }
+
+// TODO merge this with the one from runReadout.cxx
```
knopers8: I wouldn't sweat it, because when we use the workflow merging, in runBasic and runReadout we will have data producers only (nothing to configure) and the config files will need to be passed explicitly to o2-qc-run-qc

Barthelemy: Ok, so I will not press this issue.

PR #203, opened 2019-06-28
[WIP] Changes to be compatible with ROOT 6.18 (QC-192)

Barthelemy:


Barthelemy:
Thanks @pzhristov

knopers8:
Don't forget about `Modules/Skeleton/.CMakeListsEmpty.txt`, it is the one that is used as a template for modules-configurator

Barthelemy:
@knopers8 good point !
I am checking it.

Barthelemy:
After the update of CMake no changes are needed any more.

PR #218, opened 2019-07-24
Fix formatting (O2-794)

Barthelemy:


knopers8:
This is all what wasn't following the clang-format rules? Surprisingly little.

Barthelemy:
I suspect that I had already done the exercise in the past.

PR #223, opened 2019-08-23
Fix newline in new module's script (QC-229)

Barthelemy:


Barthelemy:
@aphecetche I am trying to fix an issue on CC7. When I fix it, then it breaks on Mac. Have you ever had it working on both Mac and CC7 ? 
I am considering going back to the "if Mac then else if CC7 then" because it makes it a lot easier to maintain.

Barthelemy:
I have added a commit with a fix working on both CC7 and Mac. Let's see if Laurent has a more elegant fix.

aphecetche:
@Barthelemy I believe I had it working on both platforms at some point yes. But won't be able to work on it until tomorrow afternoon at the earliest. So if the fix is needed urgently, please merge.

Barthelemy:
we can wait, no worries.

aphecetche:
@Barthelemy I gave it a try, but can't get it working on both platform just yet. Must have dreamt it...
Anyway, don't want to spend too much time on this, so I'd propose to merge this one if it works.

Barthelemy:
@aphecetche Ok, I will retest it on Monday just to be sure and then merge.

PR #224, opened 2019-08-27
[QC-156] Access to conditions in CCDB inside Tasks

knopers8:


Framework/include/QualityControl/TaskInterface.h
```diff
@@ -74,17 +84,22 @@ class TaskInterface
...
+  TObject* retrieveCondition(std::string path, std::map<std::string, std::string> metadata = {}, long timestamp = -1);
+
   std::unordered_map<std::string, std::string> mCustomParameters;
 
  private:
   // TODO should we rather have a global/singleton for the objectsManager ?
   std::shared_ptr<ObjectsManager> mObjectsManager;
   std::string mName;
+  std::string mCcdbUrl = "";
+  std::shared_ptr<o2::ccdb::CcdbApi> mCcdbApi;
```
Barthelemy: Here we should probably keep only `mCcdbApi` to avoid confusion. S

Framework/include/QualityControl/TaskInterface.h
```diff
@@ -62,6 +70,8 @@ class TaskInterface
   /// Move assignment operator
   TaskInterface& operator=(TaskInterface&& other) /* noexcept */ = default; // error with gcc if noexcept
 
+  virtual void initializeBase() final;
```
Barthelemy: as this is only dealing with ccdb I would rename it to "loadCcdb".

doc/Advanced.md
```diff
@@ -117,6 +117,32 @@ Simply call `ObjectsManager::addMetadata(...)`, like in
 ```
 This metadata will end up in the CCDB.
 
+## Access conditions from the CCDB
```
Barthelemy: Maybe it should be more explicit that there is one instance of CCDB for the QC objects and a different CCDB instance for the rest and that we have two different URLs for them.

Framework/include/QualityControl/TaskInterface.h
knopers8: This is needed to initialize mCcdbApi in a initializaiton function, instead of constructor.

Framework/include/QualityControl/TaskInterface.h
knopers8: The idea was to put everything that needs to be initialized in the parent class there, even if now it is only ccdb api. That being said, I don't mind it being `loadCcdb`

doc/Advanced.md
knopers8: i have added a few more sentences, is it now clear?

Barthelemy: good

Framework/include/QualityControl/TaskInterface.h
knopers8: I got rid of it by having `TaskInterface::loadCcdb()` take the ccdb url as an argument

PR #225, opened 2019-08-28
[QC-230] Allow to specify a full path of a library

knopers8:
This is done, but I still need make sure that it works on a real FLP suite setup.


knopers8:
I confirmed that it works on a setup with FLP suite, but RPMs generated from this PR. It is not trivial to set up a working development environment - one has to find matching QC, O2, ROOT and GCC versions to these used for building QC RPMs. As usual, something doesn't compile because of some esoteric reasons, so one has to peek around in alidist history to find a commit which compiles.
It would be much easier to have a separate branch in alidist with tags which correspond to specific configurations that we use for FLP suite releases.

Barthelemy:
Thank you for taking the time to do that. I think that we are going to move in the direction you indicate with the new weekly tags of O2 and alidist.

doc/Advanced.md
```diff
@@ -165,6 +168,67 @@ At the moment, the description of the REST api can be found in this document : h
 
 To install and run the QCG locally, and its fellow process tobject2json, please follow these instructions : https://github.com/AliceO2Group/WebUi/tree/dev/QualityControl#run-qcg-locally
 
+## Custom QC object metadata
```
Barthelemy: I think that two sections are just here because of a merge. Only "Developing QC modules on a machine with FLP suite" should be here. Same in the TOC.

knopers8: oops, indeed

knopers8: done

PR #228, opened 2019-08-30
Update modules documentation (QC-174)

Barthelemy:


Modules/CMakeLists.txt
```diff
@@ -6,3 +6,4 @@ add_subdirectory(Example)
 add_subdirectory(Skeleton)
 add_subdirectory(TOF)
 add_subdirectory(EMCAL)
+
```
knopers8: With this newline, o2-qc-module-configurator.sh will add a next module after an empty line (as below), could you delete it please?
```
...
add_subdirectory(EMCAL)

add_subdirectory(NewDetector)
```

Barthelemy: thank you for spotting this !

PR #232, opened 2019-09-02
Tpc qc pid

wiechula:
This is a first very basic QC task mainly for demonstration at the moment. I will be used as a base for further developments.


Barthelemy:
Thank you for this, looking forward to review it. Could you reformat the code ? 
Instructions are [here](https://github.com/AliceO2Group/CodingGuidelines#check-files-formatting) and [here](https://github.com/AliceO2Group/CodingGuidelines#check-commits-formatting). Let me know if you need help.

wiechula:
Hi @Barthelemy , for some reason I get the following error using clang-format:
```
AML:53:11: error: unknown enumerated scalar              
Language: ObjC                                            
          ^~~~                                            
Error reading /data/Work/software/alicesw/O2-work/QualityControl/.clang-format: Invalid argument
```
Do you have an idea?

wiechula:
Funny, if I remove those those commits:
https://github.com/AliceO2Group/QualityControl/commit/925ca86163e0e623b8f7e19e2f1f81eacad03a8c
https://github.com/AliceO2Group/QualityControl/commit/c40570f35bd0ff6dcdd88778ed092003d86091a5
it works. Don't know if my clang-format is too old, or I'm missing language support...

wiechula:
The clang-format issue reported now is not corrected if I run it locally on my machine. Any idea? I can do it manually, of course, but then the next time I'll run into the same problem. Could it be the clang-format version?

Barthelemy:
@wiechula Thank you for putting time into trying this. It could indeed be the clang-format version. What is your version ? Just let it be as it is for now.

wiechula:
@Barthelemy The version I'm using is clang-format-3.9, since I could install it from the package manager. If this is too old, I can also install Clang via aliBuild.

wiechula:
Hi @Barthelemy ,

in the `o2` and `o2-dataflow` I find these errors:
```
[ERROR] Invalid detector name : XXXXXXXXX
    Placeholder 'MISC' will be used instead
    Note: list of permitted detector names :MISC DAQ GENERAL TST BMK CTP TRG DCS ITS TPC TRD TOF PHS CPV EMC HMP MFT MCH MID ZDC FT0 FV0 FDD ACO 


/mnt/mesos/sandbox/sandbox/sw/SOURCES/QualityControl/build_QualityControl_o2/0/Framework/test/testInfrastructureGenerator.cxx(70): [4;31;49mfatal error: in "qc_factory_remote_test": critical check workflow.size() == 4 has failed [6 != 4][0;39;49m
```

I'm not sure how they relate to my PR. Any suggestions?

Barthelemy:
@wiechula Sorry about that, I have now fixed it. 
You should still see thefirst  error about an invalid detector name but this is actually correct as we want to test this beheaviour.

Framework/src/runTPCQCPID.cxx
```diff
@@ -0,0 +1,122 @@
...
+
+///
+/// \file    runTPCQCPID.cxx
+/// \author  Jens Wiechula
+///
+/// \brief run TPC PID QC task, reading tracks from file for the moment
+///
+
+// IMPORTANT:
+// for some reason, if the 'customize' functions are below the other includes,
```
Barthelemy: Indeed. @knopers8 might be able to give more details. Is it something that should be better documented ?

Barthelemy:
The failures in the tests are unrelated.

Framework/CMakeLists.txt
```diff
@@ -108,6 +108,7 @@ set(EXE_SRCS
     src/runDataProducer.cxx
     src/runQC.cxx
     src/runBasic.cxx
+    src/runTPCQCPID.cxx
```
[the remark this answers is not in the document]
wiechula: Indeed this is what we would like to have eventually, to be independent of the input type. Since there is already a generic tree reader in o2 I asked @matthiasrichter to implement a generic configurable workflow for this.

Framework/src/runTPCQCPID.cxx
knopers8: The customize() functions have to be before `#include "Framework/runDataProcessing.h"`. That header basically checks if these functions are defined by user and if so, it invokes them. There is a trick with SFINAE expressions to do that.

knopers8:
For the future, can we keep the detector-specific executables and config files inside `Modules`?

Barthelemy:
@knopers8 correct !

wiechula:
Thanks for the comments @knopers8  and @Barthelemy .
I'll move the executable to our Module and make another PR.

PR #234, opened 2019-09-03
Add a detector name to MO and use it in path for CCDB (QC-233)

Barthelemy:


Barthelemy:
There is a DataID in O2 that I hesitated to use. The problem is that we don't want to limit ourselves to only detector codes. Typically we need a category for things without categories. Also something for the DAQ and for General plots. I will see if I can enforce to stay in a set of detectors + a few extras, possibly specified in the config.

knopers8:
Also, I cannot see objects being updated on QCG with that PR. Maybe we should wait until QCG supports the new storage structure before merging it?

Barthelemy:
you should see them, it works for me
are you looking at the right path ?

knopers8:
Ah, no, I can see it now, I wasn't looking at *qc*/<detector_name>/...

Barthelemy:
Ok, so, I validate in TaskRunner and substitute it if it is not valid. Unfortunately, DetID cannot be used outside O2 for the time being (https://alice.its.cern.ch/jira/browse/O2-843), so I hard-coded the list of detectors. 

I can't really unit test the logic of the validation because TaskRunner encapsulates Taskinterface and thus i can't "see" my detector name. I could test the output of the program but this is probably an overkill.

knopers8:
Thanks! Maybe it would be a good idea to update also the config files of detector modules.

Barthelemy:
yes, will do

Barthelemy:
I found only one, right ?

knopers8:
There is also Modules/TOF/toj.json

Barthelemy:
test failure is unrelated.

PR #237, opened 2019-09-10
Stabilize CCDB tests and update API (QC-221)

Barthelemy:
- Remove getListOfTasksWithPublications
- Enable all tests but testWorkflow
- Fix the tests


Barthelemy:
@knopers8 testWorkflows does not work for me but it is labelled "manual" . it is the sole test not working for me.

knopers8:
Ok, I will try to reproduce and fix it, in the meanwhile we can proceed with enabling the rest of them, I think.

PR #244, opened 2019-09-13
fix a warning in tpc

Barthelemy:


Barthelemy:
@MichaelLettrich I have the builds failing but cannot access the logs. Could you have a look ? thank you

Barthelemy:
@ktf According to @MichaelLettrich I should rather ask you about the missing logs.

ktf:
Sorry, I took the afternoon off to babysit. I understand the problem (which is there since ever,BTW) and I've a few ideas on how to properly fix it. For the moment I will clean up the build directory (sw) which will bring us back to a working state.

PR #245, opened 2019-09-16
Print version (QC-234)

Barthelemy:


Framework/src/TaskRunner.cxx
```diff
@@ -57,6 +57,9 @@ TaskRunner::TaskRunner(const std::string& taskName, const std::string& configura
     mCycleNumber(0),
     mTotalNumberObjectsPublished(0)
 {
+  // Log the version
```
knopers8: Can we put it at the beginnings of `InfrastructureGenerator::generateLocalInfrastructure()` and `InfrastructureGenerator::generateRemoteInfrastructure()` instead? These are usually the first QC functions invoked in the code. So if something breaks before we reach the ctor of `TaskRunner`, then we would still see the version number.

Barthelemy: excellent idea, I am preparing the change

Framework/src/InfrastructureGenerator.cxx
```diff
@@ -128,7 +128,8 @@ void InfrastructureGenerator::customizeInfrastructure(std::vector<framework::Com
 void InfrastructureGenerator::printVersion()
 {
   // Log the version
-  QcInfoLogger::GetInstance() << "QC version " << o2::quality_control::core::Version::getString() << infologger::endm;;
+  QcInfoLogger::GetInstance() << "QC version " << o2::quality_control::core::Version::getString() << infologger::endm;
+  ;
```
knopers8: why semicolon?

Barthelemy: thanks for noticing it. It is just a typo.

Barthelemy:
@ktf  here is a new instance of missing logs.

ktf:
Can you force push to retrigger a build? AFAICT the builder is working correctly now...

Barthelemy:
sure

Barthelemy:
the failing test is unrelated

PR #254, opened 2019-10-02
Convenience classes for correlation and trending

knopers8:


knopers8:
I will prepare the documentation in a separate PR covering both general post-processing and these classes.

knopers8:
You can run it by doing:
```
o2-qc-run-postprocessing --config json:///path/to/Framework/postprocessing.json --name ExampleTrend --rate 5
```
To see the results changing, run o2-qc-run-basic in parallel.

Framework/src/TTreeTrend.cxx
```diff
@@ -0,0 +1,167 @@
...
+
+void TTreeTrend::configure(std::string name, o2::configuration::ConfigurationInterface& config)
+{
+  mConfig = TTreeTrendConfig(name, config);
+}
+
+void TTreeTrend::initialize(Trigger, framework::ServiceRegistry& services)
+{
+  // Preparing data structure of TTree
+  mTrend = std::make_unique<TTree>(); // todo: retrieve last TTree, so we continue trending. maybe do it optionally?
```
Barthelemy: about your TODO : yes, optionnally but it is an important feature to consider.

Framework/test/testTrendingTask.cxx
```diff
@@ -0,0 +1,33 @@
...
+#define BOOST_TEST_DYN_LINK
+
+#include <boost/test/unit_test.hpp>
+
+using namespace o2::quality_control::postprocessing;
+
+BOOST_AUTO_TEST_CASE(test_factory)
+{
+
+  //  std::string configFilePath = std::string("json://") + getTestDataDirectory() + "testSharedConfig.json";
```
Barthelemy: test is empty ?

knopers8: Ah yes, this is a leftover that I should have erased. I am not sure how I could approach testing this without using a database (I guess we don't want to do it, because such tests tend to fail for too many random reasons).

Framework/src/TTreeTrend.cxx
knopers8: I treat it as one of the first features that should be added after the initial version is merged.

knopers8:
Do you have a better name for `TTreeTrend`? I don't like it too much. Maybe `TrendingTask` could work, but it gives an impression that it can be used *only* for trending

Barthelemy:
> Do you have a better name for `TTreeTrend`? I don't like it too much. Maybe `TrendingTask` could work, but it gives an impression that it can be used _only_ for trending

As it is _mostly_ used for trending, `TrendingTask` seems ok.

Framework/test/testTrendingTask.cxx
Barthelemy: indeed, on the other hand in testCcdbDatabase the connection is first tested and only then the test executed if I am not mistaken. Moreover, we use a directory named after the PID to avoid clashes between tests. Maybe this is something that can be done in a different PR, but I would not rule it out. In last resort, the test can be marked in CMake not to be ran in GH.

knopers8: OK, I will write a test. Actually I forgot to write a test for Reductors anyway, this is needed too.

knopers8:
I've added tests for Reductors and TrendingTask. If the CI passes and you have no other objections, I will merge this.

knopers8:
The TrendingTask test is set to be manual, because it might not work when the test is run on multiple places at the same time - race conditions etc.

Barthelemy:
> I've added tests for Reductors and TrendingTask. If the CI passes and you have no other objections, I will merge this.

agreed. I have just added a bit of ouptut to understand why CCDB test was failing.

knopers8:
Not sure what is wrong, maybe I messed something up in CMake. It seems ok though

PR #255, opened 2019-10-07
Common class factory using ROOT reflection

knopers8:
Also Checkers could use it, but I would move it after we merge https://github.com/AliceO2Group/QualityControl/pull/216 to avoid conflicts before that.


Barthelemy:
Very nice ! Agreed for the checkers. Error in o2-dataflow does not seem related.

PR #257, opened 2019-10-09
[QC-244] Clean up QC task monitoring metrics

knopers8:


Barthelemy:
Thank you for this, it has also bothered me in the past. There is 1 difference between the new and old implementation : the period of monitoring data was always 10s before and now it is tied to the cycle duration. I guess it is ok but I wanted to point it out.

knopers8:
I think that the 10s rate was only for the "QC_task_Rate_objects_published_per_10_seconds" metric. The rest of the metrics would be already published at the end of cycle.
Since also objects are published only at the end of cycle, I didn't see any point to keep this metric.

Framework/src/TaskRunner.cxx
```diff
@@ -398,17 +355,41 @@ void TaskRunner::finishCycle(DataAllocator& outputs)
...
     Output{ concreteOutput.origin,
             concreteOutput.description,
             concreteOutput.subSpec,
             mMonitorObjectsSpec.lifetime },
-    dynamic_cast<TObject*>(mObjectsManager->getNonOwningArray()));
+    dynamic_cast<TObject*>(array));
 
-  return 1;
+  mLastPublicationDuration = publicationDurationTimer.getTime();
+  return objectsPublished;
```
knopers8: Also, note this change. Now we don't count all MonitorObjects as one object anymore (just because they were published in one TObjArray).

PR #263, opened 2019-10-11
Advanced random data producer

knopers8:


knopers8:
That closing and reopening was accidental. 
What I wanted to say, is that I will add a monitoring metrics here as well.

PR #269, opened 2019-11-05
Include explicitly ConfigParamRegistry.h

ktf:
InitContext will drop implicit include soon.


ktf:
@Barthelemy could you tag once this actually compile?

Barthelemy:
@ktf sure, I will do

PR #270, opened 2019-11-06
Make sure to update test files

Barthelemy:
Test files where copied once during configuration but never again even if modified.


Barthelemy:
@aphecetche FYI

aphecetche:
@Barthelemy I know this one is merged, but instead of the ugly piece of CMake custom code, what we should be using for those cases is `configure_file(input output COPYONLY)` instead of `file(COPY)` and that should just works.
(just hit the same issue on something I'm working on with alo-o2...)

Barthelemy:
@aphecetche Thank you for the hint, I will create a ticket to update this ugly piece of CMake.

Barthelemy:
@aphecetche I have tried it and indeed it works ! The only downside is that it triggers a re-configuration with cmake each time one touches such a file. I guess that it is not a problem though.

PR #272, opened 2019-11-12
Add missing header

ktf:


ktf:
@Barthelemy: one more. Can you also tag and bump alidist?

PR #273, opened 2019-11-20
Added file reader executable

alice-mch:
The new executable reads data from a binary input file and sends the contents to QC via DPL.
The input data file must be filled with contiguous RDH blocks of 8 kB. Each block is read from the file and sent out via DPL.
The output format is the same as the one of the readout proxy executable, so that it should be possible to transparently use either of the two data providers without changing the QC side.

The code should be reviewed by some DPL expert, to make sure that there are no mistakes in the way the messages are initialized and created. Clang formatting is not applied yet.


Barthelemy:
Excellent !

Remaining to be done : 
- Review by @knopers8 (Piotr, back next week)
- Documentation in mark down files to make users aware (if needed I can take care of that)
- clang-format

Framework/src/runFileReader.cxx
```diff
@@ -0,0 +1,126 @@
...
+    if (mInputFile.fail()) {
+      if (mPrint) {
+        LOG(INFO) << "end of file reached";
+      }
+      free(buf);
+      return; // probably reached eof
+    }
+
+    // create the output message
+    auto msgOut = pc.outputs().make<char>(Output{"ROUT", "RAWDATA"}, RDH_BLOCK_SIZE);
```
knopers8: As a performance improvement, you can try avoiding memcpy by using `pc.outputs().adopt()` or `pc.outputs().adoptChunk()`. This way, DPL takes ownership of your buffer and deletes it by itself, when it is no longer needed.

Barthelemy:
Would it be possible to converge on that ? there are minor changes to be done before we merge. I can take care of the documentation if necessary.

aferrero2707:
@Barthelemy @knopers8 I have implemented the two suggestions, and tested them on my FLP. As far as I can tell, the code works as the previous version, however I have no idea how to test if `pc.outputs().adoptChunk()` was successful or not. Any idea? You should also check if the syntax is really correct, as I have never used such function so far...

knopers8:
`pc.outputs().adoptChunk()` doesn't return a variable which tells if it was succesful, but it should throw exceptions if it fails. The syntax seems correct. I will try it out running to confirm it works.

knopers8:
That works:
```
dd if=/dev/urandom of=filereadertest bs=8k count=10000
o2-qc-run-file-reader -b --infile filereadertest | o2-qc --config json://${QUALITYCONTROL_ROOT}/etc/readout.json -b
```

PR #276, opened 2019-11-25
[QC-258] Fix the problem with very short cycles in run-tpcpid

knopers8:


knopers8:
@wiechula @stheckel

knopers8:
Stefan confirmed that it fixes the problem with the cycles, merging.

PR #277, opened 2019-11-25
[WIP] Benchmark module

knopers8:
This is the code that used for benchmarks results for CHEP, I will have to clean it up and see if I can unify it with the @rpacholek 's code.


Barthelemy:
Is this PR still relevant ?

knopers8:
I should come back to it at some point, but for now we can close it.

PR #278, opened 2019-11-26
(QC-259) Add documentation for publisher and reshuffle

Barthelemy:


Barthelemy:
@knopers8 Not sure whether we should have the "dataProducerExample" stuff or if we should just paste the content in the README. 

Feel free to modify or comment. The aim is not to be accurate but to be easily understandable by the users.

doc/Advanced.md
```diff
@@ -28,6 +32,36 @@
 
 [← Go back to Modules Development](ModulesDevelopment.md) | [↑ Go to the Table of Content ↑](../README.md) | [Continue to Frequently Asked Questions →](FAQ.md)
 
+
+## Plugging the QC to an existing DPL workflow
+
+Your existing DPL workflow can simply be considered a publisher. Therefore, replace `o2-qc-run-producer` with your own workflow. 
+
+For example, if TPC wants to monitor the output of the workflow `o2-qc-run-tpcpid`, modify the config file to point to the correct data and do :
```
knopers8: Maybe "For example, if TPC wants to monitor the output `{"TPC", "CLUSTERS"}` of the workflow..."? It could give a wrong impression, that a workflow may have only one output

doc/Advanced.md
```diff
@@ -28,6 +32,36 @@
...
+Your existing DPL workflow can simply be considered a publisher. Therefore, replace `o2-qc-run-producer` with your own workflow. 
+
+For example, if TPC wants to monitor the output of the workflow `o2-qc-run-tpcpid`, modify the config file to point to the correct data and do : 
+```
+o2-qc-run-tpcpid | o2-qc --config json://${QUALITYCONTROL_ROOT}/etc/tpcQCPID.json
+```
+
+## Writing a DPL data producer 
+
+For your convenience, and although this is not a QC problem, we would like to document how to write a simple data producer in the DPL. The DPL documentation can be found [here](https://github.com/AliceO2Group/AliceO2/blob/dev/Framework/Core/README.md) and for questions please head to the [forum](https://alice-talk.web.cern.ch/).
```
knopers8: Maybe "For your convenience, and it does not lie within the QC scope, we would..."? It sounds less harsh.

doc/FAQ.md
```diff
@@ -49,6 +51,17 @@ For AliceO2 libraries see the next question.
 
 Add the library name to the list `O2_LIBRARIES_NAMES` in [FindAliceO2.cmake](../cmake/FindAliceO2.cmake)
 
+## Run 
+
+### Why are my QC processes using 100% CPU ? 
+
+When running `o2-qc` or other qc binaries, the system will show that the processes use 100% of the CPU. This is due to the DPL default rate for data source devices.
```
knopers8: not only data sources, devices with inputs also have an event loop which runs with this rate

knopers8:
What is just slightly risky, is that the function names inside DataProducer.h and DataProducerExample.h are the same. If that trend continues (because users will develop more of these), we might end up with some name conflicts if many of these headers are included in one file.

Barthelemy:
Thank you , I will apply your proposals and I agree that it is better to change the name of the functions.

doc/Advanced.md
Barthelemy: ok

Barthelemy:
@ktf @MichaelLettrich Here as well the builds are failing without logs

ktf:
I have restarted the workers, the CI cache got corrupted.

PR #280, opened 2019-12-05
[QC-260] Support EndOfStream in QC

knopers8:
DataProducer can send EndOfStream now. Todo:
- [x] fix the problem with lack of EOS in TaskRunner's timer input 
- [x] implement EOS callback in TaskRunner and perhaps CheckRunner, make sure that MOs are sent and checked for the last time


knopers8:
Depends on https://github.com/AliceO2Group/AliceO2/pull/3031

Barthelemy:
All good !

PR #292, opened 2020-01-07
[WIP] Use ccdbApi methods *AsTFileAny 

Barthelemy:


Barthelemy:
TODO: avoid making two requests. This will require a new method in CcdbApi.

PR #294, opened 2020-01-10
[WIP] more output to understand the problem with failing ccdb test

knopers8:
@Barthelemy It seems that the problem is quite random, but happens only when listing the "qc" directory, which is probably the largest:
```
2020-01-10 04:43:54.281127     getPublishedObjectNames of task qc
Running 5 test cases...
unknown location(0): [4;31;49mfatal error: in "ccdb_getobjects_name": std::out_of_range: basic_string::substr: __pos (which is 11) > this->size() (which is 10)[0;39;49m
```
I think this is the place where the exception gets thrown, let's see what are the conditions when it happens.

### DIFF
--- a/Framework/src/CcdbDatabase.cxx
+++ b/Framework/src/CcdbDatabase.cxx
@@ -282,9 +282,11 @@ std::vector<std::string> CcdbDatabase::getPublishedObjectNames(std::string taskN
       } catch (std::out_of_range& ex) {
 
         ILOG(Info) << "std::out_of_range exception caught" << ENDM;
-        ILOG(Info) << "line: '" << line << "'";
-        ILOG(Info) << "taskNameEscaped: '" << taskNameEscaped << "'";
-        ILOG(Info) << "objNameStart: '" << objNameStart << "'";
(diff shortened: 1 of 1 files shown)


knopers8:
```
2020-01-10 08:33:29.172484     getPublishedObjectNames of task qc
2020-01-10 08:33:30.679828     std::out_of_range exception caught
Running 5 test cases...
unknown location(0): [4;31;49mfatal error: in "ccdb_getobjects_name": std::out_of_range: basic_string::substr: __pos (which is 11) > this->size() (which is 10)[0;39;49m
/mnt/mesos/sandbox/sandbox/sw/SOURCES/QualityControl/build_QualityControl_o2-dataflow/0/Framework/test/testCcdbDatabase.cxx(68): [1;36;49mlast checkpoint: "ccdb_getobjects_name" test entry[0;39;49m
2020-01-10 08:33:30.681511     line: '"path":"",'taskNameEscaped: 'qc'objNameStart: '11'CCDB backend selected
```
For some reason the value of "path" is empty.

PR #300, opened 2020-01-31
Fix TPC PID json file: add checker

stheckel:
A checker is added to the tpcQCPID.json file, because this is needed for the output to be stored and send to the QCG.


Barthelemy:
I will merge once the tests pass.

PR #305, opened 2020-02-06
Replacing DataAllocator::adopt by snapshot to prepare for O2 PR 2887

matthiasrichter:
The RootContext will be removed in PR AliceO2Group/AliceO2#2887 and also
DataAllocator::adopt using the deprecated RootContext will be removed.
Method has been replaced with either `snapshot` or `make`.


knopers8:
@Barthelemy The ccdb test started to fail again.
The changes are surely unrelated to this test, I will merge.

PR #308, opened 2020-02-11
updates in TPC PID, json file for piping of RecoWorkflow

tklemenz:
Minor changes in the TPC PID part and added tpcQCPID_direct.json that can be used to pipe the tpc-reco-workflow into a QC workflow.


knopers8:
Is this ready to be merged? I am making sure, because you marked it as a draft, but it looks quite complete.

tklemenz:
yes, this can be merged. thanks!

PR #311, opened 2020-02-13
Fix broken compilation on macOS

mfasDa:
Fix broken compilation introduced in #286


knopers8:
Thanks a lot, I will merge it as soon as it passes CI

PR #313, opened 2020-02-13
TPC QC Tracks workflow including json configs

stheckel:
o This is the QualityControl part of the TPC QC task for the tracks
o It implements the workflow for the O2 TPC QC Tracks task
o It can be run with the also provided json configs, either:
  - using the availble TPC track reader and the sampled json
  - piped to a TPC reco workflow using the direct json

### DIFF
--- a/Modules/TPC/src/Tracks.cxx
+++ b/Modules/TPC/src/Tracks.cxx
@@ -17,7 +17,7 @@
 #include <TCanvas.h>
 #include <TH1.h>
 #include <TH2.h>
-#include <gsl/span>
+//#include <gsl/span>
 
 // O2 includes
 #include "Framework/ProcessingContext.h"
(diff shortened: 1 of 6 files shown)


Modules/TPC/src/Tracks.cxx
```diff
@@ -0,0 +1,95 @@
...
+
+void Tracks::startOfCycle()
+{
+  QcInfoLogger::GetInstance() << "startOfCycle" << AliceO2::InfoLogger::InfoLogger::endm;
+}
+
+void Tracks::monitorData(o2::framework::ProcessingContext& ctx)
+{
+  using TrackType = std::vector<o2::tpc::TrackTPC>;
+  auto tracks = ctx.inputs().get<TrackType>("inputTracks");
```
wiechula: Here, it would be better to change it to:
```c++
using TracksType = gsl::span<o2::tpc::TrackTPC>;
const auto tracks = ctx.inputs().get<TracksType>("inputTracks");
```
and add
```c++
#include <gsl/span>
```
See discussion in https://alice.its.cern.ch/jira/browse/O2-1017.

@tklemenz : Please also consider this and the comment below for the PID qc task.

stheckel:
Hi Jens,
thanks for your input. I have implemented all the changes, but unfortunately the part with gsl/span does not work. It compiles (together with the corresponding changes in the O2 task), but when running, I get errors like:
`[27195:QC-TASK-RUNNER-Tracks]: [17:59:16][ERROR] Exception caught: Inconsistent serialization method for extracting span`
I get these errors both when running in sampled and in direct mode. In the end, the histograms are published, but empty. Looking at the printouts, I do not even get the `monitorData` messages, so it seems the issue is already in getting the TPC tracks in the monitorData function within the QC workflow.
The corresponding changes in O2 you can find [in my funcProcessTracks branch](https://github.com/stheckel/AliceO2/tree/funcProcessTracks), although, as stated, I think the issues is here in the QC.
Cheers, Stefan

stheckel:
Please note, that this code now does not compile without the corresponding change in O2. Therefore, I opened also there a [draft pull request](https://github.com/AliceO2Group/AliceO2/pull/2972).

stheckel:
Dear all,
following the discussions on [this ALICE talk ticket](https://alice-talk.web.cern.ch/t/error-using-gsl-span-instead-of-a-vector-for-the-tpc-qc-workflow/498), for the moment I switched back to the usage of a vector instead of gsl/span. This can be changed as soon as the required extension to the RootTreeReader is available. Therefore the changes in the O2 part are also not needed at the moment and I will keep the corresponding PR open as draft until I switch again to gsl/span.
@wiechula, please state, if this is fine with you, then I will mark this PR here as ready for review.
Cheers,
Stefan

stheckel:
Dear @Barthelemy, dear @knopers8,
I would like to ask you to merge this commit.
Thanks and cheers,
Stefan

Barthelemy:
Sure !

PR #318, opened 2020-02-27
Simplify the cmake to copy test files (QC-278)

Barthelemy:


Barthelemy:
@aphecetche

Barthelemy:
The only downside is that it triggers a re-configuration with cmake each time a test file is modified. Not a big deal I presume.

PR #319, opened 2020-02-27
Add info about repo_cleaner of ccdb-test

Barthelemy:


knopers8:
WIP fails for no reason, Travis doesn't respond, but this is just documentation - merging (thanks for this!)

PR #321, opened 2020-02-27
[EMCAL-526] Min and Max Ampl per cell added

cterrevo:


mfasDa:
@Barthelemy @cterrevo There is a PR to be merged in AliceO2 (https://github.com/AliceO2Group/AliceO2/pull/3035) that will break compilation of QualityControl. The PR is ready and will be merged soon. Either this PR must be merged now or it must wait for the adaption of the fix adapting the EMCAL component to the fix mentioned above.

mfasDa:
Depends on https://github.com/AliceO2Group/QualityControl/pull/326

mfasDa:
@knopers8 Even though a new build is triggered there will be an error in the PR checker due to a change in AliceO2 that needs adaption in QualityControl. The fix is #326 - please handle this with priority!

Barthelemy:
@mfasDa This should have been better coordinated. We now have build errors in all O2 PRs. I don't see a reason to force-merge during the weekend something that could have been done smoothly this week in a coordinated way. 

We now wait for the CI in #326  to make sure that we are in a stable state.

PR #322, opened 2020-02-28
Remove doc about info service

Barthelemy:


Barthelemy:
@graduta

graduta:
@Barthelemy Instead of simply replacing, could we add documentation for the ServiceDiscovery so that I can link to it from my side?

Barthelemy:
@graduta This would be on @awegrzyn I guess. I propose to carry on with two separate PRs.

awegrzyn:
Could you clarify what shall I document? The ServiceDiscovery class is very well documented already: https://github.com/AliceO2Group/QualityControl/blob/master/Framework/include/QualityControl/ServiceDiscovery.h

graduta:
From the looks of it (https://alice-talk.web.cern.ch/t/qc-task-error-servicediscovery-timeout-was-reached/505 first post), users are a bit confused on what is Consul if the Service Discovery timesout due to Consul not being installed or connected, will that affect their tasks?

I am thinking it would be good to have a few lines in which we describe what is Service Discovery - Consul and how we use it and when it should be ignored

awegrzyn:
sure, but at first I'd like to understand why they get such error? Are they using FLP suite? How come consul is not deployed? If they are not using FLP suite, which instructions they are using to get Consul installed?

graduta:
From my understanding they are not using FLP Suite. 
In QCG case they simply copied the whole (config-default.js) file whitout trying to understand what each component does so they think these errors are related to their problems which is not the case

I will also update the documentation on the GUI so that it is clear

graduta:
@Barthelemy Do you have more information about Adam's question on users using FLP suite or not?

Barthelemy:
I don't think that they use FLP suite. They installed following the instructions for developers in github repo for QC.

PR #323, opened 2020-02-28
[WIP] TPC QC Tracks: now using gsl/span

stheckel:
This is the QC workflow part of the transition from a vector to gsl/span for the handing over of the TPC tracks from the QC workflow to the O2 TPC QC Tracks task.

### DIFF
--- a/Modules/TPC/src/Tracks.cxx
+++ b/Modules/TPC/src/Tracks.cxx
@@ -17,7 +17,7 @@
 #include <TCanvas.h>
 #include <TH1.h>
 #include <TH2.h>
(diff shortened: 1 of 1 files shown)


Barthelemy:
This PR has been created a long time ago. Is it still something you plan on finishing ?

stheckel:
Hi Barth,
indeed, at some point we want to finish this, but it has low priority and therefore never happend. Actually, the PR of the O2 part of this change has been closed a while ago. I will close also the QC part now and open a new PR, once I found the time to work on this.
Cheers, Stefan

PR #324, opened 2020-02-28
Indicate how to use multiple direct data sources

knopers8:


knopers8:
Documentation only, merging

PR #326, opened 2020-02-28
[EMCAL-526, EMCAL-507] Adapt digits QC to timeframe handling

mfasDa:
- Digits publisher adapted to read trigger records branch
- DigitQC looping over trigger records and digits in span
  defined by range in trigger record
- DigitsQC reading gsl::span instead of std::vector profiting
  from copyless access to data vectors

### DIFF
--- a/Modules/EMCAL/etc/digits.json
+++ b/Modules/EMCAL/etc/digits.json
@@ -32,7 +32,7 @@
           "maxNumberCycles": "-1",
           "dataSource": {
             "type": "direct",
-            "query": "emcal-digits:EMC/DIGITS/0"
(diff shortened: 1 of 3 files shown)


mfasDa:
Depends on https://github.com/AliceO2Group/AliceO2/pull/3035

mfasDa:
@Barthelemy @knopers8 The commit in AliceO2 is now merged. So this PR should be merged first as otherwise the PR checks will fail.

knopers8:
Thanks, I will merge it as soon as the CI is green.

knopers8:
`o2-dataflow` passed. `o2` failed, but in testCcdbApi - clearly not related. However, the failure of O2 tests blocked compilation of QC - for the future, maybe we could continue building it even these fail.

Force-merging - the results `o2-dataflow` indicate that it should fix the issue which we are tackling here.

PR #327, opened 2020-03-02
Do not load the same streamer info several times (QC-279)

Barthelemy:


knopers8:
I wonder why there are duplicate entries in `streamerinfos.root` in the first place. Could it be, that they are somehow different, but they just have common names and versions?

Barthelemy:
@knopers8 I am not sure myself but when I looked at them they were the same. It could be that they come from different inheritance branches.

PR #328, opened 2020-03-02
[EMCAL-524] Fixes in stand-alone workflow

mfasDa:
- Adapt workflow definition with publisher to
  runQC.cxx
- Set location in the configuration for digits
  to "local" in order to be able to run the stand-
  alone workflow


mfasDa:
CCDB error unrelated

PR #329, opened 2020-03-03
Clean up and addition of detector name (QC-282) (OGUI-367)

Barthelemy:


Framework/include/QualityControl/QualityObject.h
```diff
@@ -61,13 +63,13 @@ class QualityObject : public TObject
   /// @return the quality of the object
   ///
   void updateQuality(Quality quality);
-  Quality getQuality() const;
+  [[nodiscard]] Quality getQuality() const;
```
knopers8: My personal opinion is that these just pollute the code without any real benefits, if you think it is justified, I am fine with it.

Barthelemy: I think that I agree, I'll remove them.

PR #330, opened 2020-03-03
[QC-270] Integrating Post-processing with AliECS

knopers8:


Framework/src/runPostProcessingOCC.cxx
```diff
@@ -0,0 +1,216 @@
...
+class PostProcessingOCCStateMachine : public RuntimeControlledObject
+{
+ public:
+  PostProcessingOCCStateMachine(std::string name, std::string configPath, double period = 1.0)
+    : RuntimeControlledObject("Post-processing task runner"), mName(name), mConfigPath(configPath), mPeriod(period)
+  {
+    mRunner = std::make_unique<PostProcessingRunner>(name, configPath);
+  }
+
+  // In all of the methods below we handle exceptions, so we can go into error state - OCC won't do that for us.
```
Barthelemy: Maybe this is something we could discuss at some point with Teo because it makes the code a bit less fancy than it could.

doc/PostProcessing.md
```diff
@@ -93,16 +93,37 @@ Each of the three methods can be invoked by one or more triggers. Below are list
 
 ## Running it
 
-The post-processing tasks can be run by using the `o2-qc-run-postprocessing` application. At the moment it does not use DPL nor OCC library, which means, that it cannot be integrated with AliECS yet, i.e. cannot be run from GUI within an FLP Suite.
+The post-processing tasks can be run by using the `o2-qc-run-postprocessing` application (only for development) or with `o2-qc-run-postprocessing-occ` (both development and production).
 
-To run the basic example, use the command below. The `--config` parameter should point to the configuration file. The `--rate` parameter specifies how often the specified triggers should be checked.
+To run the basic example, use the command below. The `--config` parameter should point to the configuration file. The `--period` parameter specifies how often the specified triggers should be checked.
```
Barthelemy: maybe say what is the unit of the period.

CMakeLists.txt
```diff
@@ -97,6 +97,7 @@ find_package(CURL REQUIRED)
 find_package(GLFW NAMES glfw3 CONFIG)
 find_package(FairMQ REQUIRED)
 find_package(FairLogger REQUIRED)
+find_package(Occ REQUIRED)
```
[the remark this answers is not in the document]
knopers8: I did that in advance https://github.com/alisw/alidist/pull/2094

Framework/src/PostProcessingRunner.cxx
```diff
@@ -26,39 +25,25 @@ namespace o2::quality_control::postprocessing
...
-    ILOG(Error) << "exception : " << e << ENDM;
-    throw;
-  } catch (...) {
-    std::string diagnostic = boost::current_exception_diagnostic_information();
-    ILOG(Error) << "Unexpected exception, diagnostic information follows:\n"
-                << diagnostic << ENDM;
-    throw;
-  }
+  mConfigFile = ConfigurationFactory::getConfiguration(mConfigPath);
+  mConfig = PostProcessingConfig(mName, *mConfigFile);
```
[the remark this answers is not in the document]
knopers8: It will throw. For `o2-qc-run-postprocessing` exceptions will reach main and the application will return, for `o2-qc-run-postprocessing-occ` the state machine will go into ERROR state.

Framework/src/runPostProcessingOCC.cxx
knopers8: Yes, I have to think if there is any case when exceptions should cause transition to ERROR state, now I can't think of any.

PR #333, opened 2020-03-05
[QC-290] Test QualityObject

knopers8:
There was just an empty test file.
Also giving more appropriate argument name to the `DatabaseInterface::retrieveQO` and `DatabaseInterface::retrieveQOJson` methods.


knopers8:
Note that testCcdbDatabase.cxx is turned off (https://github.com/AliceO2Group/QualityControl/pull/307) as it was sometimes randomly failing.

PR #334, opened 2020-03-05
Unify retrieve ccdb functions

Barthelemy:
This is a general cleanup but is also needed by the gui to access both MO and QO in a uniform way.

### DIFF
--- a/Framework/src/CcdbDatabase.cxx
+++ b/Framework/src/CcdbDatabase.cxx
@@ -130,9 +130,8 @@ void CcdbDatabase::storeMO(std::shared_ptr<o2::quality_control::core::MonitorObj
   ccdbApi.storeAsTFile(mo.get(), path, metadata, from, to);
 }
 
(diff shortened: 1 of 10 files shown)


Framework/src/CcdbDatabase.cxx
```diff
@@ -146,22 +145,63 @@ std::shared_ptr<core::MonitorObject> CcdbDatabase::retrieveMO(std::string taskNa
...
+std::string CcdbDatabase::retrieveJson(std::string path, long timestamp)
+{
+  auto tobj = retrieveTObject(path, timestamp);
+  if (tobj == nullptr) {
+    return std::string();
+  }
+  TObject* toConvert;
+  if (tobj->IsA() == MonitorObject::Class()) {
+    std::shared_ptr<MonitorObject> mo = std::dynamic_pointer_cast<MonitorObject>(tobj);
+    toConvert = mo->getObject();
```
knopers8: I don't think I see `toConvert` deleted later

Barthelemy: fixed

PR #337, opened 2020-03-09
Add indirect header which came with Data Sampling

knopers8:
I plan to post a header cleanup in Data Sampling which should reduce the compilation time a bit. These two headers were previously brought by DataSampling, soon they won't be.


ktf:
Merging as trivial.

PR #340, opened 2020-03-10
Adding a basic Cluster and Track QC task

tomas-herman:
I added a basic Cluster and Track QC tasks together with a data producers which read MC files and feed it to the QC continuously. They directly follow from the basic digit QC task.

### DIFF
--- a/Modules/MFT/include/MFT/BasicTrackQcTask.h
+++ b/Modules/MFT/include/MFT/BasicTrackQcTask.h
@@ -26,10 +26,10 @@ using namespace o2::quality_control::core;
 namespace o2::quality_control_modules::mft
 {
 
-/// \brief Example Quality Control DPL Task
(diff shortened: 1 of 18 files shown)


Modules/MFT/include/MFT/BasicTrackQcTask.h
```diff
@@ -0,0 +1,56 @@
...
+// Quality Control
+#include "QualityControl/TaskInterface.h"
+
+using namespace o2::quality_control::core;
+
+namespace o2::quality_control_modules::mft
+{
+
+/// \brief Example Quality Control DPL Task
+/// It is final because there is no reason to derive from it. Just remove it if needed.
```
Barthelemy: Coudl you update the doxygen ?

tomas-herman:
Thanks for the comments, all should be implemented.

PR #342, opened 2020-03-12
Add possibility to set up ConfigurableParam from the command line

wiechula:


wiechula:
We were wondering how to configure e.g. limits for the checkers or adapt histogram ranges if possible. What we use typically in O2 for this is ConfigurableParam. This would add it to the qc, such that you can modify parameters from the command line.

Barthelemy:
Nice, I did not know about it. 
Do I understand correctly that the user can pass the parameters on the command line or in a config file and then it feeds a file called `o2qc_configuration.ini` ? 
What happens to this file then ? do you use it in your module ? 

By the way, I should make clear that at P2 there will be a configuration system (based on Consul) and thus your parameters will be editable and settable there (there should be a GUI as well). I don't think that it makes a problem to have both systems available, one for your development and one in commissioning/prod.

knopers8:
Wouldn't it be convenient for you to put these parameters together with the rest of QC tasks and checks configuration? There are "taskParameters" inside the task configuration structure, which are then accessible as a protected member `mCustomParameters` inside TaskInterface.
```
      "QcTask": {
        "active": "true",
        "moduleName": "QcSkeleton",
        "detectorName": "TST",
        "cycleDurationSeconds": "10",
        "maxNumberCycles": "-1",
        "dataSource_comment": "The other type of dataSource is \"direct\", see basic-no-sampling.json.",
        "dataSource": {
          "type": "dataSamplingPolicy",
          "name": "tst-raw"
        },
        "taskParameters": {
          "myOwnKey": "myOwnValue"
        },
        "location": "remote"
      }
```
Simliarly, you shoudl be able to add "checkParameters" to a Check configuration.

wiechula:
Hi @Barthelemy , @knopers8 ,
thanks for the quick feedback!
I don't insist on configurableParam. I'm only used to it and was not aware of other methods.

I thought it is a nice feature. For me using parameters inside a json file is also fine in principle. One advantage of ConfigurableParam is that the configuration can also transparently be stored / read from CCDB. You can check the documentation here:
https://github.com/AliceO2Group/AliceO2/blob/dev/Common/SimConfig/doc/ConfigurableParam.md

@Barthelemy : ConfigurableParam offers the possibility to read parameters from CCDB, json, ini file (which can be also passed on the command line), or command line parameters with some hierarchy. The full configuration of the parameters is stored in the file {{o2qc_configuration.ini}}, yes, that you have it for later verification or to play back the configuration.
Inside your class you simply access your ConfigurableParam class as singleton, not the ini file. The commands in DPL I added take care to set the values appropriately in the singleton.

My question would be if it makes sense to have different ways of configuring parameters. It seems to me that for simulation / reconstruction, the standard now is ConfigurableParam. For QC it seems to be the json file and later on in addition some kind of configuration system (which by the way I think could also be interfaced to ConfigurableParam). From a user point of view I would prefer to have one system. But this is just a suggestion. As I wrote I would not insist and you can take this PR as a basis for discussion.

Cheers,
Jens

Barthelemy:
@wiechula I agree on the fact that a single system is better for the users (everyone actually). We are not very strongly attached to a system rather than another but we follow what WP8 decides in matters of Configuration. I think that you made a nice argument about the ConfigurableParam. 

I will ask @vascobarroso as WP8 convener to chime in and share his thought on that.

vascobarroso:
@wiechula, as mentioned by @Barthelemy, the plan in production is to store configuration parameters in Consul which are pushed (not pulled) to the tasks by AliECS during the CONFIGURE transition (or in RUNNING for flash configurations that do not affect/change the physics data). The idea is to keep CCDB reserved for conditions and calibration data.

@teo touches a bit this topic in his [ALiECS presentation of today](https://indico.cern.ch/event/888263/contributions/3757669/attachments/2005148/3348559/2020-03-18-AliECS-update.pdf) at the FLP/PDP Plenary, mainly what is done and what remains to be done. 

There is another scenario which is when people run in standalone mode without AliECS (mainly for development/benchmark/test purposes). For this indeed we currently have different solutions, we should make an effort to unify. Once things settle a bit more I'll organise a discussion on this topic.

wiechula:
@Barthelemy , @vascobarroso , thanks for the feedback.
So to me it seems that we need 2 systems. One for online mode using consul, one for development and offline mode using something else. A discussion would be welcome.
Does anything speak against merging this solution for the moment?

knopers8:
I would propose to wait until we have a discussion. Once we merge it and people start using it, it will be harder to change it or remove if needed.
It seems that there is a need to agree on a unified system (or two, but with seamless transitiion between them). It should be easier to converge if we don't add any more complexity now.

wiechula:
Ok, this is sounds good. I think it would be good to have this discussion soon, to progress with the developments.

Barthelemy:
Closing until a decision is made.

PR #346, opened 2020-03-16
fix beautify (QC-299)

Barthelemy:


Barthelemy:
Tested a variety of cases (1-2 tasks with 1-2 objects connected in different combinations to a check).

doc/ModulesDevelopment.md
```diff
@@ -16,11 +16,11 @@
...
       * [Modification of a Task](#modification-of-a-task)
       * [Check](#check)
-        * [Configuration](#configuration)
-        * [Implementation](#implementation)
+         * [Configuration](#configuration)
+         * [Implementation](#implementation)
       * [Committing code](#committing-code)
 
-<!-- Added by: bvonhall, at:  -->
+<!-- Added by: barth, at: Lun 16 mar 2020 13:18:24 CET -->
```
knopers8: It looks like you have some clever way to update the table of contents automatically. How do you do that?

Framework/src/Check.cxx
```diff
@@ -122,6 +122,13 @@ void Check::initConfig(std::string checkName)
   mCheckConfig.detectorName = checkConfig.get<std::string>("detectorName", "DET");
   mLatestQuality->setDetectorName(mCheckConfig.detectorName);
 
+  // Determine whether we can beautify
+  // See QC-299 for details
+  if (numberOfTaskSources > 1) {
+    mBeautify = false;
+    ILOG(Warning) << "Beautification disabled because more than one source is used in this Check (" << mCheckConfig.checkName << ")" << ENDM;
```
knopers8: I am not exactly sure, but `Check::initConfig()` is probably invoked even during workflow generation. If so, then this warning will be visible in each device in a workflow, so we might unnecessarily pollute the logs. If that happens indeed, could you put this to `Check::init()`, which AFAIK should be invoked only by an initializing CheckRunner.

doc/ModulesDevelopment.md
Barthelemy: I download this software : https://github.com/ekalinin/github-markdown-toc

And then run it on the file.

Framework/src/Check.cxx
Barthelemy: good call, changing it

Barthelemy: I have also moved the rest of the logging from the function. I understand now why we were getting this info multiple times in the console.

knopers8: Nice, thanks!

PR #349, opened 2020-03-18
update monitoring metrics

Barthelemy:


Framework/src/TaskRunner.cxx
```diff
@@ -379,11 +382,12 @@ void TaskRunner::publishCycleStats()
   double totalDurationActivity = mTimerTotalDurationActivity.getTime();
 
   // monitoring metrics
-  mCollector->send({ mNumberBlocks, "QC_task_Numberofblocks_in_cycle" });
+  mCollector->send({ mNumberBlocks, "QC_task_Numberofblocks_received_in_cycle" });
```
knopers8: Isn't "blocks" outdated anyway? It might be easier to understand if we use "messages"

Barthelemy: yes, updating.

Barthelemy: done

PR #350, opened 2020-03-18
Add task and checker for TOF Error Words in Raw Data

njacazio:
- Add Counter Class
- Add Counter class for TOF Compressed data
- Add Task/Checker for Counters on TOF Compressed data
- Add Json configuration for TOF Compressed data counters QC


njacazio:
Dear @Barthelemy , I think we are ready for review ;)

Barthelemy:
@njacazio Ok, great, I am going to review it now. Meanwhile, could you fix the warning/error ?

njacazio:
Dear @Barthelemy , indeed there is a series of warnings blocking the PR.
They are of this type:
```
Diagnostics.h:97:7: error: 'o2::quality_control_modules::tof::Diagnostics' has a field 'o2::quality_control_modules::tof::Diagnostics::mRDHCounter' whose type uses the anonymous namespace [-Werror=subobject-linkage]
```
I must admit I don't know an easy solution to avoid this, do you have any idea?

Modules/TOF/include/Base/Counter.h
```diff
@@ -0,0 +1,119 @@
...
+
+namespace o2::quality_control_modules::tof
+{
+
+/// \brief Class to count events
+/// \author Nicolo' Jacazio
+template <typename Tc, Tc size, const TString* names>
+class Counter
+// : TObject /*final*/
+// todo add back the "final" when doxygen is fixed
```
Barthelemy: you can remove this , it is now ok

Modules/TOF/include/TOF/Diagnostics.h
```diff
@@ -0,0 +1,131 @@
...
+
+} // namespace counters
+
+using namespace counters;
+
+/// \brief TOF Quality Control class for Decoding Compressed data for TOF Compressed data QC Task
+/// \author Nicolo' Jacazio
+class Diagnostics /*final*/
+  : public DecoderBase
+// todo add back the "final" when doxygen is fixed
```
Barthelemy: idem

Modules/TOF/src/TOFCheckCompressedCounter.cxx
```diff
@@ -0,0 +1,98 @@
...
+        result = Quality::Medium;
+      }
+    }
+  }
+  return result;
+}
+
+std::string TOFCheckCompressedCounter::getAcceptedType() { return "TH1F"; }
+
+void TOFCheckCompressedCounter::beautify(std::shared_ptr<MonitorObject> mo, Quality checkResult)
```
Barthelemy: Have you retested the beautify since v0.23 ?

Modules/TOF/tofdiagnostics.json
```diff
@@ -0,0 +1,77 @@
...
+              "type": "Task",
+              "name": "TaskDiagnostics",
+              "MOs": ["RDHCounterCrate0"]
+          }]
+      }
+    }
+  },
+  "dataSamplingPolicies": [
+    {
+      "id": "pippo",
```
Barthelemy: I am sure Pippo will be honoured !

Modules/TOF/include/Base/Counter.h
njacazio: Ok, done

Modules/TOF/include/TOF/Diagnostics.h
njacazio: Ok, done

Modules/TOF/src/TOFCheckCompressedCounter.cxx
njacazio: No, not yet, I will try to recompile with latest master and see whether it has an effect

Barthelemy:
I don't know what is this error but I am having a look

Modules/TOF/tofdiagnostics.json
njacazio: My bad ;) Changed!

Barthelemy:
I am afraid I don't know where this warning is coming from. I read a lot of things on the web but it is not clear to me how it applies to your code.

njacazio:
Yes, I'm kind of in the same condition.. Someone suggests of moving implementation to a separate file but this has no effect.. I can try to solve the problem with a dummy class in order to simplify a bit the situation.

njacazio:
I am afraid I don't understand the problem here.. Anyway, it seems related to the template form that I used.
What I could do is remove the template and use a standard class, something like:
```namespace o2::quality_control_modules::tof
{

/// \brief Class to count events
/// \author Nicolo' Jacazio
// template <typename Tc, Tc size, const TString* names>
class Dummy
{
 public:
  /// \brief Constructor
  Dummy() = default;
  /// Destructor
  ~Dummy() = default;
  UInt_t Size() const { return size; }

 private:
  /// Containers to fill
  static const UInt_t size = 100;
  TString names[size] = { "" };
  uint32_t counter[size] = { 0 };
};```

Barthelemy:
that should work

njacazio:
Ok, but this will have an impact on how the current workflow works..
It is not optimal but if it is really needed I will implement it in the next days.
Thank you Barth!

Barthelemy:
There is maybe a better option ?  I don't know

njacazio:
Dear Barth, I managed to simplify a bit and fix the problem. Now it should be working. Let's wait for the checks.

Modules/TOF/src/TOFCheckCompressedCounter.cxx
njacazio: I will probably do this asap the PR is merged and I will recompile everything from scratch!
Marking this as resolved so far!

PR #351, opened 2020-03-19
RepoCleaner: Get config file from git (QC-286)

Barthelemy:


Framework/script/RepoCleaner/repoCleaner.py
```diff
@@ -97,6 +101,22 @@ def parseConfig(config_file_path):
 
     return {'rules': rules, 'ccdb_url': ccdb_url}
 
+def downloadConfigFromGit():
+    """
+    Download a config file from git.
+    :param config_git: True if the file must be downloaded from git.
```
knopers8: There is actually no argument for this function, probably a leftover.

Barthelemy:
thanks fixed

PR #353, opened 2020-03-20
Align forward declaration to what is used in O2

ktf:


ktf:
One more reason for NOT disabling the tests of the dependent packages.

ktf:
@knopers8 @Barthelemy I am merging this as it's trivial enough and should fix the issue with compilation of the alidist PRs. Can you please do a new release as soon as you find suitable?

PR #356, opened 2020-03-23
Add the json file for the basic trending of the TPC

cimordas:
Basic trending of one histogram per class (PID and tracks) using the latest version with the arrays of DataSources, and default checkers


knopers8:
Hello, please mark it as "Ready for review" if we should have a look at it.

cimordas:
Hello,
It is now ready for review. I wanted to add the trending of the quality of the checkers inside.

knopers8:
The configuration file seems correct. Merging without waiting for CI checks, as it is only configuration.

PR #357, opened 2020-03-27
Removed runDataProcessing.h header

alice-mch:
The inclusion of the unnecessary runDataProcessing.h header in the MCH pedestal processing task is preventing proper linking under macOS. Surprisingly the linking works fine under Linux.


Barthelemy:
it needs to be rebased. I propose to close it in favour of https://github.com/AliceO2Group/QualityControl/pull/358

PR #360, opened 2020-03-31
Store a test object in the CCDB 

Barthelemy:
with its version number for future backward compatibility testing

### DIFF
--- a/Framework/test/testCcdbDatabase.cxx
+++ b/Framework/test/testCcdbDatabase.cxx
@@ -18,6 +18,7 @@
 #include <unordered_map>
 #include "QualityControl/CcdbDatabase.h"
 #include "QualityControl/QcInfoLogger.h"
+#include "QualityControl/Version.h"
 
 #define BOOST_TEST_MODULE CcdbDatabase test
 #define BOOST_TEST_MAIN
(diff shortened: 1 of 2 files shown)


Framework/test/testCcdbDatabase.cxx
```diff
@@ -99,6 +100,24 @@ BOOST_AUTO_TEST_CASE(ccdb_store)
...
+BOOST_AUTO_TEST_CASE(ccdb_store_for_future_tests)
+{
+  // this test is storing a version of the objects in a different directory.
+  // The goal is to keep old versions of the objects, in old formats, for future backward compatibility testing.
+  test_fixture f;
+
+  TH1F* h1 = new TH1F("to_be_kept", "asdf", 100, 0, 99);
+  h1->FillRandom("gaus", 12345);
+  shared_ptr<MonitorObject> mo1 = make_shared<MonitorObject>(h1, "task", "TST_KEEP");
+  mo1->addMetadata("Run", o2::quality_control::core::Version::GetQcVersion().getString());
```
knopers8, line 112: "Run" instead of "Version" is on purpose?

knopers8: Ah, with the other PR it makes sense. You want to keep one object per each version and you are going to use the "one per run" policy?

Barthelemy: yes, it is exactly that.

PR #361, opened 2020-03-31
Temporarily move monitoring to use send instead of sendGrouped

awegrzyn:


Barthelemy:
out of curiosity : why ?

awegrzyn:
This will be brought back with slightly improved interface from Monitoring v3.0.0

PR #362, opened 2020-03-31
(QC-239)  Storage and retrieval of the encapsulated object, not the MO

Barthelemy:
All tests are passing.


CMakeLists.txt
```diff
@@ -12,7 +12,7 @@ include(GNUInstallDirs)
 # ---- Project ----
 
 project(QualityControl
-        VERSION 0.24.0 # TODO update this automatically when there are new releases
+        VERSION 0.25.0 # TODO update this automatically when there are new releases
```
Barthelemy, line 15: Important to do it already as we have a new data format.

Framework/basic.json
```diff
@@ -16,7 +16,7 @@
         "url": "infologger:///debug?qc"
       },
       "consul": {
-        "url": "http://consul-test.cern.ch:8500"
+        "url": "consul-test.cern.ch:8500"
```
knopers8, line 19: This is just a cosmetic change or it doesn't work without it? Maybe by consequence the urls in other files should be changed then?

Framework/include/QualityControl/MonitorObject.h
```diff
@@ -82,12 +82,19 @@ class MonitorObject : public TObject
...
   void addMetadata(std::string key, std::string value);
-
-  std::map<std::string, std::string> getMetadataMap() const;
+  /// \brief Add key value pairs that will end up in the database as metadata of the object
+  /// Add all the key-value pairs in the map to the MonitorObject. It will be stored in the database as metadata.
+  /// If a key already exists the value will NOT be updated.
+  void addMetadata(std::map<std::string, std::string> pairs);
+  /// \brief Update the value of metadata.
+  /// If the key does not exist it will ignore it.
+  void updateMetadata(std::string key, std::string value);
```
knopers8, line 95: Why have this distinction between `add` and `update`?

Framework/src/CcdbDatabase.cxx
```diff
@@ -126,46 +126,87 @@ void CcdbDatabase::storeMO(std::shared_ptr<o2::quality_control::core::MonitorObj
...
+  ccdbApi.storeAsTFileAny<TObject>(obj, path, metadata, from, to);
+}
+
+void CcdbDatabase::storeQO(std::shared_ptr<QualityObject> qo)
+{
+  // metadata
+  map<string, string> metadata;
+  // QC metadata (prefix qc_)
+  metadata["qc_version"] = Version::GetQcVersion().getString();
+  metadata["qc_quality"] = std::to_string(qo->getQuality().getLevel());
```
knopers8, line 144: Maybe also "qc_detector_name" and "qc_check_name" similarly to MOs? For example, it might be convenient for quickly filtering only the content related to one detector.

Framework/test/testCcdbDatabase.cxx
```diff
@@ -120,41 +128,71 @@ BOOST_AUTO_TEST_CASE(ccdb_store_for_future_tests)
...
+  BOOST_CHECK(!jsonMO.empty());
+
+  auto jsonQO = f.backend->retrieveJson("qc/checks/TST_KEEP/check", 1585647427642);
+  BOOST_CHECK(!jsonQO.empty());
+
+  jsonQO = f.backend->retrieveQOJson("qc/checks/TST_KEEP/check", 1585647427642);
+  BOOST_CHECK(!jsonQO.empty());
 }
 
+//BOOST_AUTO_TEST_CASE(ccdb_retrieve_data_025)
```
knopers8, line 181: is this still needed?

Modules/CMakeLists.txt
```diff
@@ -4,9 +4,9 @@ add_subdirectory(Common)
...
-add_subdirectory(MUON)
-add_subdirectory(TPC)
-add_subdirectory(ITS)
-add_subdirectory(MFT)
+#add_subdirectory(TOF)
+#add_subdirectory(EMCAL)
+#add_subdirectory(MUON)
+#add_subdirectory(TPC)
+#add_subdirectory(ITS)
+#add_subdirectory(MFT)
```
knopers8, line 12: Leftover

Framework/basic.json
Barthelemy: no reason for the change, I just changed it back and forth and forgot what was the original format. I will roll back.

Framework/include/QualityControl/MonitorObject.h
Barthelemy: `addMetadata` used to replace existing values if the key already existed. With the introduction of `addMetadata(map)` I realized that it is not in line with STL which does not update the values in such case but ignore it. So now, both `addMetadata` only add new pairs and ignore the ones whose key is identical. 

It makes it impossible to actually update a value and forced me to introduce `update`.

Framework/src/CcdbDatabase.cxx
Barthelemy: Good idea. I did not do it because we continue storing QualityObjects but you are right that for filtering it is useful. Adding it.

Framework/include/QualityControl/MonitorObject.h
knopers8: Indeed, that seems legit.

Framework/test/testCcdbDatabase.cxx
Barthelemy: yes. We will have to add it when updating the format in the future.

Barthelemy:
no idea why travis is failing
@MichaelLettrich any idea ?

MichaelLettrich:
Unfortunately I don't have any access to travis (we are working on that still). @ktf  will have to look into it :/

Barthelemy:
I will blame a network problem between travis and a repo

Barthelemy:
@knopers8 I have fixed the CCDB test, but had to add a missing parameter to retrieveTObject. You might want to take a look at my last commit.

PR #364, opened 2020-04-02
Fix compilation issues on MacOs 

njacazio:
- Fix also the type of the Size() method
- Add check on data type


njacazio:
@Barthelemy , I fixed it.
It seems that it was related to how the logger works.
Ready to merge the PR

Barthelemy:
Thank you very much, it works.

PR #365, opened 2020-04-06
QC-267 Fix failing listing test

Barthelemy:
Wait for #362

### DIFF
--- a/Framework/test/testCcdbDatabase.cxx
+++ b/Framework/test/testCcdbDatabase.cxx
@@ -72,12 +72,10 @@ BOOST_AUTO_TEST_CASE(ccdb_getobjects_name)
   CcdbDatabase* ccdb = static_cast<CcdbDatabase*>(f.backend.get());
   ILOG(Info) << "getListing()" << ENDM;
   auto tasks = ccdb->getListing("/qc");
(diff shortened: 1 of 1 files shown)


Framework/test/testCcdbDatabase.cxx
```diff
@@ -66,24 +65,20 @@ BOOST_AUTO_TEST_CASE(ccdb_create)
...
-      Objects.insert({ task, object });
-    }
-  }
+  auto tasks = ccdb->getListing("/qc");
+  BOOST_CHECK_GT(tasks.size(), 5); // we know that there are a few
+  BOOST_CHECK_LT(tasks.size(), 100); // we know that there are not too many
+  // print but only for TST
+  auto objects = f.backend->getPublishedObjectNames("/qc/TST");
+  BOOST_CHECK_GT(objects.size(), 0);
+  BOOST_CHECK_LT(objects.size(), 200);
```
knopers8, line 80: I am afraid that someone eventually will break those limits be accident. Maybe it would be enough to check if size != 0?

Barthelemy: I agree

PR #367, opened 2020-04-09
Fix warnings modules

Barthelemy:


Barthelemy:
Following the errors in https://github.com/alisw/alidist/pull/2180 I have fixed the warnings in the MCH module. @aferrero2707 

https://alice.its.cern.ch/jira/browse/QC-315 (why is -WError not applied in QC??)

Barthelemy:
Funny, it does not get the same errors when building QC or when building alidist. Maybe a different machine architecture. I will do something else then.

Barthelemy:
Hopefully this one is ok. I tested on Mac and CC7. This is really annoying how non-portable these formats are. Maybe iostream would help.

aferrero2707:
@Barthelemy thanks for fixing all those warnings! At some point I will have to clean up those verbose messages and probably integrate them with the info logger.

Barthelemy:
The difficulty has been to make it work on both Mac and CC7. I could easily get one of the two but not both. This also explains why it went through : the Mac build is not tested in QC PRs (this is being fixed).

PR #376, opened 2020-04-27
Set c++ standard to 17

Barthelemy:


Barthelemy:
This is just to follow what O2 does.

knopers8:
Then maybe the statements in the lines 73-76 are no longer needed?

Barthelemy:
hmmm I had missed that. I would then do it slightly differntly.

PR #377, opened 2020-04-28
Adapt get ccdb object names

Barthelemy:
The JSON returned by the CCDB has changed format.


Barthelemy:
@knopers8 I am merging without review as it blocks other PRs

PR #386, opened 2020-05-04
QC-327 fix the path to the config file

Barthelemy:


Framework/src/runBasic.cxx
```diff
@@ -133,6 +133,6 @@ std::string getConfigPath(const ConfigContext& config)
   // The the optional one by the user
   auto userConfigPath = config.options().get<std::string>("config-path");
   // Finally build the config path based on the default or the user-base one
-  std::string path = std::string("json:/") + (userConfigPath.empty() ? defaultConfigPath : userConfigPath);
```
knopers8, line 136: That didn't work?

Barthelemy: apparently it lacks a `/`. for sure on Mac it failed. I'll give it a try on CC7.

knopers8: ok, for CC7 both worked as far as I remember

Barthelemy: I cannot test at the moment due to the alicache problem.

Barthelemy: ok it works both on Mac and CC7 now

knopers8: ok, i don't mind either of these alternatives, just wanted to know if both work.

PR #390, opened 2020-05-07
[QC-292] Make multi-node setups work nicely with ServiceDiscovery

knopers8:
Needs to be tested.
@awegrzyn If I have multiple processes running with `ServiceDiscovery` which use a common Consul URL, have different `id`s, but `_register` exactly the same object paths, will I destroy something? Especially the online mode in QCG?


Barthelemy:
Do we merge it or do you want to test it further ?

knopers8:
We merge it.

PR #391, opened 2020-05-08
Docu: mergeability of MonitorObjects in multi node setups

knopers8:


knopers8:
documentation only, merging

PR #392, opened 2020-05-11
Removing additional exception information for better clarity

matthiasrichter:
I have been trying to help @tklemenz with debugging a qc workflow. There is an unhandled exception
```
[8009:QC-TASK-RUNNER-Clusters]: [19:10:11][ERROR] Unhandled exception reached the top of main: Object not found error, device shutting down.
```
which I wanted to understand. Since there is no further name string in the exception message my first assumption was that the methods are called with an empty string. But then I found that the implementation of of `AliceO2::Common::ObjectNotFoundError` drops additional information and for clarity we better skip writing more details into the exception until this is fixed.

https://github.com/AliceO2Group/Common/blob/3f92059e08f191ce6c65a5a9a119e5aba02b89ec/include/Common/Exceptions.h#L28

We can also take this suggestion as a trigger to fix the implementation in AliceO2/Common and maybe even add a sanity check for the arguments in `stopPublishing`/`getMonitoObject` to make the exception message more usable.

Last but not least, I suggest to handle the exception on the level of the QC framework, and only rethrow with additional information if the condition is really fatal.

### DIFF
--- a/Framework/src/ObjectsManager.cxx
+++ b/Framework/src/ObjectsManager.cxx
@@ -90,7 +90,12 @@ void ObjectsManager::stopPublishing(const string& name)
 {
   auto* mo = dynamic_cast<MonitorObject*>(mMonitorObjects->FindObject(name.data()));
   if (mo == nullptr) {
-    BOOST_THROW_EXCEPTION(ObjectNotFoundError() << errinfo_object_name(name));
+    // TODO: ideally we want to pass the object name to the exception but the implementation
+    // of AliceO2::Common::ObjectNotFoundError drops additional information. For clarity we
+    // better skip
+    //    << errinfo_object_name(name));
+    // until this is fixed
+    BOOST_THROW_EXCEPTION(ObjectNotFoundError());
   }
   mMonitorObjects->Remove(mo);
 }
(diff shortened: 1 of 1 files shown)


Barthelemy:
Hi Matthias, 

I agree with most of it. I will try to translate it in corresponding issues in JIRA and will point to them from here. 

1. Fix AliceO2::Common::ObjectNotFoundError
2. add a sanity check for the arguments in stopPublishing/getMonitoObject
3. Reconsider the rethrowing of exceptions in QC (for non fatal exceptions)

It seems that there is a problem with this PR. 

Cheers,

matthiasrichter:
Thanks, I think if especially https://alice.its.cern.ch/jira/browse/O2-1416 is expected to be fixed very soon we can also drop this PR.

PS: there was a typo, now the compilation works.

Barthelemy:
Having another look at it, it seems that the implementation does not drop additional information. If I `cout` an `ObjectNotFound` exception I get:
```
/Users/barth/dev/alice/sw/SOURCES/Common-O2/master/0/test/testExceptions.cxx(17): Throw in function void foo()
Dynamic exception type: boost::wrapexcept<AliceO2::Common::ObjectNotFoundError>
std::exception::what: AliceO2::Common::Exception
[AliceO2::Common::errinfo_object_name_*] = objectName
```
This being said, I understand that FairMQ / DPL (?) prints the what() and thus we miss the object name. I have prepared this PR to fix it : https://github.com/AliceO2Group/Common/pull/36
If you agree with it, we can drop this PR as you suggest.

matthiasrichter:
Indeed, the problem is the implementation of `ObjectNotFoundError::what`, sorry for not being clear enough. I think in the end we should close this PR if the cause of the issue is fixed. I just want to make sure that all what is intended to be the output of the exception ends up in the message.

One comment on AliceO2Group/Common#36, as it is right now one would need to wrap the actual object name into `errinfo_object_name`, right? So I can adjust this PR accordingly, if you want to keep it like that, see my comments there.

Barthelemy:
The latest versions of O2 and QC should provide the full details of the exceptions including the stack trace. This is also the case if the exception escapes the QC framework (ie. reaches the DPL driver).

Therefore I am going to close this issue. Feel free to reopen if you think that something has actually not been addressed.

PR #393, opened 2020-05-11
[QC-331] Fix standalone QC infrastructures - add Data Sampling

knopers8:
@Barthelemy Can you try with this? I think I just forgot to generate Data Sampling in standalone infrastrastructures, that should fix it.


knopers8:
Ok, it finally built on my laptop. It seems to fix the problem. I will add a unit test for that tomorrow.

Barthelemy:
it works ! thanks

PR #394, opened 2020-05-11
Add first working prototype of QC on TOF digits

njacazio:
Checking:
- hit multiplicity
- hit time
- hit ToT

- Improve checker with useful messages
- Made limits (hard) in the header file -> to be switched to configurable

### DIFF
--- a/Modules/TOF/src/TOFCheckRawsMulti.cxx
+++ b/Modules/TOF/src/TOFCheckRawsMulti.cxx
@@ -142,7 +142,7 @@ Quality TOFCheckRawsMulti::check(std::map<std::string, std::shared_ptr<MonitorOb
             else { // High multiplicity running e.g. Pb-Pb
               if (isLowMultContentHigh) {
                 result = Quality::Medium;
-                shifter_msg = Form("High low-multiplicity counts (%.2f higher than total)!", fracAtLowMult);
+                shifter_msg = Form("Low-multiplicity counts are high\n(%.2f higher than total)!", fracAtLowMult);
               } else if (multiMean > maxTOFrawhitsPbPb) {
                 //assume that good range of multi in PbPb goes from 20 to 500 tracks
                 result = Quality::Medium;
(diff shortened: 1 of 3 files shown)


njacazio:
Ciao @ktf, I think my PR build is stuck, could you please have a look?

ktf:
@MichaelLettrich can you take a look? Thanks.

MichaelLettrich:
the macos builder died. restarted and should be up and running soon.

ktf:
can you restart the linux one as well?

Modules/TOF/src/TOFCheckRawsMulti.cxx
```diff
@@ -80,73 +81,79 @@ Quality TOFCheckRawsMulti::check(std::map<std::string, std::shared_ptr<MonitorOb
...
-              isINT7AverageHigh = kTRUE;
+          } else { // Running with collisions
+            const Bool_t isZeroBinContentHigh = (zeroBinIntegral > fracAtZeroMult * totIntegral);
+            const Bool_t isLowMultContentHigh = (lowMIntegral > fracAtLowMult * totIntegral);
+            const Bool_t isINT7AverageLow = (multiMean < minTOFrawhits);
+            const Bool_t isINT7AverageHigh = (multiMean > maxTOFrawhits);
 
             // if (AliRecoParam::ConvertIndex(specie) == AliRecoParam::kLowMult) {
-            if (0) {
+            if (0) { // TODO: Low multiplicity running, how to check if it is pp? Probably this can be simplified in the json
```
knopers8, line 119: Good question. One way would be to add a parameter to a Check json configuration:
```
"MyCheck": {
  ...
  "checkParameters" : {
    "runType" : "cosmics"
  }
 ...
}
```
But in a longer run, I think we should consider adding this kind of metadata to each MonitorObject in framework, as soon as we are able to retrieve it. @Barthelemy what do you think?

njacazio: I was thinking of removing all the ifs from the task and simply feeding it different json files depending on the running conditions so as to entirely delegate this to the initial configuration but  suggestions are most welcome!

PR #395, opened 2020-05-12
Add isBeingPublished

Barthelemy:


Barthelemy:
@ktf @MichaelLettrich I think that o2-dataflow is stuck again.

PR #397, opened 2020-05-14
QC-334 Rework the catching and rethrowing of exceptions

Barthelemy:


Barthelemy:
I tried to limit the number of try.. catch blocks by adding them at the highest possible level in the call stack. At the moment we systematically rethrow because if it reaches this point it is too late. Users should catch them in their code ideally. Initialization and configuration issues are mostly non-recoverable.

Framework/src/TaskRunner.cxx
```diff
@@ -142,16 +166,16 @@ CompletionPolicy::CompletionOp TaskRunner::completionPolicyCallback(o2::framewor
     }
   }
 
-  LOG(DEBUG) << "Completion policy callback. "
-             << "Total inputs possible: " << inputs.size()
-             << ", data inputs: " << dataInputsPresent
-             << ", timer inputs: " << (action == CompletionPolicy::CompletionOp::Consume);
+  ILOG(Debug) << "Completion policy callback. "
```
knopers8, line 169: I would prefer if we kept this debug log as it was. We don't have a way to suppress infologger debug logs in DPL and this specific one will occur a lot. As soon as this is fixed, we can use infologger interface here.

Barthelemy:
if https://github.com/AliceO2Group/AliceO2/pull/3597 is accepted (and it is), I am not sure we need this PR. Your thoughts ?

knopers8:
Perhaps getting rid of exception handling in `initDatabase` would be still beneficial, as DPL should catch and print it as well (shouldn't it?). Also the example of non-fatal exception in `SkeletonTask` could still stay.

PR #400, opened 2020-05-15
Add and remove monitoring metrics

Barthelemy:
Keep only 4 metrics : 
- qc_objects_received
- qc_checks_executed 
- qc_mo_stored
- qc_qo_stored

They all have their rate published.

### DIFF
--- a/Framework/include/QualityControl/CheckRunner.h
+++ b/Framework/include/QualityControl/CheckRunner.h
@@ -194,9 +194,7 @@ class CheckRunner : public framework::Task
 
   // monitoring
   std::shared_ptr<o2::monitoring::Monitoring> mCollector;
-  //  std::chrono::system_clock::time_point startFirstObject;
-  //  std::chrono::system_clock::time_point endLastObject;
-  int mTotalNumberHistosReceived;
+  int mTotalNumberObjectsReceived;
   int mTotalNumberCheckExecuted;
   int mTotalNumberQOStored;
   int mTotalNumberMOStored;
--- a/Framework/include/QualityControl/CheckRunner.h
+++ b/Framework/include/QualityControl/CheckRunner.h
@@ -29,6 +29,7 @@
 #include <Framework/Task.h>
 #include <Headers/DataHeader.h>
 #include <Monitoring/MonitoringFactory.h>
+#include <Configuration/ConfigurationInterface.h>
 #include <Framework/DataProcessorSpec.h>
 // QC
 #include "QualityControl/CheckInterface.h"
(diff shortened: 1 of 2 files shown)


Framework/include/QualityControl/CheckRunner.h
```diff
@@ -193,9 +194,12 @@ class CheckRunner : public framework::Task
 
   // monitoring
   std::shared_ptr<o2::monitoring::Monitoring> mCollector;
-  std::chrono::system_clock::time_point startFirstObject;
-  std::chrono::system_clock::time_point endLastObject;
+  //  std::chrono::system_clock::time_point startFirstObject;
+  //  std::chrono::system_clock::time_point endLastObject;
   int mTotalNumberHistosReceived;
```
knopers8, line 199: Could you take an opportunity to change it to `mTotalNumberMOReceived` for example? Histos are a bit misleading.

Framework/src/CheckRunner.cxx
```diff
@@ -254,10 +250,12 @@ void CheckRunner::run(framework::ProcessingContext& ctx)
...
 
   // monitoring
-  endLastObject = system_clock::now();
   if (timer.isTimeout()) {
     timer.reset(1000000); // 10 s.
-    mCollector->send({ mTotalNumberHistosReceived, "objects" }, o2::monitoring::DerivedMetricMode::RATE);
+    mCollector->send({ mTotalNumberHistosReceived, "qc_objects_received" }, DerivedMetricMode::RATE);
+    mCollector->send({ mTotalNumberCheckExecuted, "qc_checks_executed" }, DerivedMetricMode::RATE);
+    mCollector->send({ mTotalNumberQOStored, "qc_qo_stored" }, DerivedMetricMode::RATE);
+    mCollector->send({ mTotalNumberMOStored, "qc_mo_stored" }, DerivedMetricMode::RATE);
```
knopers8, line 258: Perhaps we should stick either to `objects` or `mo`, since we mean the same. In TaskRunner we use `objects`.

Barthelemy: actually I had a commit that I forgot to push that did exactly that.

Barthelemy: wait, I meant another one, here I wanted to differentiate MOs from QO stored. I could merge them and then I would call them objects.

Framework/include/QualityControl/CheckRunner.h
Barthelemy: totally agree

Framework/src/CheckRunner.cxx
knopers8: Differentiating MOs and QOs makes sense completely. I was more reluctant to use `objects` and `mo` interchangeably. However, as I think of it now, we will allow not only MOs here, but also other kinds of objects might be injected here, which won't be anyway stored. So the distinction is actually justified here.

Barthelemy:
@MichaelLettrich I think that there is still a problem with o2-dataflow

Barthelemy:
there is a stuck travis but the other one is ok.

PR #401, opened 2020-05-20
Adapt to new ConfigParamRegistry API

ktf:


ktf:
The API is going to change to allow additional feature. This should allow new and old API to work.

knopers8:
And there is a legit compilation problem.

ktf:
Ok, updated.

ktf:
Merging this as everything is green and only a test was changed.

ktf:
Can we have a new tag with this in, when you have a moment?

Barthelemy:
ok, I will do it today.

PR #406, opened 2020-05-26
[QC-337] Fix testCheckWorkflow

knopers8:
The main problem was caused by erasing the elements of a set while iterating on it. Also, I took liberty of fixing typos and updating a few parts of code to the recent changes in DPL.


Barthelemy:
Nice, the good old "erase while iterate"  problem. Thanks for fixing it.

knopers8:
It is interesting that it worked fine on CC7/GCC anyway.

PR #408, opened 2020-05-28
MCH: removed obsolete o2-qc-run-file-reader executable

aferrero2707:
The QC file reader is now replaced by o2-raw-file-reader-workflow


Barthelemy:
thanks

PR #410, opened 2020-05-29
Functional test (QC-336)

Barthelemy:


knopers8:
```
21/35 Test #26: functional_test ..................***Failed    0.02 sec
/usr/bin/sh: o2-qc-functional-test.sh: command not found
```
You might need to set executable rights in git so they are propagated elsewhere, I remember having similar situation once.

Barthelemy:
@knopers8 right, I have prepared another fix, but actually I should also what you say. I thought that chmoding it before adding in git was enough.

Barthelemy:
This will fix the problem : https://github.com/alisw/alidist/pull/2283

PR #414, opened 2020-06-05
Add metadata to Quality (QC-308)

Barthelemy:


Framework/test/testCcdbDatabase.cxx
```diff
@@ -107,22 +107,6 @@ BOOST_AUTO_TEST_CASE(ccdb_store_for_future_tests)
   f.backend->storeQO(qo1);
 }
 
-BOOST_AUTO_TEST_CASE(ccdb_retrieve, *utf::depends_on("ccdb_store"))
```
knopers8, line 110: I don't understand why we don't need this anymore.

Barthelemy: I am not sure how we ended up with this. The code is identical to `ccdb_store_for_future_tests` but slightly outdated. It does not test anything more than `ccdb_store_for_future_tests`.

knopers8: Ah, ok. I didn't see the other one in the diff.

PR #415, opened 2020-06-08
Increase the timeout of the tests for the CI

Barthelemy:


Barthelemy:
So it seems that the timeout duration is not the problem. The problem is tracked and followed up by Giulio here : https://alice.its.cern.ch/jira/browse/QC-345

Barthelemy:
Apparently rebooting the Mac worker has fixed the problem for QC.

Barthelemy:
Seems that there are still cases, on CC7, when the tests time out. To be safe, we extend them.

PR #417, opened 2020-06-09
MCH: global plots

seperrin:
Added GlobalHistogram to plot quantities in detector coordinates.
Regrouped histograms in electronics mapping in a single plot (instead of one per solar)

### DIFF
--- a/Modules/MUON/MCH/CMakeLists.txt
+++ b/Modules/MUON/MCH/CMakeLists.txt
@@ -5,6 +5,7 @@ set(MODULE_NAME "QcMuonChambers")
 set(SRCS
   src/Mapping.cxx
   src/Decoding.cxx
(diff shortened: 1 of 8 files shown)


aferrero2707:
@Barthelemy I guess that the timeout failures in the macOS checks are addressed by [O2 PR 415](https://github.com/AliceO2Group/QualityControl/pull/415), which actually fails due to the same timeouts...
Do you have an idea when this will be fixed?

Barthelemy:
@aferrero2707 Indeed, the duration of the timeout is not actually helping. The issue to follow is here : https://alice.its.cern.ch/jira/browse/QC-345
Something has happened last weekend (?) that broke these tests. QC itself has not been updated in a while. Thus it seems to be caused by a dependency. I am not able to reproduce the problem on my Mac.

PR #419, opened 2020-06-11
DO NOTE MERGE Try to rollback for tests

Barthelemy:


Barthelemy:
@MichaelLettrich Any idea why the macos tests are orange ?

MichaelLettrich:
machine got restarted apparently. I will restart the builders.

Barthelemy:
it seems that reverting 5 commits fixes the problem ?

Barthelemy:
still it fails in AliceO2 from time to time and in alidist, both use v0.26 so anyways I don't believe it is one specific commit.

Barthelemy:
apparently rebooting the Mac machine solved our tests problems.

PR #420, opened 2020-06-12
Cleanup checks code (QC-170)

Barthelemy:


Barthelemy:
@knopers8 What do you think ? do we include this in v1.0 ? do I unWIP ?

Framework/src/InfrastructureGenerator.cxx
```diff
@@ -368,47 +370,34 @@ void InfrastructureGenerator::generateCheckRunners(framework::WorkflowSpec& work
...
-      InputNames singleEntry{ label };
-      // Init empty Check vector to appear in the next step
-      checkRunnerMap[singleEntry];
-      storeVectorMap[singleEntry].push_back(label);
-    }
+    // If there is no Check for a given input, create a candidate for a sink device
+    InputNames singleEntry{ label };
+    // Init empty Check vector to appear in the next step
+    checksMap[singleEntry];
+    storeVectorMap[singleEntry].push_back(label);
```
knopers8, line 395: I don't understand why this can be put outside of the if statement.

Barthelemy: Now that I look at it with fresh eyes, I wonder the same. I will restore it.

Barthelemy:
The tests were testing the deleted methods which were used nowhere else. I can probably remove them altogether.

Framework/src/InfrastructureGenerator.cxx
knopers8: OK. Also that first loop was doubled before, but it seemed actually redundant. Still, I would give it a go with a few configurations just in case.

Barthelemy: I tried with a few different ones, with and without checks, and it seems to work. I am now considering adding tests to replace the old ones that were not testing anything anymore.

knopers8: Thanks. Actually, could you please make the refactoring of `InfrastructureGenerator::generateCheckRunners` a separate commit, in case we have to revert it?

Barthelemy:
I had a look at the coverage of CheckRunner & Co. It is actually high and they are covered, if not in dedicated tests, in other tests. So I am not going to spend time right away to rewrite the commented out tests but will rather remove them altogehter if you agree.

Framework/src/InfrastructureGenerator.cxx
Barthelemy: agreed
on it

knopers8:
Let's see what CI says.

knopers8:
> So I am not going to spend time right away to rewrite the commented out tests but will rather remove them altogehter if you agree.

Sure

PR #421, opened 2020-06-12
Add a histogram producer

Barthelemy:
- Add the producer for histograms
- Send a TObjArray from the histograms producer.

### DIFF
--- a/Framework/include/QualityControl/HistoProducer.h
+++ b/Framework/include/QualityControl/HistoProducer.h
@@ -1,0 +1,39 @@
+// Copyright CERN and copyright holders of ALICE O2. This software is
+// distributed under the terms of the GNU General Public License v3 (GPL
+// Version 3), copied verbatim in the file "COPYING".
+//
+// See http://alice-o2.web.cern.ch/license for full licensing information.
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+///
+/// \file   HistoProducer.h
+/// \author Barthelemy von Haller
+///
+
+#ifndef QUALITYCONTROL_HISTOPRODUCER_H
+#define QUALITYCONTROL_HISTOPRODUCER_H
+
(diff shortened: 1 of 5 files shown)


Barthelemy:
Let me know if you think it is better to have a separate runHistoProducer.

Framework/include/QualityControl/HistoProducer.h
```diff
@@ -0,0 +1,39 @@
...
+/// \brief Returns a random data producer specification which publishes on {"TST", "RAWDATA", <index>}
+///
+/// \param myParam  The value the producer should produce.
+/// \return         A fixed number producer specification
+framework::DataProcessorSpec getHistoProducerSpec(size_t myParam);
+
+/// \brief Returns an algorithm generating random messages
+///
+/// \param output   Origin, Description and SubSpecification of data to be produced
+/// \param  myParam  The value the producer should produce.
```
knopers8, line 33: The description of `myParam` is unclear in this context (same in the function above)

Framework/src/HistoProducer.cxx
```diff
@@ -0,0 +1,79 @@
...
+///
+/// \file   HistoProducer.cxx
+/// \author Barthelemy von Haller
+///
+#include "QualityControl/HistoProducer.h"
+
+#include <random>
+#include <Common/Timer.h>
+#include <TH1F.h>
+#include <Common/Timer.h>
```
knopers8, line 20: Timer included twice

Barthelemy:
So I have split it. 

To be noted that the try...catch in the Printer, to test if it is a TObjArray, is not very nice. The solution Giulio advised is to use `TypeIdHelpers::uniqueId<T>()` from `Framework/TypeIdHelpers.h`, put it in an extra header and use the same in a switch on the receiving side. I will give it a try later on, but for this simple utility it is good enough I believe.

knopers8:
Thanks. I am not merging it in case that you want to improve that try-catch thing before.

Barthelemy:
I think that it is ok for now like that.

PR #427, opened 2020-06-17
QC-350 automatic disconnection from consul

Barthelemy:


awegrzyn:
Just to remind, this will deregister service with its tags (QC task with its published objects) from Consul when health check fails for `1m` (`1m` is minimum time).

PR #428, opened 2020-06-18
[QC-351] Use ptree to configure PP and get config pushed by Control

knopers8:
This is in principle done, but I have to confirm it works within an AliECS workflow.


knopers8:
It works when tested with `peanut`

PR #429, opened 2020-06-18
Refactoring generateCheckRunners

Barthelemy:
after https://github.com/AliceO2Group/QualityControl/pull/420


Barthelemy:
commits to be cleaned up

PR #431, opened 2020-06-18
(QC-352) Functional test clean up

Barthelemy:
Make sure we clean up the leftovers of the functional test if any


Barthelemy:
no idea why travis fails

knopers8:
Maybe yaml files are ignored and travis fails when it is not given any files to check.

knopers8:
There are also check results which should probably be removed as well.

Barthelemy:
thanks, I think that you are right for travis.

PR #432, opened 2020-06-18
[EMCAL-524] Add option for cell publisher in the publisher application

mfasDa:


knopers8:
Hi, there is warning (treated as error):
```
/mnt/mesos/sandbox/sandbox/sw/SOURCES/QualityControl/build_QualityControl_o2-dataflow/0/Modules/EMCAL/src/runQCEMCALDigits.cxx: In function 'o2::framework::WorkflowSpec defineDataProcessing(const o2::framework::ConfigContext&)':
/mnt/mesos/sandbox/sandbox/sw/SOURCES/QualityControl/build_QualityControl_o2-dataflow/0/Modules/EMCAL/src/runQCEMCALDigits.cxx:78:55: error: typedef 'using digitInputType = class std::vector<o2::emcal::Digit>' locally defined but not used [-Werror=unused-local-typedefs]
   using digitInputType = std::vector<o2::emcal::Digit>;
```

mfasDa:
Oops, was a leftover from the previous version. I removed it now. Please merge when the build tests are fine.

We will decommission the digits part in a second step together with the corresponding change in the task. Up to now cells and digits were storing the same information, just one in a compressed format. However due to a recent change in the EMCAL simulation the digit was converted as a raw digit so we can no longer use it for tower information and have to stick with the cell format here.

PR #434, opened 2020-06-23
Canvas options (QC-253)

Barthelemy:
Let users set drawOptions and display hints on MOs.

### DIFF
--- a/Framework/src/ObjectsManager.cxx
+++ b/Framework/src/ObjectsManager.cxx
@@ -28,6 +28,9 @@ using namespace std;
 namespace o2::quality_control::core
 {
 
+const std::string ObjectsManager::gDrawOptionsKey = "drawOptions";
+const std::string ObjectsManager::gDisplayHints = "displayHints";
+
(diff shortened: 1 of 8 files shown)


Framework/src/ObjectsManager.cxx
```diff
@@ -124,4 +127,28 @@ int ObjectsManager::getNumberPublishedObjects()
...
+
+void ObjectsManager::setDisplayHint(const std::string& objectName, const std::string& hints)
+{
+  MonitorObject* mo = getMonitorObject(objectName);
+  mo->addOrUpdateMetadata(gDisplayHintsKey, hints);
+}
+
+void ObjectsManager::setDisplayHint(TObject* obj, const std::string& hints)
+{
+  MonitorObject* mo = getMonitorObject(obj->GetName());
```
knopers8, line 150: This is unrelated to this PR, but maybe we should consider protecting those MOs from being deleted by returning a const pointer (with non-const object pointed to).

Barthelemy: Agreed, could you create a JIRA ticket ?

knopers8: https://alice.its.cern.ch/jira/browse/QC-361

PR #437, opened 2020-06-25
[EMCAL-524] Add occupancy histograms

cterrevo:
Adapt task also to Cells

### DIFF
--- a/Modules/EMCAL/src/DigitsQcTask.cxx
+++ b/Modules/EMCAL/src/DigitsQcTask.cxx
@@ -5,12 +5,14 @@
 
 #include <TCanvas.h>
 #include <TH2.h>
+#include <TProfile2D.h>
 
 #include <DataFormatsEMCAL/EMCALBlockHeader.h>
 #include <DataFormatsEMCAL/TriggerRecord.h>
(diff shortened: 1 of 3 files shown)


Modules/EMCAL/src/DigitsQcTask.cxx
```diff
@@ -95,21 +107,32 @@ void DigitsQcTask::monitorData(o2::framework::ProcessingContext& ctx)
...
+
       try {
+        // fill here mDigitOccupancy for column and rows
+        //if we fill phy vs eta plots integrated: filled with eta phi GlobalRowColFromIndex  from Geometry
+
+        auto [row, col] = mGeometry->GlobalRowColFromIndex(digit.getTower());
+        if (digit.getEnergy() > 0)
+          mDigitOccupancy->Fill(col, row);
+        mIntegratedOccupancy->Fill(col, row, digit.getEnergy());
+        std::cout << " fill histo integrated occupancy  and occupancy with col" << col << " row " << row << " energy as weight " << digit.getEnergy() << std::endl;
```
knopers8, line 130: Please use some logging facility instead of `cout` to make sure that the messages reach infologger in production.

PR #438, opened 2020-06-25
Register the list of QOs in Consul from CheckRunner (QC-358)

Barthelemy:


Barthelemy:
The CheckRunner is a different process than the TaskRunner which already publishes the list of MonitorObjects published to the ServiceDiscovery. I have added code to the CheckRunner to have him publish the list of QualityObjects its Checks create. 

I have one issue though. The healthcheck fails due to 
> [11902:QC-CHECK-RUNNER-QcCheck]: 2020-06-25 13:13:31.001062 !!! Error - ServiceDiscovery::runHealthServer - bind: Address already in use

I am unsure which address we are talking about here. @awegrzyn do you know ?

awegrzyn:
Health checks are done in a way that your process binds a port (I think range is `7777+`); then Consul tries to connect to this port in order to find out whether process is still alive.
Your error comes from https://github.com/AliceO2Group/QualityControl/blob/fd47daf9aca36e9fc9438a7bd0e1303b531d47ef/Framework/src/ServiceDiscovery.cxx#L109 and it means that either the port you trying to bind is already taken by another process/service or there's already health check running.
I remember that in the final solution it should be AliECS giving you port to bind.

Barthelemy:
I presume that the port is already taken by the TaskRunner. I'll +1 the port for CheckRunner. Thank you for the explanation, very clear.

Framework/include/QualityControl/CheckRunner.h
```diff
@@ -198,6 +206,10 @@ class CheckRunner : public framework::Task
   // Checks cache
   std::map<std::string, std::shared_ptr<MonitorObject>> mMonitorObjects;
 
+  // Service discovery
+  std::shared_ptr<ServiceDiscovery> mServiceDiscovery;
```
Barthelemy, line 210: I would have liked a unique_ptr but a `Task` cannot have a unique_ptr for some reasons.

awegrzyn: Do you know why? I'm just curious.

Barthelemy: no, and I am trying to reproduce the problem in a minimal setup to report it.

knopers8: I think it is because `Task` is transformed into `DataProcessorSpec` and the latter is copied somewhere in DPL.

Framework/include/QualityControl/CheckRunner.h
```diff
@@ -37,6 +37,7 @@
 #include "QualityControl/MonitorObject.h"
 #include "QualityControl/QcInfoLogger.h"
 #include "QualityControl/Check.h"
+#include "QualityControl/ServiceDiscovery.h"
```
knopers8, line 40: Could you try forward-declaring `ServiceDiscovery` here? I see that we need only the pointer in this header. 
```suggestion
namespace o2::quality_control::core
{
class ServiceDiscovery;
}
```
Then .cxx file of course needs the include statement.

Framework/include/QualityControl/ServiceDiscovery.h
```diff
@@ -34,6 +34,7 @@ class ServiceDiscovery
  public:
   /// Sets up CURL and health check
   /// \param url 		Consul URL
+  /// \param name
```
knopers8, line 37: indeed, i forgot about it. Could you add the argument description?
```suggestion
  /// \param name		Service name
```

Framework/src/CheckRunner.cxx
```diff
@@ -319,6 +322,32 @@ void CheckRunner::send(QualityObjectsType& qualityObjects, framework::DataAlloca
...
 }
 
+void CheckRunner::updateServiceDiscovery(const QualityObjectsType& qualityObjects)
+{
+  if (mServiceDiscovery == nullptr) {
+    return;
+  }
+
+  // insert into the list of paths the QOs' paths.
+  // TODO it seems unoptimal to do it all the time. After a while, aren't we sure that we have run all the checks ?
```
knopers8, line 332: Don't we actually know all the QO names during initialization? Why doing it once is not enough?

Barthelemy:
yes, but probably we can make it work with the proper set of copy constructors in ServiceDiscovery and CheckRunner. I am giving it a try.

Framework/src/CheckRunner.cxx
```diff
@@ -40,6 +40,7 @@ using namespace o2::configuration;
 using namespace o2::monitoring;
 using namespace o2::quality_control::core;
 using namespace o2::quality_control::repository;
+using namespace std;
```
knopers8, line 43: I don't see where you access some STL class without `std` here. Please remove it if it is not necessary.
```suggestion
```

Framework/src/ServiceDiscovery.cxx
```diff
@@ -91,13 +91,13 @@ void ServiceDiscovery::_register(const std::string& objects)
   boost::property_tree::json_parser::write_json(ss, pt);
 
   send("/v1/agent/service/register", ss.str());
-  QcInfoLogger::GetInstance() << "Registration to ServiceDiscovery" << AliceO2::InfoLogger::InfoLogger::endm;
+  ILOG(Info) << "Registration to ServiceDiscovery : " << objects << ENDM;
```
knopers8, line 94:
```suggestion
  ILOG(Info) << "Registration to ServiceDiscovery: " << objects << ENDM;
```

Framework/include/QualityControl/CheckRunner.h
Barthelemy: It is a rabbit hole which starts with `CheckRunner` not having a default copy constructor due to `mServiceDiscovery` not having one. It cannot have one because of `curlHandle` not having a move constructor somehow. 

It is certainly fixable but I am not sure that it is worth the trouble. Would you be fine with having the shared_ptr instead ?

Framework/include/QualityControl/CheckRunner.h
Barthelemy: sure

Framework/src/CheckRunner.cxx
Barthelemy: Right, I got mistaken with the new policy `OnEachSeparately`. Actually each Check has either 1 single output or as many outputs as inputs in case of `OnEachSeparately`, correct ? 

I'll rework this part.

Framework/include/QualityControl/CheckRunner.h
knopers8: Fine with me. I don't think there is any risk of concurrent access here.

Framework/src/CheckRunner.cxx
knopers8: Hmm, in case of `OnEachSeparately` it is as many as there are MOs. But notice that if someone declares "all" MOs, then you don't know the amount of them. So in that case indeed we don't know the number of QOs during the initializaiton.
So maybe it is actually worth to leave it like that. Another approach for `OnEachSeparately` would be to declare only the Check, not the QOs themselves, then we could do it once, during init.

Barthelemy: Ok, just remembered what was the issue. If we use `OnEachSeparately` and the config file says 
```
{
          "type": "Task",
          "name": "TaskName",
          "MOs": "all"
        }
```
then we don't know what are the MOs and thus cannot build the corresponding paths during initialization. 

Am I missing something ?

knopers8: Yes, exactly.
So you were right to have the update function inside `run()`

Barthelemy: Added some info to remember why it is like that and how to improve it if we want.

PR #439, opened 2020-06-25
Adapt to support new services API

ktf:
A new service declaration API will be introduced shortly. This
makes sure QC is compatible with both old and new API.


ktf:
This is required to have AliceO2Group/AliceO2#3889 merged. Can we have a new tag with it?

Barthelemy:
I will tag v0.26.3 with this commit on top of v0.26.2.

ktf:
Thanks.

Barthelemy:
`build/QualityControl/o2` seems stuck

PR #441, opened 2020-06-25
Adapt to changes in DPL ServiceSpec API

ktf:


ktf:
@Barthelemy The fix on master was missing one... Could you tag again?

Barthelemy:
There is something wrong with the CI

knopers8:
It seems like it reacts only to PRs to master.

Barthelemy:
@ktf So it is ok on master (from what I see) but not in v0.26.3 ?
@knopers8 indeed

--> I am going to make the same change as the one propose in this PR on top of v0.26.3 and tag v0.26.4

Barthelemy:
Ok, it is here : https://github.com/AliceO2Group/QualityControl/tree/v0.26.4

PR #443, opened 2020-07-01
Trending task for ITS threshold runs monitoring. New reductor for THn…

iravasen:
…Sparse also added. New TrendingTaskConfig to allow different json configuration file layout.


JianLIUhep:
Hi @iravasen, thanks a lot for the pull request on the ITS online post-processing task.  It seems it failed the clang-formating check. Could you please run clang-format to this PR? Thank you.
You can follow the instructions here: https://github.com/AliceO2Group/CodingGuidelines#alice-o2-coding-guidelines

Barthelemy:
Thanks for the contribution, I am going to review it today or tomorrow.

iravasen:
Hi @Barthelemy , thanks a lot for the review. I implemented all your comments/suggestions. I removed runPostProcessingITS and RootClassFactory (was copied to Modules/ITS for a test and then I forgot to remove it). I tested the new modifications and everything is working.

Modules/ITS/itsQCTrending.json
```diff
@@ -0,0 +1,719 @@
...
+        "moduleName": "QualityControl",
+        "detectorName": "ITS",
+        "dataSources": [
+          {
+            "types": ["repository","repository"],
+            "paths": ["qc/ITS/ITSTHRTask0/Threshold/Layer0/Stave0/Threshold","qc/ITS/ITSTHRTask0/DeadPixel/Layer0/Stave0/DeadPixelHITMAP"],
+            "names": [ "Threshold_l0_0", "DeadPixels_l0_0" ],
+            "reductorNames": ["o2::quality_control_modules::common::TH1Reductor", "o2::quality_control_modules::common::THnSparseReductor"],
+            "moduleNames": ["QcCommon","QcCommon"]
+          },
```
knopers8, line 38: Hi, I will have a closer look in August, now I only had a glimpse. Weren't you able to specify one reductor and only multiple names? For example TPC did that here https://github.com/AliceO2Group/QualityControl/blob/7ceb400fa2fc2374dfc43d6073838e51416fbf0e/Modules/TPC/run/tpcQCSimpleTrending.json#L35

So in your case it would be
```
          {
            "type": "repository",
            "path": "qc/ITS/ITSTHRTask0/Threshold/Layer0/Stave0/Threshold",
            "names": [ "Layer0/Stave0/Threshold/Threshold_l0_0",
                     "Layer0/Stave0/Threshold/Threshold_l0_1", ... ,
                     "Layer0/Stave1/Threshold/Threshold_l0_0", ...,
                     "Layer1/Stave4/Threshold/Threshold_l0_0" ], (basically all Threshold histograms listed here)
            "reductorName": "o2::quality_control_modules::common::TH1Reductor"
            "moduleName": "QcCommon",
          },
```
and similarly the second one for DeadPixelHITMAPs

Modules/ITS/itsQCTrending.json
```diff
@@ -0,0 +1,719 @@
...
+          {
+            "types": ["repository","repository"],
+            "paths": ["qc/ITS/ITSTHRTask2B/Threshold/Layer2/Stave19/Threshold","qc/ITS/ITSTHRTask2B/DeadPixel/Layer2/Stave19/DeadPixelHITMAP"],
+            "names": [ "Threshold_l2_19", "DeadPixels_l2_19" ],
+            "reductorNames": ["o2::quality_control_modules::common::TH1Reductor", "o2::quality_control_modules::common::THnSparseReductor"],
+            "moduleNames": ["QcCommon","QcCommon"]
+          }
+        ],
+        "plots": [
```
knopers8, line 369: Regardless of the potential improvement for dataSources, I don't see how to improve it similarly for plots with the current version of TrendingTask.
I will have to see in August if there might be some way of creating multiple very similar plots without too much boilerplate

Modules/ITS/itsQCTrending.json
Barthelemy: @iravasen I let you comment but it could go to another PR if we want to merge this one.

Modules/ITS/itsQCTrending.json
Barthelemy: Thanks

Barthelemy:
Approved -> automatic CI launched

Modules/ITS/itsQCTrending.json
iravasen: Hi @knopers8, thanks for your comments. One way to compact the json would be something like:

"path": ["qc/ITS/ITSTHRTask0/Threshold/Layer0/Stave0/Threshold", "qc/ITS/ITSTHRTask0/Threshold/Layer0/Stave1/Threshold",....],
"names": ["Threshold_l0_0", "Threshold_l0_1", ....] 

And the same for the dead pixels. This for me would be (probably) the easiest since our real paths on CCDB are the one indicated into "paths" while the "names" are just names to distinguish the different branches of the tree and are not part of the path. In fact I do retrieveMO(dataSource.path, "") to retrieve the object and not retrieveMO(dataSource.path, dataSource.name). If you agree, I propose to go ahead with the PR and then I will do this modification asap (it will take some time because I will have also to modify the TrendingTaskConfig). Please let me know your opinion. Thanks a lot.

iravasen: or as you said but:
"path": "qc/ITS",
"names": ["ITSTHRTask0/Threshold/Layer0/Stave0/Threshold", ....]

but in this case I would have very long names of the branches (provided that one doesn't not any string editing).

Modules/ITS/itsQCTrending.json
iravasen: Yes, I will have also to think about in the next days.

Modules/ITS/itsQCTrending.json
knopers8: As you prefer. though I would prefer if the original TrendingTaskConfig was kept as it is. Longer branch names are not a big deal IMO, but probably I am looking at it from a performance point of view, not user convenience :)
Anyway, please go forward as you and Barth like, I will have a closer look after I am back in August.

iravasen: Thanks a lot @knopers8, I will work on this soon.

Modules/Common/include/Common/THnSparseReductor.h
```diff
@@ -0,0 +1,48 @@
...
+#include "QualityControl/Reductor.h"
+
+namespace o2::quality_control_modules::common
+{
+
+/// \brief A Reductor which obtains the most popular characteristics of THnSparse.
+///
+/// A Reductor which obtains the most popular characteristics of THnSparse.
+/// It produces a branch in the format: "mean[NDIM]/D:stddev[NDIM]:entries[NDIM]"
+class THnSparseReductor : public quality_control::postprocessing::Reductor
```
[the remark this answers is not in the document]
iravasen: I have a pull request opened, I will change this. Thanks.

Modules/ITS/include/ITS/TrendingTaskConfigITS.h
```diff
@@ -0,0 +1,56 @@
...
+//
+// See http://alice-o2.web.cern.ch/license for full licensing information.
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+///
+/// \file   TrendingTaskConfig.h
+/// \author Piotr Konopka
```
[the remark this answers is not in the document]
iravasen: I have a pull request opened, I will change this. Thanks.

iravasen:
Hello Piotr, 
thanks a lot for your revision. The list of path will change soon and after this change I do not know at the moment. For the moment I would keep the json and then I think your suggestion it's really good when everything will be finalized. 
Thanks, best,
Ivan

PR #445, opened 2020-07-14
Create a QCDB path manager class (QC-343)

Barthelemy:


Barthelemy:
still to be done: make sure that the class is used everyweher.

knopers8:
Please confirm if it does not break post processing if you plan to merge it before i come back. Post processing config files use MOs and QOs paths directly, so there is a risk that something breaks.

Barthelemy:
I did not plan to merge it before August.

Barthelemy:
Did you approve because you tested the post-processing ?

knopers8:
yes

PR #447, opened 2020-07-15
[EMCAL-657] Trending of Occupancy per Super Module

cterrevo:


cterrevo:
Commit tested and ready to be merged

mfasDa:
The Werror is not so ideal: The cell index object is a well-defined structure represented as a tuple of supermodule, module and eta/phi position in the module. This fully identified the cell, and it is unnecessary to provide helper functions to access either of those information. As we anyhow have to decode the full cell index information we return this. For the purpose in the task we only need to access the supermodule ID but not the other ones, therefore the other variables are unused. The Werror prevents us from using c++17 structured binding. @Barthelemy Any way to circumvent the Werror and still use structured binding?

knopers8:
One way to suppress the warning is to add those lines:
```
(void)mod;
(void)iphi;
(void)ieta;
```

cterrevo:
We commited a possible solution to fix the issue.

PR #451, opened 2020-07-27
Add new trending task for fake-hit rate runs + new reductor for TH2 (Xline reductor) + trending vs run number + new more compact structure for json

iravasen:
- [DONE] new trending task for fake-hit rate runs
- [DONE] new TH2 reductor to save data from TH2 line by line (data from X-projections of TH2 needed for ITS post-processing analysis)
- [DONE] more compact layout of json for fake-hit rate and threshold runs.
- [DONE] adding run numbers on x axis of post-processing plots + modification of json accordingly
- [DONE] run numbers added also for threshold postprocessing-task + modification of json accordingly
- [DONE] comments from Piotr (THnSparseReductor -> THnSparse5Reductor + modification of author name)

Compatibility with trends both vs time and vs runs supported.


Modules/Common/src/TH2XlineReductor.cxx
```diff
@@ -0,0 +1,67 @@
...
+}
+
+/*std::vector<Double_t> TH2XlineReductor::getMean()
+{
+  return mStats.mean;
+}*/
+
+const char* TH2XlineReductor::getBranchLeafList()
+{
+  return " ";
```
knopers8, line 35: Sorry, I am not very fluent with TTrees and TBranches. That can work? If you leave a space in the leaf list, it will somehow inspect this structure with vectors? The TTree will correctly store the ever-growing vectors?

Modules/Common/src/TH2XlineReductor.cxx
```diff
@@ -0,0 +1,67 @@
...
+        sum += histo->GetBinContent(ix, iy);
+      }
+      Double_t meanx = sum / entriesx;
+      mStats.mean.push_back(meanx);       //mean
+      mStats.entries.push_back(entriesx); //entries
+      mStats.mean_scaled.push_back(meanx * 512. * 1024.);
+      sum = 0.;
+      for (int ix = 1; ix <= histo->GetNbinsX(); ix++) {
+        Double_t binc = histo->GetBinContent(ix, iy);
+        if (binc > 0.)
```
knopers8, line 57: same here

Modules/Common/src/TH2XlineReductor.cxx
```diff
@@ -0,0 +1,67 @@
...
+    for (int iy = 1; iy <= histo->GetNbinsY(); iy++) {
+      for (int ix = 1; ix <= histo->GetNbinsX(); ix++) {
+        if (histo->GetBinContent(ix, iy) > 0.)
+          entriesx++;
+        sum += histo->GetBinContent(ix, iy);
+      }
+      Double_t meanx = sum / entriesx;
+      mStats.mean.push_back(meanx);       //mean
+      mStats.entries.push_back(entriesx); //entries
+      mStats.mean_scaled.push_back(meanx * 512. * 1024.);
```
knopers8, line 53: Where do these constants come from? Is it something ITS-specific? If so, could you please move this reductor to the ITS module?

Modules/ITS/include/ITS/TrendingTaskITSFhr.h
```diff
@@ -0,0 +1,107 @@
...
+  void update(Trigger, framework::ServiceRegistry&) override;
+  void finalize(Trigger, framework::ServiceRegistry&) override;
+
+  // other functions (mainly style)
+  void SetLegendStyle(TLegend* leg);
+  void SetGraphStyle(TGraph* g, int col, int mkr);
+  void SetGraphNameAndAxes(TGraph* g, std::string name, std::string title,
+                           std::string xtitle, std::string ytitle, double ymin,
+                           double ymax, std::vector<std::string> runlist);
+  void PrepareLegend(TLegend* leg, int layer);
```
knopers8, line 72: Is there a reason to make them public?

Modules/ITS/src/TrendingTaskITSFhr.cxx
```diff
@@ -0,0 +1,327 @@
...
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+///
+/// \file    TrendingTaskITSFhr.cxx
+/// \author  Ivan Ravasenga on the structure from Piotr Konopka
+///
+
+#include "ITS/TrendingTaskITSFhr.h"
+#include "../../../Framework/src/RootClassFactory.h"
```
knopers8, line 17: I will move it to QualityControl/RootClassFactory.h, so you don't have to use such path.

Modules/ITS/src/TrendingTaskITSFhr.cxx
```diff
@@ -0,0 +1,327 @@
...
+#include "QualityControl/DatabaseInterface.h"
+#include "QualityControl/MonitorObject.h"
+#include "QualityControl/QcInfoLogger.h"
+#include "QualityControl/Reductor.h"
+#include "../../Common/include/Common/TH2XlineReductor.h"
```
knopers8, line 22: #include "Common/TH2XlineReductor.h" doesn't work? If not, I will have to see why.

Modules/ITS/src/TrendingTaskITSFhr.cxx
```diff
@@ -0,0 +1,327 @@
...
+  mDatabase->storeMO(mo);
+}
+
+void TrendingTaskITSFhr::trendValues()
+{
+  // We use current date and time. This for planned processing (not history). We
+  // still might need to use the objects
+  // timestamps in the end, but this would become ambiguous if there is more
+  // than one data source.
+  mTime = TDatime().Convert();
```
knopers8, line 103: Just to let you know, I am preparing a small modification which will add a timestamp to incoming triggers. This will eventually allow to run post-processing on any selection of timestamps, so also older data. Then you can consider taking the time from `trigger.timestamp`.

Modules/ITS/src/TrendingTaskITSFhr.cxx
```diff
@@ -0,0 +1,327 @@
...
+  ilay = 0;
+  countplots = 0;
+  TCanvas* c[NLAYERS * NTRENDSFHR];
+  TLegend* legstaves[NLAYERS];
+  for (int idx = 0; idx < NLAYERS * NTRENDSFHR; idx++) // define canvases
+    c[idx] = new TCanvas(
+      Form("fhr_%s_trends_L%d", trendnames[idx % NTRENDSFHR].c_str(),
+           idx / NTRENDSFHR),
+      Form("fhr_%s_trends_L%d", trendnames[idx % NTRENDSFHR].c_str(),
+           idx / NTRENDSFHR));
```
knopers8, line 199: Please use brackets {}

Modules/Common/src/TH2XlineReductor.cxx
iravasen: This was something I forgot to remove. Actually I'm not using getBranchLeafList() for this kind of reductor because of the std::vector (look at lines 55-61 of TrendingTaskITSFhr.cxx). I didn't find any solution to give a leaf list with std::vectors.

Modules/Common/src/TH2XlineReductor.cxx
iravasen: doing it in a commit at the end

Modules/Common/src/TH2XlineReductor.cxx
iravasen: Yes, it is something specific to the ITS. Ok I move the reductor in the ITS module. I do this in a final commit.

Modules/ITS/include/ITS/TrendingTaskITSFhr.h
iravasen: No, there isn't. Moving it to private in the last commit.

Modules/ITS/src/TrendingTaskITSFhr.cxx
iravasen: Sorry I cannot try anymore since I moved already that reductor to ITS module

Modules/ITS/src/TrendingTaskITSFhr.cxx
iravasen: Very good to know thanks!

Modules/ITS/src/TrendingTaskITSFhr.cxx
```diff
@@ -0,0 +1,327 @@
...
+
+  for (const auto& source : mConfig.dataSources) {
+    std::unique_ptr<Reductor> reductor(root_class_factory::create<Reductor>(
+      source.moduleName, source.reductorName));
+    if (source.reductorName.find("TH2Xline") != std::string::npos) {
+      TH2XlineReductor::mystat* mystc = (TH2XlineReductor::mystat*)reductor->getBranchAddress();
+      mTrend->Branch(Form("%s_mean", source.name.c_str()), &(mystc->mean));
+      mTrend->Branch(Form("%s_stddev", source.name.c_str()), &(mystc->stddev));
+      mTrend->Branch(Form("%s_entries", source.name.c_str()), &(mystc->entries));
+      mTrend->Branch(Form("%s_occupancy", source.name.c_str()), &(mystc->mean_scaled));
```
knopers8, line 60: So, just out of curiosity, what will be stored with branches specified in this manner? Each entry in a branch will contain the full growing vector or perhaps its current last element?

Modules/Common/src/TH2XlineReductor.cxx
knopers8: OK. This is not very clean, but I don't know myself if one can use std::vectors in leaf list. Though, i am pretty sure you can use an array of variable size ( [see "An double array with variable size"](https://root.cern.ch/doc/master/classTTree.html). So perhaps you could use vector.data() and vector.size() there. I guess that would be better than breaking the contract of Reductor.

iravasen: I try to see what I can do. I will be back to you in the next days since I'm also busy with other things.

Modules/ITS/src/TrendingTaskITSFhr.cxx
iravasen: it's the current last element but now that I'm thinking I never clean the vectors. Uhm, let me check this (I'm using just a single run for my tests and I didn't see issues but there could be something strange).

iravasen:
Hi, 
I moved the TH2Xline reductor into ITS module and, inside this reductor, I replaced the std::vector with normal arrays. This simplifies a lot the full chain and keep the role of the reductor inside the frame.

Modules/Common/src/TH2XlineReductor.cxx
iravasen: fixed

Modules/ITS/src/TrendingTaskITSFhr.cxx
iravasen: as agreed I left that path and you will modify it once you will move the file to QualityControl

Modules/ITS/src/TrendingTaskITSFhr.cxx
iravasen: at present I added run numbers on the x axis and we will test the post processing next week with shifts. We will see how to proceed.

knopers8:
Great! Just out of curiosity, couldn't it be done with std::vector and its .data() and size()? I was thinking to give it a try in THnSparseReductor, so it would have no determined limits.

knopers8 (approved):

knopers8:
Travis CI results are doubled again, with one being stuck, another being successful. Other checks are green. Forcing the merger

iravasen:
Thanks for merging! I tried some combinations with .size() and .data() but didn't work. Then I gave up because I had to move on other stuff. One of the main problems I had is that when you do vector.size() in the getBranchLeafList(), the vectors are still empty so the size is 0. Of course this doesn't happen if you have already a dimension (like 5 for the THnSparseReductor). In fact then you see the error when running the post-processing.

knopers8:
I see, thanks!

PR #452, opened 2020-08-06
Implement missing handler in view of abstract DecoderBase

preghenella:
@njacazio 
This PR implements two missing decoder handlers.
This is needed because the functions will be shortly declared as pure virtual in the base class in the O2.


njacazio:
thanks!

knopers8:
Do you need a QC release after this is merged?

preghenella:
I don't think so, @njacazio ?

njacazio:
No, we don't care so far

PR #453, opened 2020-08-07
[QC-347] Error bars for graphs drawn by TrendingTask

knopers8:
Based on the proposition of @cimordas and Marcel Lesch (sorry, I don't remember your nickname).
Here is what has changed:
- "error" changed to "graphErrors" as it only can be applied on graphs,
- added a warning if "graphErrors" are used in a different type of plot,
- errors are not drawn if "graphErrors" is empty, instead being equal to "none" - default behaviour should be invisible,
- the main graph is still drawn by TTree::Draw(), so we get uniform behaviour. TGraphError is used only to draw error bars.
- avoided the possible memory leak by deleting errorGraph.

@cimordas @stheckel Please tell me what you think. I especially wonder if using TGraphError only to draw error bars is okay with you (you might not have influence on how the error bars look like, though we can add another option for that, if needed).

To do:
- update the documentation

### DIFF
--- a/Framework/include/QualityControl/TrendingTaskConfig.h
+++ b/Framework/include/QualityControl/TrendingTaskConfig.h
@@ -36,6 +36,7 @@ struct TrendingTaskConfig : PostProcessingConfig {
     std::string varexp;
     std::string selection;
     std::string option;
+    std::string graphErrors;
   };
 
   struct DataSource {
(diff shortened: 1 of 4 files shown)


cimordas:
Hello,
I am perfectly fine with the TGraphErrors only for the errors themselves. This seems less complicated than what we tried to do.
I think the normal appearance of the error bars will be fine for us. What do you think, @stheckel ?
Best,
Cindy

stheckel:
Hi, I agree with Cindy and think, that the standard appearance of error bars should be fine. In the end, as anyhow one could add an option for this, if needed, this should be ok. Also the renaming and protection make sense to me. Maybe @wiechula could also have a look.

wiechula:
Yes, sounds good so far. The rest we'll see once we have some experience.

knopers8:
Thanks for the input!

PR #456, opened 2020-08-12
[EMCAL-646] Adapt QualityControl to changes in EMCAL raw parsing

mfasDa:
- EMCAL raw parser uses RDHAny and RDHUtils instead
  of a fixed raw version, needs adaption in accessing header
  fields
- RawReaderMemory and AltroDecoder are no longer
  templated due to the change to RDHAny
Change necessary due to the commit in O2: https://github.com/AliceO2Group/AliceO2/pull/4128


mfasDa:
Adapting QualityControl to changes in https://github.com/AliceO2Group/AliceO2/pull/4128. Both changes depend on each other due to the PR checker, therefore they need to be merged in sync.

knopers8:
Two builds, o2/cc7 and o-dataflow/macOS have passed, I don't expect any failures in cc7/o2-dataflow. Merging, as this is urgent to unblock failing CI.

PR #457, opened 2020-08-12
[QC-357] Follow the change of the Data Sampling directory and namespace

knopers8:
This prepares QC for the change of the Data Sampling directory and namespace. It supports both versions - the current one and after the change, so it should be painless if done in right order.

### DIFF
--- a/Framework/CMakeLists.txt
+++ b/Framework/CMakeLists.txt
@@ -95,7 +95,7 @@ target_link_libraries(QualityControl
                              ms_gsl::ms_gsl
                              QualityControlTypes
                              O2::Mergers
-                             O2::DataSampling
+                             $<$<TARGET_EXISTS:O2::DataSampling>:O2::DataSampling>
                       PRIVATE Boost::system
                               $<$<BOOL:${ENABLE_MYSQL}>:MySQL::MySQL>
                               $<$<BOOL:${ENABLE_MYSQL}>:ROOT::RMySQL> ROOT::Gui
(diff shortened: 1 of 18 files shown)


Framework/CMakeLists.txt
```diff
@@ -95,6 +95,7 @@ target_link_libraries(QualityControl
                              ms_gsl::ms_gsl
                              QualityControlTypes
                              O2::Mergers
+                             O2::DataSampling
```
knopers8, line 98: @Barthelemy Is it possible to link the library, but do not complain if it is not found?

Barthelemy: yes, with generator expressions: https://cmake.org/cmake/help/v3.2/manual/cmake-generator-expressions.7.html

Barthelemy: Probably something like `"$<$<O2::DataSampling>:O2::DataSampling>"` (I don't remember exactly)

Barthelemy:
Should I review or wait ?

Framework/CMakeLists.txt
knopers8: Thanks, probably `$<$<TARGET_EXISTS:O2::DataSampling>:O2::DataSampling>` should work. I am giving it a try now

knopers8:
I will fix linking to DataSampling first, no need for the review yet.

knopers8:
It should be OK now, you can review.

Barthelemy (approved):
All good. We should point users to this PR when they are going to make breaking chagnes to O2.

PR #460, opened 2020-08-17
Doc metadata drawoptions

Barthelemy:
I realized that documentation was missing for setDrawOptions and setDisplayHints. I have reshuffled a bit the doc.


doc/Advanced.md
```diff
@@ -45,6 +51,79 @@ For example, if TPC wants to monitor the output `{"TPC", "CLUSTERS"}` of the wor
...
+
+To run it, do: 
+```yaml
+o2-qc-run-producer | o2-qc-run-histo-producer --producers 3 --histograms 3 | o2-qc --config  json://${QUALITYCONTROL_ROOT}/etc/advanced-external-histo.json 
+```
+
+### Limitations
+
+1. Objects sent by the external device must be either a TObject or a TObjArray. In the former case, the object will be sent to the checker encapsulated in a MonitorObject. In the latter case, each TObject of the TObjArray is encapsulated in a MonitorObject and is sent to the checker.
+2. Although we use a query syntax in the external task configuration, we only consider the first element defined there. If there are several sources we ignore all but the first one.
```
knopers8, line 124: Is it still true? I think I have complained about it in some previous PR and you have removed this limiattion.

Barthelemy: indeed, good catch. Removing the limitation.

PR #458, opened 2020-08-14
[QC-386] Add timestamp to post-processing triggers

knopers8:
TODO
- update tests
- document

PR #461, opened 2020-08-18
[QC-388] Use timestamps in TrendingTask

knopers8:
After https://github.com/AliceO2Group/QualityControl/pull/458

This is an example how to retrieve data with use of timestamps. This will allow for running post-processing on older data by using triggers with associated timestamps.
@iravasen you have your own post processing tasks, so you might be interested about this.

### DIFF
--- a/Framework/include/QualityControl/TrendingTask.h
+++ b/Framework/include/QualityControl/TrendingTask.h
@@ -56,9 +56,10 @@ class TrendingTask : public PostProcessingInterface
     Int_t runNumber = 0;
   };
 
-  void trendValues();
-  void storePlots();
-  void storeTrend();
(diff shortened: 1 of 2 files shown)


knopers8:
Hmm, the results should be also stored with the same timestamp. I have to fix that.

knopers8:
@Barthelemy What would be the correct timestamp for post-processing results? Just like in normal MOs and QOs - from the current timestamp to 10 years in the future? Or perhaps from the first timestamp in trend, until the most recent?

Anyway, `DatabaseInterface::storeMO` has to be extended to take timestamps first.

Barthelemy:
could you just fix the conflicts before I review ?

knopers8:
oh, sure

Barthelemy (approved):

PR #462, opened 2020-08-20
[QC-383] Rerunning post-processing with selected timestamps

knopers8:


Framework/include/QualityControl/PostProcessingRunner.h
```diff
@@ -49,6 +49,8 @@ class PostProcessingRunner
   void stop();
   /// \brief Reset transition. Throws on errors.
   void reset();
+  /// \brief Runs the task over selected timestamps, performing the full start, run, stop cycle.
```
Barthelemy, line 52: could you add a line about the expected argument ?

Barthelemy (commented):
just a small comment, otherwise, all good.

Framework/include/QualityControl/PostProcessingRunner.h
knopers8: done

knopers8:
testCcdbDatabase is failing randomly again https://ali-ci.cern.ch/alice-build-logs/AliceO2Group/QualityControl/462/4242a7342aab0dbe355ac2bb9bae4f53edf79c75/build_QualityControl_o2-dataflow/fullLog.txt

Barthelemy:
unfortunately the logs are ok now

Barthelemy (approved):

Framework/include/QualityControl/PostProcessingRunner.h
Barthelemy: ty

knopers8:
Meh, i thought that each build has a different link, so saving it would be enough.

PR #468, opened 2020-08-28
Get Run Number from FairMQ options and pass it to the user (QC-272)

Barthelemy:


Framework/src/TaskRunner.cxx
```diff
@@ -75,7 +78,7 @@ void TaskRunner::init(InitContext& iCtx)
   ILOG(Info) << "initializing TaskRunner" << ENDM;
 
   // registering state machine callbacks
-  iCtx.services().get<CallbackService>().set(CallbackService::Id::Start, [this]() { start(); });
+  iCtx.services().get<CallbackService>().set(CallbackService::Id::Start, [this, &options = iCtx.options()]() { start(options); });
```
knopers8, line 81: This assumes that the reference to `iCtx.options()` will be still valid during the start. This isn't completely safe, but I guess we don't have other choice for now.

Framework/src/TaskRunner.cxx
```diff
@@ -339,15 +350,15 @@ void TaskRunner::startOfActivity()
   mTimerTotalDurationActivity.reset();
   mTotalNumberObjectsPublished = 0;
 
-  Activity activity(mConfigFile->get<int>("qc.config.Activity.number"),
+  Activity activity(mRunNumber,
```
knopers8, line 353: If we don't respect "qc.config.Activity.number" anymore, then we should delete it in config files. Otherwise, we could respect it if mRunNumber == 0.

knopers8 (changes requested):
Thanks! Just to be on the safe side, could you reset mRunNumber to 0 after stop() and reset()? So we accidentaly don't start another run with older run number.

Framework/src/TaskRunner.cxx
Barthelemy: I agree, will do it

Framework/src/TaskRunner.cxx
Barthelemy: Indeed, we don't.

Barthelemy:
I think that I have addressed all your commnets.

Framework/src/TaskRunner.cxx
```diff
@@ -200,8 +203,16 @@ void TaskRunner::endOfStream(framework::EndOfStreamContext& eosContext)
   mNoMoreCycles = true;
 }
 
-void TaskRunner::start()
+void TaskRunner::start(const ConfigParamRegistry& options)
 {
+  try {
+    mRunNumber = options.get<int>("runNUmber");
```
knopers8, line 209: typo "runNUmber"

knopers8 (approved):

Barthelemy:
@ktf @MichaelLettrich It seems that the mac build is stuck

MichaelLettrich:
@Barthelemy builders have been restarted

PR #469, opened 2020-08-31
Attempt at fixing the crashes with testFactory (QC-395)

Barthelemy:
the crash happens in alidist, difficult to say if it is the problem.


Barthelemy:
Merging as it is a test and there is a duplicate travis CI stuck.

PR #471, opened 2020-08-31
Switch to new infologger macro to specify level and severity at once (QC-396)

Barthelemy:


Framework/src/CheckRunner.cxx
```diff
@@ -139,12 +139,13 @@ CheckRunner::CheckRunner(std::vector<Check> checks, std::string configurationSou
...
     mTotalNumberMOStored(0)
 {
+  ILOG.setFacility("Check");
   try {
     mConfigFile = ConfigurationFactory::getConfiguration(configurationSource);
   } catch (...) {
     // catch the exceptions and print it (the ultimate caller might not know how to display it)
-    ILOG(Fatal) << "Unexpected exception during initialization:\n"
-                << boost::current_exception_diagnostic_information(true) << ENDM;
+    ILOG << LogFatalSupport << "Unexpected exception during initialization:\n"
```
knopers8, line 147: I see that here exceptions are logged as fatal, while in Checks they are just errors (however, also fatal in fact). Is there any reason for this difference?

Framework/src/ServiceDiscovery.cxx
```diff
@@ -124,7 +124,7 @@ void ServiceDiscovery::runHealthServer(unsigned int port)
     }
   } catch (std::exception& e) {
     mThreadRunning = false;
-    ILOG(Error) << "ServiceDiscovery::runHealthServer - " << e.what() << ENDM;
+    std::cerr << "ServiceDiscovery::runHealthServer - " << e.what() << std::endl;
```
knopers8, line 127: why cerr?

Framework/src/runPostProcessingOCC.cxx
```diff
@@ -91,10 +91,10 @@ class PostProcessingOCCStateMachine : public RuntimeControlledObject
     try {
       mRunner->start();
     } catch (const std::exception& ex) {
-      ILOG(Error) << "Exception caught: " << ex.what() << ENDM;
+      ILOG << LogErrorSupport << "Exception caught: " << ex.what() << ENDM;
```
knopers8, line 94: Why some catch statements have the "support" severity, while others have "Ops"?

Modules/MFT/src/BasicTrackQcTask.cxx
```diff
@@ -37,11 +37,11 @@ BasicTrackQcTask::~BasicTrackQcTask()
 
 void BasicTrackQcTask::initialize(o2::framework::InitContext& /*ctx*/)
 {
-  ILOG(Info) << "initialize BasicTrackQcTask" << ENDM; // QcInfoLogger is used. FairMQ logs will go to there as well.
+  ILOG << LogInfoSupport << "initialize BasicTrackQcTask" << ENDM; // QcInfoLogger is used. FairMQ logs will go to there as well.
 
   // this is how to get access to custom parameters defined in the config file at qc.tasks.<task_name>.taskParameters
   if (auto param = mCustomParameters.find("myOwnKey"); param != mCustomParameters.end()) {
-    ILOG(Info) << "Custom parameter - myOwnKey: " << param->second << ENDM;
+    ILOG << LogInfoSupport << "Custom parameter - myOwnKey: " << param->second << ENDM;
```
knopers8, line 44: devel

Modules/Skeleton/src/SkeletonTask.cxx
```diff
@@ -31,11 +31,11 @@ SkeletonTask::~SkeletonTask()
 
 void SkeletonTask::initialize(o2::framework::InitContext& /*ctx*/)
 {
-  ILOG(Info) << "initialize SkeletonTask" << ENDM; // QcInfoLogger is used. FairMQ logs will go to there as well.
+  ILOG << LogInfoSupport << "initialize SkeletonTask" << ENDM; // QcInfoLogger is used. FairMQ logs will go to there as well.
 
   // this is how to get access to custom parameters defined in the config file at qc.tasks.<task_name>.taskParameters
   if (auto param = mCustomParameters.find("myOwnKey"); param != mCustomParameters.end()) {
-    ILOG(Info) << "Custom parameter - myOwnKey: " << param->second << ENDM;
+    ILOG << LogInfoSupport << "Custom parameter - myOwnKey: " << param->second << ENDM;
```
knopers8, line 38: devel

knopers8 (changes requested):
Can we have something like this?
```
ILOG(Info, Ops) << "my log" << ENDM;
```
with some defaults for the severity (`support` or `ops` being the default perhaps?)

Barthelemy:
> Can we have something like this?

We can I would probably provide it as a global function rather than a macro. There are variadic parameters for macro nowadays but the resulting code is awful.

Framework/src/runPostProcessingOCC.cxx
Barthelemy: Indeed, I have put all of them with Support

Framework/src/CheckRunner.cxx
Barthelemy: not really, I will try to review the ones that need fatal.

Barthelemy: I will also send them to the Operator. If something does not start or simply stops, they should know.

Framework/src/ServiceDiscovery.cxx
Barthelemy: no idea

knopers8:
> This being said, the number of characters to type is roughly the same (3 characters difference, which I could easily make disappear as I can remove the prefix "Log"). Is there any other arguments in favour of one option or the other ?

1. Having default severities - even less typing.
2. More familiar interface, which resembles the one of fairlogger - less confusion.
3. Just my personal esthetic feelings ;) Having concatenated arguments like "ErrorOps" seems less neat to me than putting them as separate arguments in `ILOG(Error, Ops)`.

However, if we go that way, `Ops` or `Support` keywords shouldn't be just substituted by the preprocessor, because it is likely that someone uses them in other context.

I guess that removing the "Log" could also make it shorter and wouldn't be that bad. I let you decide, with my arguments convincing you or not.

Barthelemy:
Ok, done, ILOG can take 1 or 2 arguments.

Framework/include/QualityControl/QcInfoLogger.h
```diff
@@ -58,14 +62,25 @@ class QcInfoLogger : public AliceO2::InfoLogger::InfoLogger
...
-#define ILOG(severity) ILOG_INST << AliceO2::InfoLogger::InfoLogger::Severity::severity
 #define ILOGI ILOG_INST << AliceO2::InfoLogger::InfoLogger::Info
 #define ILOGW ILOG_INST << AliceO2::InfoLogger::InfoLogger::Warning
 #define ILOGE ILOG_INST << AliceO2::InfoLogger::InfoLogger::Error
 #define ILOGF ILOG_INST << AliceO2::InfoLogger::InfoLogger::Fatal
 #define ENDM AliceO2::InfoLogger::InfoLogger::endm;
 
+#define NUM_ARGS_(_1, _2, _3, _4, _5, _6, _7, _8, TOTAL, ...) TOTAL
+#define NUM_ARGS(...) NUM_ARGS_(__VA_ARGS__, 6, 5, 4, 3, 2, 1, 0)
+#define CONCATE_(X, Y) X##Y
```
knopers8, line 75: concatenate?

knopers8:
Do you know if i can see it working in DPL? I try with adding `--infologger-severity support` to `o2-qc-run-basic`, but all logs seem to be present anyway.

Framework/include/QualityControl/QcInfoLogger.h
Barthelemy: ok

Barthelemy:
Do you mean to have the output of DPL redirected to infologger ? 
I have not tested it directly. I am not sure what is this argument `--infologger-severity`. I tested with FairLogger though.

knopers8:
> Do you mean to have the output of DPL redirected to infologger ?

Probably the other way. To use the ILOG macros and see the logs the terminal, running DPL. `infologger-severity` should filter ot logs which are more "severe" than indicated.

knopers8:
Anyway, this is probably a separate issue, which should be understood by Giulio and Sylvain.

knopers8 (approved):

Barthelemy:
I agree that it is a different issue. I am fixing the format and then merging.

Barthelemy:
@MichaelLettrich There seems to be a problem with the mac builders.

MichaelLettrich:
@Barthelemy there seems to be a problem that git has on macos with the LLVM repo - cloning/ updating sporadically fails. @ktf changed the build scripts in a way that should fix the issue. I relaunched the builders after an update.

PR #476, opened 2020-09-07
Add namespace indicator for runtime_error

knopers8:
This would collide with `o2::framework::runtime_error` after https://github.com/AliceO2Group/AliceO2/pull/4150 is merged


Barthelemy (approved):

Barthelemy:
good catch

PR #477, opened 2020-09-07
Adapt to the new service registry

knopers8:
Needed for https://github.com/AliceO2Group/AliceO2/pull/4150
I still have to see if the thing with references works by accident or it is actually valid. I see that objects returned from CCDB are kind of valid, but they have timestamp "-1". Not sure if this is related.
Also I should update the ITS code.


knopers8:
This works for both versions of ServiceRegistry.

Barthelemy (approved):

PR #478, opened 2020-09-09
Fix TrendingTask test

knopers8:


knopers8:
Trivial, fixes the test, merging.

PR #479, opened 2020-09-14
Improvements to avoid random crash in testFactory

Barthelemy:
I strongly suspect something wrong with the way strings are handled between ObjectsManager and ServiceDiscovery. Copying TaskConfig in ObjectsManager should help.


Barthelemy:
Is InfoLogger thread safe ? I am starting to think that this is the culprit. In ServiceDiscovery we launch a thread and just call our static instance of InfoLogger there. What do you think @sy-c ?

Barthelemy:
So I have mostly made sure that we don't use the same infologger from the ServiceDiscovery health thread and also improved the copies of the config.

knopers8 (approved):

PR #483, opened 2020-09-16
Introduce an UpdatePolicyManager (QC-342)

Barthelemy:
Policies and revisions were spread over `Check` and `CheckRunner` making it difficult to comprehend and to reuse (e.g. in the future aggregators). 

This PR proposes an extraction of these concepts in a separate class called `PolicyManager`. The `Check` class is unaware of the notion of revisions. It still holds its own configuration and thus the policy definition. As a consequence the PolicyManager is indirectly influenced by the way the `Check` class overwrites the policy name in certain cases. This is something we should reconsider in the future. 

I have kept the use of lambdas to implement the actual logic of the different types of policies. Again, this is something we might want to reconsider in the future. The fact that there is a `policyHelper` parameter necessary to hold a state of one of the type of policies is a sign that we might want to move away from simple lambdas. A more traditional OO approach with a common interface and a set of classes implementing the actual policy function could be a solution. 

All in all it is not my dream refactoring but it achieves the extraction of the code related to policies and revisions from `Check` and `CheckRunner` and improves the separation of concerns.


Barthelemy:
No idea what is going on with the CI.

Barthelemy:
@ktf o2-dataflow  is still failing

ktf:
restarted

Framework/test/testPolicyManager.cxx
```diff
@@ -0,0 +1,365 @@
...
+// or submit itself to any jurisdiction.
+
+///
+/// \file   testPolicyManager.cxx
+/// \author Barthelemy von Haller
+///
+
+#include "QualityControl/PolicyManager.h"
+#include <DataSampling/DataSampling.h>
+using namespace o2::utilities;
```
knopers8, line 18: Not needed

knopers8 (changes requested):
Thanks, this looks much better already!

I would propose to change `Policy` to `UpdatePolicy`, since the former is too general IMO. Likewise, `PolicyManager` to `UpdatePolicyManager`

Could you also change the PR name to something more accurate? After all, Policies didn't become classes yet.

In later refactoring, I also thought about making UpdatePolicies return object names which should be checked. That should clean up the `Check::check()` method, which is a monster now. This probably is for later though.

Barthelemy:
Thanks for the great review. I'll update the PR accordingly. 

@MichaelLettrich It seems that the mac builder is broken again.

knopers8:
I have tried manually if the OneEachSeparately policy still works - it seems it does.

Framework/src/PolicyManager.cxx
```diff
@@ -0,0 +1,168 @@
...
+     * Return true if any MOs were updated.
+     * Inner policy - used for `"MOs": "all"`
+     * Might return true even if MO is not used in Check
+     */
+
+    policy = []() {
+      // Expecting check of this policy only if any change
+      return true;
+    };
+
```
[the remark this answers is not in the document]
Barthelemy: I left it in order not to touch the logic that pre-existed between the policies and the Check class. The Check updates the policy depending for example on the "all" parameter and sets it to this _OnGlobalAny. I would leave it for now and wait for the next round of refactoring.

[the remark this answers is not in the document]
knopers8: ok

Framework/CMakeLists.txt
```diff
@@ -65,7 +65,7 @@ add_library(QualityControl
             src/HistoProducer.cxx
             src/DataProducerExample.cxx
             src/MonitorObjectCollection.cxx
-            src/PolicyManager.cxx)
+  src/UpdatePolicyManager.cxx)
```
knopers8, line 68: Please fix the indentation

Framework/CMakeLists.txt
```diff
@@ -121,7 +121,7 @@ add_root_dictionary(QualityControl
                             include/QualityControl/TrendingTask.h
                             include/QualityControl/Reductor.h
                             include/QualityControl/MonitorObjectCollection.h
-                            include/QualityControl/PolicyManager.h
+  include/QualityControl/UpdatePolicyManager.h
```
knopers8, line 124: Please fix the indentation

Framework/src/TaskRunner.cxx
```diff
@@ -280,8 +280,7 @@ void TaskRunner::loadTopologyConfig()
 {
   auto taskConfigTree = getTaskConfigTree();
   auto policiesFilePath = mConfigFile->get<std::string>("dataSamplingPolicyFile", "");
-  ConfigurationInterface* config = policiesFilePath.empty() ? mConfigFile.get() : ConfigurationFactory::getConfiguration(policiesFilePath).get();
-  auto policiesTree = config->getRecursive("dataSamplingPolicies");
+  ConfigurationInterface* config = policiesFilePath.empty() ? mConfigFile.get() : ConfigurationFactory::getConfiguration(policiesFilePath).get(); // todo make sure that it is ok to use the internal pointer of the unique_ptr
```
knopers8, line 283: nice catch, it looks dangerous indeed

knopers8 (approved):
Thanks for the fixes, please just fix the formatting in CMakeLists after CLion (i guess) destroyed it.

PR #486, opened 2020-09-21
[QC-405] Implement the NewObject trigger in Post-processing

knopers8:


Framework/test/testTriggers.cxx
```diff
@@ -65,4 +74,40 @@ BOOST_AUTO_TEST_CASE(test_trigger_once)
...
   BOOST_CHECK_EQUAL(once(), TriggerType::No);
   BOOST_CHECK_EQUAL(once(), TriggerType::No);
-}
\ No newline at end of file
+}
+
+// fixme: this might fail when invoked concurrently.
+//  I keep it here as disabled, it can be used when needed,
+//  but ideally we should find a way to test it continuously.
+/*
```
Barthelemy, line 82: Perhaps you could use the PID in the name of the object to avoid clashes. This is what we do in most database tests if I remember correctly.

Barthelemy (approved):

Framework/test/testTriggers.cxx
knopers8: OK

Barthelemy: I must admit that my memory failed me. I thought I had done it for the database tests but I cannot find it back.

knopers8: I also remember now we had something like that and I can't find it. Not a big deal though.

knopers8: Ah, yes, it is in the functional test.

PR #487, opened 2020-09-22
Fix the warning about DICTPAYLOAD

Barthelemy:


knopers8 (approved):
I see, indeed a lot of these don't need a dictionary.

PR #490, opened 2020-09-23
add ITS online FEE task

JianLIUhep:
Looking into the ITS diagnostic payload to spot the detector errors.


Barthelemy:
the failure is spurious

Modules/ITS/src/ITSFeeTask.cxx
```diff
@@ -0,0 +1,201 @@
...
+    if (mEnableLayers[ilayer]) {
+      ILOG(Info) << "Enable layer : " << ilayer << ENDM;
+    }
+  }
+  configFile >> mRunNumberPath;
+}
+
+void ITSFeeTask::endOfCycle()
+{
+  std::ifstream runNumberFile("/home/its/QC_Online/workdir/infiles/RunNumber.dat"); //catching ITS run number in commissioning
```
Barthelemy, line 167: that is likely to break on any other machine than yours

Modules/ITS/src/ITSFeeTask.cxx
```diff
@@ -0,0 +1,201 @@
...
+  if (runNumberFile) {
+    std::string runNumber;
+    runNumberFile >> runNumber;
+    ILOG(Info) << "runNumber : " << runNumber << ENDM;
+    if (runNumber == mRunNumber) {
```
Barthelemy, line 172: perhaps `if (runNumber != mRunNumber){
    getObjectsManager()->addMetadata(mTFInfo->GetName(), "Run", runNumber);
    getObjectsManager()->addMetadata(mErrorFlagVsFeeId->GetName(), "Run", runNumber);
    getObjectsManager()->addMetadata(mErrorFlag->GetName(), "Run", runNumber);
    mRunNumber = runNumber;
}
I think that it is logically the same but it avoids the `goto`.

Barthelemy (commented):
Very good, I have just a couple of comments. Could you consider them ? Cheers

Modules/ITS/src/ITSFeeTask.cxx
JianLIUhep: Amended. Passing the run number through the cfg for ITS commissioning.

Modules/ITS/src/ITSFeeTask.cxx
JianLIUhep: removed. no need to make the checking here.

Barthelemy (approved):

PR #492, opened 2020-09-28
Daq task dplrawparser

Barthelemy:
A first proper monitoring of raw data, although this is still mostly experimental. 
The class should be renamed at some point (RawDataTask ? ). For now I see it really as the raw data monitoring to happen on FLPs, whether after readout or after STFB. It works in both cases. 
Tested with random data (not much interesting but it does not crash) and with HMP recorded data (all good).

### DIFF
--- a/Modules/Common/src/EverIncreasingGraph.cxx
+++ b/Modules/Common/src/EverIncreasingGraph.cxx
@@ -29,55 +29,55 @@ using namespace std;
 
 namespace o2::quality_control_modules::common
 {
-  void EverIncreasingGraph::configure(std::string /*name*/) {}
+void EverIncreasingGraph::configure(std::string /*name*/) {}
 
-  Quality EverIncreasingGraph::check(std::map<std::string, std::shared_ptr<MonitorObject>> * moMap)
-  {
-    auto mo = moMap->begin()->second;
-    Quality result = Quality::Good;
-    auto* g = dynamic_cast<TGraph*>(mo->getObject());
+Quality EverIncreasingGraph::check(std::map<std::string, std::shared_ptr<MonitorObject>>* moMap)
+{
+  auto mo = moMap->begin()->second;
+  Quality result = Quality::Good;
+  auto* g = dynamic_cast<TGraph*>(mo->getObject());
 
-    // simplistic and inefficient way to check that points are always increasing
-    int nbPoints = g->GetN();
-    double lastY = -DBL_MAX;
-    for (int i = 1; i < nbPoints; i++) {
-      double x, y;
-      g->GetPoint(i, x, y);
-      if (y < lastY) {
-        result = Quality::Bad;
-        break;
-      }
-      lastY = y;
+  // simplistic and inefficient way to check that points are always increasing
+  int nbPoints = g->GetN();
+  double lastY = -DBL_MAX;
+  for (int i = 1; i < nbPoints; i++) {
+    double x, y;
+    g->GetPoint(i, x, y);
+    if (y < lastY) {
+      result = Quality::Bad;
+      break;
     }
-
-    return result;
+    lastY = y;
   }
 
-  std::string EverIncreasingGraph::getAcceptedType() { return "TGraph"; }
+  return result;
+}
 
-  void EverIncreasingGraph::beautify(std::shared_ptr<MonitorObject> mo, Quality checkResult)
-  {
-    ILOG(Info, Support) << "Beautify" << ENDM;
+std::string EverIncreasingGraph::getAcceptedType() { return "TGraph"; }
 
-    if (checkResult == Quality::Null || checkResult == Quality::Medium) {
-      return;
-    }
+void EverIncreasingGraph::beautify(std::shared_ptr<MonitorObject> mo, Quality checkResult)
+{
+  ILOG(Info, Support) << "Beautify" << ENDM;
 
-    auto* g = dynamic_cast<TGraph*>(mo->getObject());
-    if (!g) {
-      ILOG(Error, Support) << "MO should be a graph" << ENDM;
-      return;
-    }
-
-    auto* paveText = new TPaveText(0.3, 0.8, 0.7, 0.95, "NDC");
-    if (checkResult == Quality::Good) {
-      paveText->SetFillColor(kGreen);
-      paveText->AddText("No anomalies");
-    } else if (checkResult == Quality::Bad) {
-      paveText->SetFillColor(kRed);
-      paveText->AddText("Block IDs are not always increasing");
-    }
-    g->GetListOfFunctions()->AddLast(paveText);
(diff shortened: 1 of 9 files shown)


Modules/Common/src/EverIncreasingGraph.cxx
```diff
@@ -0,0 +1,83 @@
...
+
+namespace o2::quality_control_modules::common
+{
+void EverIncreasingGraph::configure(std::string /*name*/) {}
+
+Quality EverIncreasingGraph::check(std::map<std::string, std::shared_ptr<MonitorObject>>* moMap)
+{
+  auto mo = moMap->begin()->second;
+  Quality result = Quality::Good;
+  auto* g = dynamic_cast<TGraph*>(mo->getObject());
```
knopers8, line 38: check for nullptr if dynamic_cast fails

Modules/Common/src/EverIncreasingGraph.cxx
```diff
@@ -0,0 +1,83 @@
...
+
+  // simplistic and inefficient way to check that points are always increasing
+  int nbPoints = g->GetN();
+  double lastY = -DBL_MAX;
```
knopers8, line 42: Consider this, if you agree it is more clear and safer:
```
std::numeric_limits<double>::lowest
```
https://en.cppreference.com/w/cpp/types/numeric_limits/lowest says it actually returns `-DBL_MAX`

Modules/Common/src/EverIncreasingGraph.cxx
```diff
@@ -0,0 +1,83 @@
...
+    return;
+  }
+
+  auto* paveText = new TPaveText(0.3, 0.8, 0.7, 0.95, "NDC");
+  if (checkResult == Quality::Good) {
+    paveText->SetFillColor(kGreen);
+    paveText->AddText("No anomalies");
+  } else if (checkResult == Quality::Bad) {
+    paveText->SetFillColor(kRed);
+    paveText->AddText("Block IDs are not always increasing");
```
knopers8, line 78: These aren't necessarily block IDs, especially now, as you have moved this check to `Common`.

Modules/Daq/include/Daq/DaqTask.h
```diff
@@ -17,18 +17,15 @@
...
+#include <Headers/DAQID.h>
+#include <map>
 
 class TH1F;
-class TGraph;
-class TObjString;
-class TCanvas;
-class TPaveText;
 
 using namespace o2::quality_control::core;
```
knopers8, line 25: We shouldn't actually use `using namespace` in headers, but this is prevalent among Tasks in modules. However, it is not a big deal, because we load the classes with ROOT and these headers do not pollute anything.

Modules/Daq/include/Daq/DaqTask.h
```diff
@@ -52,12 +49,32 @@ class DaqTask final : public TaskInterface
...
+  std::map<o2::header::DAQID::ID, std::string> mSystems;
+
+  // ** objects we publish **
+
+  // Message related
+  // Block = the whole InputRecord, i.e. the thing we receive and analyse in monitorData(...)
+  // SubBlock = a single input of the InputRecord
+  TH1F* mInputRecordPayloadSize; // filled with the sum of the payload size of all the inputs of an inputrecord
+  TH1F* mNumberInputs;           // filled with the number of inputs in each InputRecord we encounter
+  TH1F* mInputSize;              // filled with the number of inputs in each InputRecord we encounter
```
knopers8, line 67: The comment seems to be a duplicate of the one above.

Modules/Daq/include/Daq/DaqTask.h
```diff
@@ -52,12 +49,32 @@ class DaqTask final : public TaskInterface
...
+  TH1F* mNumberRDHs;             // filled with the number of RDHs found in each InputRecord we encounter
```
knopers8, line 68: You might want to use ` = nullptr`, instead of setting the pointers in constructors. It is said to be faster to compile and I think it is clearer & easier to maintain.

Modules/Daq/src/DaqTask.cxx
```diff
@@ -15,140 +15,249 @@
...
+void DaqTask::startOfActivity(Activity& activity)
 {
-  QcInfoLogger::GetInstance() << "startOfActivity" << AliceO2::InfoLogger::InfoLogger::endm;
-  mPayloadSize->Reset();
-  mNumberSubblocks->Reset();
-  mSubPayloadSize->Reset();
-  mIds->Set(0);
-  mNPoints = 0;
+  ILOG(Info, Support) << "startOfActivity: " << activity.mId << ENDM;
+  reset(); // TODO make sure this is what we want actually, but most probably it is
```
knopers8, line 104: It shouldn't have any real influence, because newly created histograms shouldn't contain any data and during stop()  we do:
```
  endOfActivity();
  mTask->reset();
```

Modules/Daq/src/DaqTask.cxx
```diff
@@ -15,140 +15,249 @@
...
+      for (size_t word = 0; word < 2; word++) {
+        if (i + col * 2 + word < representation.size()) {
+          ILOG(Info) << representation[i + col * 2 + word];
+        } else {
+          ILOG(Info) << "   ";
+        }
+      }
+      ILOG(Info) << " | ";
+    }
+    ILOG(Info) << ENDM;
```
knopers8, line 167: Here we print with severity `Support`, but the page is printed with `Ops`. Is there a reason to make it inconsistent?

knopers8 (changes requested):
To reduce the amount of objects coming to Mergers, we might allocate a system only after we receive data from certain detector. It is not crucial now, but we should perhaps do it before it hits us back.

Otherwise, i think it is good, I have some minor suggestions.

Barthelemy:
Thank you for the review, I will update with the required changes. 

Concerning decreasing the amount of plots being unnecessarily sent, I agree and I had already logged it in https://alice.its.cern.ch/jira/browse/QC-425. 

A more "deterministic" ways of implementing it would be to know which systems are part of a given activity. We used to have that in AMORE with the logbook informing us about the detectors present in a given run. 

Would it be ok if I do it in a different PR ?

Modules/Common/src/EverIncreasingGraph.cxx
Barthelemy: thanks, this is nice.

Modules/Daq/include/Daq/DaqTask.h
Barthelemy: you are abolutely correct. I have removed it because we should try to be consistent.

Barthelemy:
I have applied all the modifications you proposed, except the one to limit the number of histograms that I propose to tackle in a separate PR.

knopers8:
Sure, that one can be addressed later.

knopers8 (approved):

PR #498, opened 2020-10-01
[QC-420] Integrate post-processing with DPL

knopers8:
TODO
- merge QC-419
- make "tasks", "checks", "dataSamplingPolicies" and "postprocessing" structures optional in config files
- fix the problem with DPL when two PP tasks are running and finishing one causes the other one to finish
- documentaiton

### DIFF
--- a/Framework/src/PostProcessingDevice.cxx
+++ b/Framework/src/PostProcessingDevice.cxx
@@ -1,0 +1,133 @@
+// Copyright CERN and copyright holders of ALICE O2. This software is
+// distributed under the terms of the GNU General Public License v3 (GPL
+// Version 3), copied verbatim in the file "COPYING".
+//
+// See http://alice-o2.web.cern.ch/license for full licensing information.
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+///
+/// \file   PostProcessingDevice.cxx
+/// \author Piotr Konopka
+///
+
+#include "QualityControl/PostProcessingDevice.h"
+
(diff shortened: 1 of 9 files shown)


knopers8:
All done, but we need https://github.com/AliceO2Group/AliceO2/pull/4506 merged first

knopers8:
Hmm, I tihnk that double Travis appears when [WIP] is removed. The previous test stays in the list and a new one is added. I will make a fake PR to confirm.

Framework/src/PostProcessingDevice.cxx
```diff
@@ -0,0 +1,133 @@
...
+
+  // registering state machine callbacks
+  ctx.services().get<CallbackService>().set(CallbackService::Id::Start, [this]() { start(); });
+  ctx.services().get<CallbackService>().set(CallbackService::Id::Stop, [this]() { stop(); });
+  ctx.services().get<CallbackService>().set(CallbackService::Id::Reset, [this]() { reset(); });
+}
+
+void PostProcessingDevice::run(framework::ProcessingContext& ctx)
+{
+  // todo: enable once check runners can store the results.
```
Barthelemy, line 55: is this todo relevant to this class ?

Barthelemy (approved):

Framework/src/PostProcessingDevice.cxx
knopers8: weird, i must have messed something up with the commits, since it is not enabled

knopers8: ah no, it is fine. This part is done in the next PR, at this stage we still store objects directly to the repository inside PostProcessing, because we don't send them to checkers yet. This part is correct.

PR #505, opened 2020-10-06
Add missing ENDM for logging the completion action in TaskRunner

knopers8:


knopers8:
i commented it out, because it would be produced just as often as that log message above, which was considered too much.

Barthelemy:
Thank you, I was actually writing a comment to ask and comment it out until we prune the Trace level messages using the new infologger version.

Barthelemy (approved):

PR #506, opened 2020-10-06
[QC-421] Allow to check PP tasks results in CheckRunners

knopers8:
Done, but needs QC-420 merged first.

### DIFF
--- a/Framework/src/Check.cxx
+++ b/Framework/src/Check.cxx
@@ -25,6 +25,7 @@
 #include "QualityControl/TaskRunner.h"
 #include "QualityControl/InputUtils.h"
 #include "QualityControl/RootClassFactory.h"
+#include "QualityControl/PostProcessingDevice.h"
 // Fairlogger
 #include <fairlogger/Logger.h>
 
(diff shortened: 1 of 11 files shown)


knopers8:
@Barthelemy Could you have a look when you have some time?

Framework/src/Check.cxx
```diff
@@ -88,12 +90,15 @@ void Check::initConfig(std::string checkName)
   mNumberOfTaskSources = 0;
   for (const auto& [_key, dataSource] : checkConfig.get_child("dataSource")) {
     (void)_key;
-    if (dataSource.get<std::string>("type") == "Task" || dataSource.get<std::string>("type") == "ExternalTask") {
+    if (auto sourceType = dataSource.get<std::string>("type");
+        sourceType == "Task" || sourceType == "ExternalTask" || sourceType == "PostProcessing") {
```
Barthelemy, line 94: I see what you mean with "it is not very elegant all these OR conditions" however I am glad to know that we can declare a variable inside a if clause, this is new to me.

Barthelemy (approved):

Framework/src/Check.cxx
knopers8: This is since c++17

PR #507, opened 2020-10-07
Try to reproduce the issue with double travis

knopers8:
1. Submit a PR with the [WIP] tag.
2. Wait for the result of Travis
3. Remove [WIP]


knopers8:
hmm, it seems this is not the way to reproduce it.

PR #510, opened 2020-10-08
Adapt to usage of ConstMCLabelContainer in TPC

davidrohr:
@Barthelemy @sawenzel : There is an incompatible change in the reading of TPC data in AliceO2Group/AliceO2#4529, which requires this PR in QualityControl. Can we create a new QualityControl tag with this PR, then I would force merge the alidist bump and the O2 PR.

### DIFF
--- a/Modules/TPC/src/Clusters.cxx
+++ b/Modules/TPC/src/Clusters.cxx
@@ -24,10 +24,6 @@
 
 #include <bitset>
 
-using namespace o2::framework;
-using namespace o2::header;
-using namespace o2::tpc;
-
 // root includes
 #include <TCanvas.h>
 #include <TH1.h>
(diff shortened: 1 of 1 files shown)


Barthelemy:
@davidrohr ok. I will force merge and release a new patch version. There are simple and safer, if not elegant, ways of handling incompatible changes for a future case. Perhaps something along the line of : https://github.com/AliceO2Group/QualityControl/pull/457/files

davidrohr:
@Barthelemy : thx! I know one can do it in a safer way but then it's a multi-stage process with multiple alidist builds, so I wouldn't say it's simpler per se. Anyway, I'll follow up the next CI runs and clean up the mess in case I broke something :)

PR #513, opened 2020-10-12
Use o2-qc instead of o2-qc-run-qc in the doc

knopers8:


Barthelemy (approved):

knopers8:
macOS log is missing, changes are only in the documentation, safe to merge

PR #516, opened 2020-10-15
[QC-449] Avoid warning about ptr exposed which will be destroyed

Barthelemy:
> error: object backing the pointer will be destroyed at the end of the full-expression [-Werror,-Wdangling-gsl]


Barthelemy:
Seen here: https://github.com/AliceO2Group/QualityControl/pull/515

Barthelemy:
Perhaps the DataSampling could take shared_ptr in its interface instead / in addition to the raw pointer.

PR #519, opened 2020-10-16
[QC-452] Add extra monitoring metrics about incoming data

Barthelemy:
To be used in the dataflow dashboard

### DIFF
--- a/Framework/src/TaskRunner.cxx
+++ b/Framework/src/TaskRunner.cxx
@@ -132,7 +132,7 @@ void TaskRunner::run(ProcessingContext& pCtx)
 
   if (dataReady) {
     mTask->monitorData(pCtx);
-    mNumberMessages++;
+    updateMonitoringStats(pCtx);
   }
 
   if (timerReady) {
@@ -396,8 +396,9 @@ void TaskRunner::startCycle()
 {
   QcInfoLogger::GetInstance() << "cycle " << mCycleNumber << " in " << mTaskConfig.taskName << ENDM;
   mTask->startOfCycle();
-  mNumberMessages = 0;
+  mNumberMessagesReceivedInCycle = 0;
   mNumberObjectsPublishedInCycle = 0;
+  mDataReceivedInCycle = 0;
   mTimerDurationCycle.reset();
   mCycleOn = true;
 }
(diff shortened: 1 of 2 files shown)


Framework/src/TaskRunner.cxx
```diff
@@ -421,16 +422,32 @@ void TaskRunner::finishCycle(DataAllocator& outputs)
   }
 }
 
+void TaskRunner::updateMonitoringStats(ProcessingContext& pCtx)
+{
+  mNumberMessagesReceivedInCycle++;
+  for (auto&& input : pCtx.inputs()) {
```
knopers8, line 428: this might not take into account multi-part messages. I would use InputRecordWalker instead.

Framework/src/TaskRunner.cxx
```diff
@@ -421,16 +422,32 @@ void TaskRunner::finishCycle(DataAllocator& outputs)
...
+    if (input.header != nullptr) {
+      const auto* header = o2::header::get<header::DataHeader*>(input.header);
```
knopers8, line 430: i am not sure if DataHeader is guaranteed for all non-empty messages. could you add a check for that? if there is no data header, a warning probably should be enough.

knopers8 (changes requested):

knopers8:
testccdbdatabase has just failed on o2-dataflow, it seems the issue is still there.
[testccdb.txt](https://github.com/AliceO2Group/QualityControl/files/5399966/testccdb.txt)

Framework/src/TaskRunner.cxx
Barthelemy: yes, absolutely, fixing it.

Barthelemy:
Pushed a commit that should address the two points. Thank you for the logs of the failing test.

Framework/src/TaskRunner.cxx
```diff
@@ -421,16 +423,36 @@ void TaskRunner::finishCycle(DataAllocator& outputs)
   }
 }
 
+void TaskRunner::updateMonitoringStats(ProcessingContext& pCtx)
+{
+  mNumberMessagesReceivedInCycle++;
+  for (const auto& input : InputRecordWalker(pCtx.inputs())) {
+    if (input.header != nullptr) {
```
knopers8, line 430: InputRecordWalker walks only on non-empty headers anyway, no need for that check.

knopers8 (approved):

PR #520, opened 2020-10-19
[QC-454] Fix testCcdbDatabase 

Barthelemy:
that can fail due to the metadata of the object being spurriously different (Date)


knopers8 (approved):
interesting, so "Date" shows the time of retrieval, not storage?

Barthelemy:
Yes, `Created` and `Last-modified` are specific to the object.

PR #521, opened 2020-10-19
[QC-455] retrieveConditionAny in users' tasks

Barthelemy:


knopers8 (approved):
Thanks, indeed we needed that. Though I am a bit worried that depending on detector libraries will hit us back some day.

Barthelemy:
@knopers8 that is true, my intention was to test exactly what Markus needed. I could iterate on it and use a custom class from the DAQ module (we need a dictionary which we don't have in the tests!).

PR #523, opened 2020-10-23
Align TOF diagnostic naming scheme to O2

njacazio:
- Fix diagnostic and compressed raw data reader tasks [ALICETOF-29] 
  - Using correct loop when reading raw data
- Simplified Counter class
  - Remove intermediate class

### DIFF
--- a/Modules/TOF/src/TaskCompressedData.cxx
+++ b/Modules/TOF/src/TaskCompressedData.cxx
@@ -202,15 +202,20 @@ void TaskCompressedData::startOfCycle()
 
 void TaskCompressedData::monitorData(o2::framework::ProcessingContext& ctx)
 {
-  /** receive input **/
-  for (auto& input : ctx.inputs()) {
-    /** input **/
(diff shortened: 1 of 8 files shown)


njacazio:
@preghenella, now using O2 labelling!

Modules/TOF/src/TaskCompressedData.cxx
```diff
@@ -202,15 +202,20 @@ void TaskCompressedData::startOfCycle()
...
-    auto payloadInSize = headerIn->payloadSize;
-    mDecoder.setDecoderBuffer(payloadIn);
-    mDecoder.setDecoderBufferSize(payloadInSize);
-    mDecoder.decode();
+  for (auto iit = ctx.inputs().begin(), iend = ctx.inputs().end(); iit != iend; ++iit) {
+    if (!iit.isValid()) {
+      continue;
+    }
+    /** loop over input parts **/
+    for (auto const& input : iit) {
```
knopers8, line 210: Maybe you are already aware, but there is Framework/InputRecordWalker.h which does exactly that, I think

knopers8 (approved):

PR #527, opened 2020-10-27
[QC-463] Daqtask only publish included det

Barthelemy:


knopers8 (approved):
Thanks! The code will grow in a bit ugly way if the number of histogram kinds grows, but we can address that later.

Barthelemy:
I agree on the fact that it should be rewritten when the number of plots grows.

knopers8:
We could have some kind of extended map, which automatically creates and registers histograms if you try to access a new key.

Barthelemy:
yes

PR #528, opened 2020-10-27
No need to check if O2::DataSampling exists

knopers8:
The changes to O2, which move Data Sampling to separate library, have been already merged.


Barthelemy:
macos test unrelated

Barthelemy (approved):

PR #529, opened 2020-10-27
Deactivate tests that relied on old data

Barthelemy:
Following the failure of ccdb-test, the data in previous format is lost. Deactivating these tests until we decide if we scrap them or try to recreate the data.


knopers8 (approved):

knopers8:
macOS build fails on CCDB api test in O2, which of course has to be broken. force-merging

PR #530, opened 2020-10-28
[EMCAL-524] Skip SOX headers

mfasDa:
Skip SOX headers (start of run, page containing no
payload) as they will crash the raw parsing

### DIFF
--- a/Modules/EMCAL/src/RawTask.cxx
+++ b/Modules/EMCAL/src/RawTask.cxx
@@ -246,6 +246,12 @@ void RawTask::monitorData(o2::framework::ProcessingContext& ctx)
       mHistogram->Fill(header->payloadSize);
       mTotalDataVolume->Fill(1., header->payloadSize);
 
+      // Skip SOX headers
+      auto rdhblock = reinterpret_cast<const o2::header::RDHAny*>(input.payload);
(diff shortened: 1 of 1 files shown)


Barthelemy:
@ktf @TimoWilken The mac build seems to have failed.

TimoWilken:
@Barthelemy Sorry, I don't have access to the mac machines and Giulio is on holiday today. Maybe @MichaelLettrich does?

Barthelemy:
@mfasDa There is a genuine error on the mac build.

mfasDa:
Indeed I tested on Linux which seemed to allow the implicit cast. I will push a fix for mac in a couple of minutes.

Modules/EMCAL/src/RawTask.cxx
```diff
@@ -281,7 +287,8 @@ void RawTask::monitorData(o2::framework::ProcessingContext& ctx)
           first = false;
         }
         if (feeID > 40)
```
Barthelemy, line 289: we advise to "use curly braces even when the body of the statement fits on one line". I did not catch it before. I will approve this PR but you can keep it in your mind.

Barthelemy (approved):

PR #535, opened 2020-11-02
Add ITS online Track task

JianLIUhep:


Modules/ITS/src/ITSTrackTask.cxx
```diff
@@ -0,0 +1,222 @@
...
+
+  delete mGeom;
+}
+
+void ITSTrackTask::initialize(o2::framework::InitContext& /*ctx*/)
+{
+
+  QcInfoLogger::GetInstance() << "initialize ITSTrackTask" << AliceO2::InfoLogger::InfoLogger::endm;
+  publishHistos();
+  std::string dictfile = o2::base::NameConf::getDictionaryFileName(o2::detectors::DetID::ITS, "", ".bin");
```
Barthelemy, line 57: how will you guarantee that you have this file in production ? or will you get this dictionary from the CCDB in production ?

Modules/ITS/src/ITSTrackTask.cxx
```diff
@@ -0,0 +1,222 @@
...
+    hClusterUsage->SetBinContent(1, 1. * mNClustersInTracks / mNClusters);
+    mNClustersInTracks = 0;
+    mNClusters = 0;
+  }
+}
+
+void ITSTrackTask::endOfCycle()
+{
+
+  std::ifstream runNumberFile("/home/its/QC/workdir/infiles/RunNumber.dat"); //catching ITS run number in commissioning
```
Barthelemy, line 122: Why don't you do it at startOfActivity ?

Modules/ITS/src/ITSTrackTask.cxx
```diff
@@ -0,0 +1,222 @@
...
+  addObject(hClusterUsage);
+  formatAxes(hClusterUsage, "", "nCluster in track / Total cluster", 1, 1.10);
+}
+
+void ITSTrackTask::getEnableLayers()
+{
+  std::ifstream configFile("Config/ConfigLayers.dat"); //passing enabled layers into QC for ITS commissioning
+  for (int ilayer = 0; ilayer < NLayer; ilayer++) {
+    configFile >> mEnableLayers[ilayer];
+    if (mEnableLayers[ilayer])
```
Barthelemy, line 192: we recommend to "use curly braces even when the body of the statement fits on one line"

Modules/ITS/src/ITSTrackTask.cxx
```diff
@@ -0,0 +1,222 @@
...
+      LOG(INFO) << "enable layer : " << ilayer;
+  }
+}
+
+void ITSTrackTask::addObject(TObject* aObject)
+{
+  if (!aObject) {
+    LOG(INFO) << " ERROR: trying to add non-existent object ";
+    return;
+  } else
```
Barthelemy, line 202: we recommend to "use curly braces even when the body of the statement fits on one line"

Barthelemy:
Good, I have just a couple of questions and noticed a few missing curly braces.

Modules/ITS/src/ITSTrackTask.cxx
JianLIUhep: The dictionary will be created by the ITS reconstruction device. I have no idea how to get it in production at the moment but will confirm that with reconstruction people.

Modules/ITS/src/ITSTrackTask.cxx
JianLIUhep: It's a workaround for our commissioning as the RunNumber file is created after the QC workflow already started. It would be completely different in production on how to get the ITS run number.

Modules/ITS/src/ITSTrackTask.cxx
JianLIUhep: modified.

Modules/ITS/src/ITSTrackTask.cxx
JianLIUhep: modified

Modules/ITS/src/ITSTrackTask.cxx
Barthelemy: Ok, it can stay like that for now but perhaps add a "todo" or a ticket in JIRA to remember to update it to something else for production. You could perhaps use the CCDB ?

Modules/ITS/src/ITSTrackTask.cxx
Barthelemy: ok

Barthelemy (approved):

Modules/ITS/src/ITSTrackTask.cxx
JianLIUhep: I guess so. I will discuss it with Jouri. The ticket was created: https://alice.its.cern.ch/jira/browse/QC-467
Thank you.

PR #536, opened 2020-11-02
Add FT0 module

Kavaldrin:
As requested, first part of QC for FT0 detector.
It is worth to mention that there is a still unsolved issue which may affect this code.
https://alice.its.cern.ch/jira/browse/QC-459


Modules/FT0/include/FT0/ChannelsCheck.h
```diff
@@ -0,0 +1,44 @@
...
+// See http://alice-o2.web.cern.ch/license for full licensing information.
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+///
+/// \file   ChannelsCheck.h
+/// \author Milosz Filus
+/// Dummy check for FT0 detector
```
Barthelemy, line 14: Is it really a dummy check ?

Modules/FT0/include/FT0/DigitsCheck.h
```diff
@@ -0,0 +1,46 @@
...
+// See http://alice-o2.web.cern.ch/license for full licensing information.
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+///
+/// \file   ChannelsCheck.h
+/// \author Milosz Filus
+/// Another dummy check for FT0 detector
```
Barthelemy, line 14: SAme question

Barthelemy (commented):
Hi, 

thank you. 

I have a question: why not fill mChargeHistogram directly in the task ? 
it seems like an extra step to first populate a TTree with every single event you get and then process the tree to populate the histogram. 
Moreover, it seems that there mChargeHistogram in both the Task and the PostProcessing.

Otherwise I don't really have comments on the code itself.

Modules/FT0/include/FT0/ChannelsCheck.h
Kavaldrin: Hello,
all of these checks will be probably changed by experts who know what to check in this data. That is why I called them dummy checks.
The same case for monitoring objects.

The post processing histogram generation is just to show how to read tree from database and then do some stuff on that tree.

Kavaldrin:
Hello Barth,
all of these check will be probably modified / replaced by experts who know how to check that data in a proper way.
The same thing for monitoring object.
My intention was to create easy start for them, that they can check how to do some things based on this code.
Also pp task is here just to present how to read tree from database and then do some stuff on this tree and the example is histogram generation.

Modules/FT0/multinode-config.json
```diff
@@ -0,0 +1,106 @@
+{
```
knopers8, line 1: Hi, I saw on todays slides by Alla that you had a multinode workflow where you merge *sampled data*, instead of *monitor objects*. Is that really true, or that was some kind of mistake in slides?

Kavaldrin: To be honest I'm a bit confused, because from generated graph https://i.imgur.com/sCPfHVe.jpg (using -g option) there is no task execution on local node (which would create monitoring object), but merger works on monitoring objects, so input of merger should be a monitoring object. Well I need to look on implementation because it looks like I'm missing something.

Barthelemy: Let us know if you need some clarifications.

knopers8: See the slide 5, this is how it is supposed to look like in the big picture:
 https://indico.cern.ch/event/873380/contributions/3850119/attachments/2036439/3409783/WP7plenaryMay.pdf
Some parallel proxies are replaced with one, as far as I remember, but otherwise it should match what you get (disregarding devices generated automatically by DPL).

Kavaldrin: @knopers8 
Thank you for the slide with explanation.
Right it seems I made a mistake in a slide, there is a task execution and monitoring object creation before sending it to remote device, but I do not undestand why the generated graph does not contain task runner.
Maybe adding -g option ends workflow generation before attaching task runner to workflow?

knopers8: I doubt it. I've just crosschecked on my machine and the proxy after dispatcher is still created. 

Does your machine name (or the --host argument) match one from your list of machines in a data sampling policy? A proxy for sampling policies won't be created on machines which don't need it.

Kavaldrin: @knopers8 Well I will update multinode-config to my current version, because I completely removed data sampling policies + remote task.

knopers8: Yes, so to make your local task BasicDigitQcTask appear, you have to run QC with `--local --host 2a` (or 3a), so QC knows that this is the proper machine to load this task. I am not sure however, how far can you get away with mocking the names instead of using real machines.

knopers8 (approved):

PR #537, opened 2020-11-03
[QC-402] Initial support and documentation for running QC with Analysis

knopers8:
@Barthelemy The new cmake part seems to work, but I don't know if this is the safest and cleanest way to do this.

### DIFF
--- a/Modules/Example/etc/analysisDerived.json
+++ b/Modules/Example/etc/analysisDerived.json
@@ -1,0 +1,55 @@
+{
+  "qc": {
+    "config": {
+      "database": {
+        "implementation": "CCDB",
+        "host": "ccdb-test.cern.ch:8080",
+        "username": "not_applicable",
+        "password": "not_applicable",
+        "name": "not_applicable"
+      },
+      "Activity": {
+        "number": "42",
+        "type": "2"
+      },
+      "monitoring": {
+        "url": "infologger:///debug?qc"
+      },
+      "consul": {
+        "url": "http://consul-test.cern.ch:8500"
+      },
+      "conditionDB": {
+        "url": "ccdb-test.cern.ch:8080"
+      }
+    },
+    "tasks": {
+      "AnalysisQcTask": {
+        "active": "true",
+        "className": "o2::quality_control_modules::example::AnalysisTask",
+        "moduleName": "QcExample",
+        "detectorName": "TST",
+        "cycleDurationSeconds": "10",
+        "maxNumberCycles": "-1",
+        "dataSource": {
+          "type": "direct",
+          "query": "aod-data:AOD/HASH/0"
+        },
+        "location": "remote"
+      }
+    },
+    "checks": {
+      "DummyAnalysisCheck": {
+        "active": "true",
+        "className": "o2::quality_control_modules::skeleton::SkeletonCheck",
+        "moduleName": "QcSkeleton",
+        "policy": "OnAny",
+        "detectorName": "TST",
+        "dataSource": [{
+          "type": "Task",
+          "name": "AnalysisQcTask",
+          "MOs": ["example"]
+        }]
+      }
+    }
+  }
+}
--- a/cmake/AddQCWorkflow.cmake
+++ b/cmake/AddQCWorkflow.cmake
@@ -58,7 +58,7 @@ function(o2_add_qc_workflow)
   endif()
   message(STATUS "o2_add_qc_workflow called with config file ${PARSED_ARGS_CONFIG_FILE_PATH} to generate ${PARSED_ARGS_WORKFLOW_NAME}")
 
-  set(jsonDumpFile ${CMAKE_INSTALL_PREFIX}/share/dpl/${PARSED_ARGS_WORKFLOW_NAME}.json)
+  set(jsonDumpFile ${CMAKE_CURRENT_SOURCE_DIR}/${PARSED_ARGS_WORKFLOW_NAME}.json)
   message(STATUS "jsonDumpFile: ${jsonDumpFile}" )
   set(qcExecutable o2-qc)
 
--- a/cmake/AddQCWorkflow.cmake
+++ b/cmake/AddQCWorkflow.cmake
@@ -58,7 +58,7 @@ function(o2_add_qc_workflow)
   endif()
   message(STATUS "o2_add_qc_workflow called with config file ${PARSED_ARGS_CONFIG_FILE_PATH} to generate ${PARSED_ARGS_WORKFLOW_NAME}")
 
-  set(jsonDumpFile ${CMAKE_CURRENT_SOURCE_DIR}/${PARSED_ARGS_WORKFLOW_NAME}.json)
+  set(jsonDumpFile ${CMAKE_INSTALL_PREFIX}/share/dpl/${PARSED_ARGS_WORKFLOW_NAME}.json)
   message(STATUS "jsonDumpFile: ${jsonDumpFile}" )
   set(qcExecutable o2-qc)
 
--- a/cmake/AddQCWorkflow.cmake
+++ b/cmake/AddQCWorkflow.cmake
@@ -58,17 +58,16 @@ function(o2_add_qc_workflow)
   endif()
   message(STATUS "o2_add_qc_workflow called with config file ${PARSED_ARGS_CONFIG_FILE_PATH} to generate ${PARSED_ARGS_WORKFLOW_NAME}")
 
-  set(jsonDumpFile ${CMAKE_CURRENT_BINARY_DIR}/${PARSED_ARGS_WORKFLOW_NAME}.json)
+  set(jsonDumpFile ${CMAKE_CURRENT_SOURCE_DIR}/${PARSED_ARGS_WORKFLOW_NAME}.json)
   message(STATUS "jsonDumpFile: ${jsonDumpFile}" )
   set(qcExecutable o2-qc)
 
   add_custom_command(
     OUTPUT ${jsonDumpFile}
     VERBATIM
-    COMMAND echo "${qcExecutable} -b --config json:/${CMAKE_INSTALL_PREFIX}/${PARSED_ARGS_CONFIG_FILE_PATH} --dump-workflow --dump-workflow-file ${jsonDumpFile}"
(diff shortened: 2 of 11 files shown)


knopers8:
There is something wrong with generation of the DPL dumps, the errors are legits.

knopers8:
I don't understand what is the cause of these errors. The workflows crash when generating DPL dump files, but in a weird place. The failed assertation on mac might be a clue:
```
Assertion failed: (infos.size() == deviceSpecs.size()), function runStateMachine, file /System/Volumes/Data/build/ci_checks/qualitycontrol_checker_macos_o2-dataflow_ci_0/sw/SOURCES/O2/ae5b8dc89d_O2_DATAFLOW/ae5b8dc89d/Framework/Core/src/runDataProcessing.cxx, line 1275.
```
Still, it works always on my machine, even when running two such workflows in parallel.

Barthelemy:
I am going to try on my mac

Barthelemy:
Same on my Mac. The command it tries to execute is : 
```
o2-qc -b --config json://Users/bvonhall/dev/alice/sw/SOURCES/QualityControl/master/0/Modules/Example/etc/analysisDirect.json --dump-workflow --dump-workflow-file o2-qc-example-analysis-direct.json
```
Now trying on CC7

Barthelemy:
If I remove `--dump-workflow-file o2-qc-example-analysis-direct.json` it works.

knopers8:
thanks for checking. We definitely need that argument `--dump-workflow-file o2-qc-example-analysis-direct.json`. I will try to investigate on mac.

knopers8:
This will hopefully work on CI, but I don't know yet how to generate a DPL dump, where --config argument of o2-qc will always point to config files installed in ${QUALITYCONTROL_ROOT}/etc. If hyperloop uses RPMs, then the config file paths might be different than during the build.

Modules/Example/etc/analysisDerived.json
```diff
@@ -0,0 +1,55 @@
...
+        "url": "ccdb-test.cern.ch:8080"
+      }
+    },
+    "tasks": {
+      "AnalysisQcTask": {
+        "active": "true",
+        "className": "o2::quality_control_modules::example::AnalysisTask",
+        "moduleName": "QcExample",
+        "detectorName": "TST",
+        "cycleDurationSeconds": "10",
```
Barthelemy, line 31: Naive question: does it make sense for something running asynchronously offline to have a cycle in seconds rather than number of events ?

cmake/AddQCWorkflow.cmake
```diff
@@ -0,0 +1,35 @@
...
+# under share/dpl directory
+#
+
+function(o2_add_qc_workflow workflowName configurationFile)
+
+  set(jsonDumpFile ${workflowName}.json)
+  set(qcExecutable o2-qc)
+
+  install(CODE "message(\"Generating a workflow dump with the command: ${qcExecutable} -b --config json:/${CMAKE_INSTALL_PREFIX}/${configurationFile} --dump-workflow --dump-workflow-file ${CMAKE_CURRENT_BINARY_DIR}/${jsonDumpFile}\")")
+  install(CODE "execute_process(
```
Barthelemy, line 28: I did not know one could specify some code for the installation step. Interesting. The only caveat with this is that it is not explicitely dependent on the qcExecutable. I don't think it hurts here, as a change in the executable would probably not change the workflow.json produced here. 

You could consider using `add_custom_command` (https://cmake.org/cmake/help/latest/command/add_custom_command.html) to output a file and make it depend on the executable. This being said, if it works like that I am not sure it is worth the trouble.

Barthelemy (approved):
Very good, see my 2 comments but you can go ahead and merge.

cmake/AddQCWorkflow.cmake
knopers8: I tried it, but it complained that the target is in another library (or directory, i don't remember), that would be probably the cleanest way otherwise.
Anyway, this still doesn't work, it links to wrong libraries, so we still get segfaults.

knopers8: Ah, this specific version cannot even find the executable, it does if I add a `bin/` prefix, but then it links to wrong libraries

Modules/Example/etc/analysisDerived.json
knopers8: in principle yes, but i would wait for any particular requirements about this.

Barthelemy:
Ok, then I will work on the cmake asap.

Barthelemy:
Here is how I would do it in CMake : 
```
function(o2_add_qc_workflow)

  cmake_parse_arguments(
    PARSED_ARGS # prefix of output variables
    "" # list of names of the boolean arguments (only defined ones will be true)
    "WORKFLOW_NAME;CONFIG_FILE_PATH" # list of names of mono-valued arguments
    "" # list of names of multi-valued arguments (output variables are lists)
    ${ARGN} # arguments of the function to parse, here we take the all original ones
  )
  # note: if it remains unparsed arguments, here, they can be found in variable PARSED_ARGS_UNPARSED_ARGUMENTS
  if(NOT PARSED_ARGS_WORKFLOW_NAME OR NOT PARSED_ARGS_CONFIG_FILE_PATH)
    message(FATAL_ERROR "You must provide a workflow name and a config file path")
  endif()
  message(STATUS "o2_add_qc_workflow called with config file ${PARSED_ARGS_CONFIG_FILE_PATH} to generate ${PARSED_ARGS_WORKFLOW_NAME}")

  set(jsonDumpFile ${CMAKE_CURRENT_BINARY_DIR}/${PARSED_ARGS_WORKFLOW_NAME}.json)
  message(STATUS "jsonDumpFile: ${jsonDumpFile}" )
  set(qcExecutable o2-qc)

  add_custom_command(
    OUTPUT ${jsonDumpFile}
    VERBATIM
    COMMAND ${qcExecutable} -b --config json:/${CMAKE_INSTALL_PREFIX}/${PARSED_ARGS_CONFIG_FILE_PATH} --dump-workflow --dump-workflow-file ${jsonDumpFile}
    COMMAND echo "${qcExecutable} -b --config json:/${CMAKE_INSTALL_PREFIX}/${PARSED_ARGS_CONFIG_FILE_PATH} --dump-workflow --dump-workflow-file ${jsonDumpFile}"
    DEPENDS ${qcExecutable} ${configurationFile})

  get_filename_component(filename ${jsonDumpFile} NAME_WE)
  add_custom_target(${filename} ALL DEPENDS ${jsonDumpFile})

  # will install the rootmap and pcm files alongside the target's lib
  install(FILES ${jsonDumpFile}
          DESTINATION ${CMAKE_INSTALL_DATADIR}/dpl)

endfunction()
```
But your approach would totally work so I am not sure you should touch it. 

The culprit is somewhere else. Take the `master` and just do `o2-qc-run-basic  --dump-workflow --dump-workflow-file test.json` -> crash: 
```
[INFO] Dumping used configuration in dpl-config.json

 *** Break *** segmentation violation
[/usr/lib/system/libsystem_platform.dylib] _sigtramp (no debug info)
[<unknown binary>] (no debug info)
[/Users/bvonhall/dev/alice/sw/osx_x86-64/O2/0_O2_DATAFLOW-1/lib/libO2Framework.dylib] runStateMachine(std::__1::vector<o2::framework::DataProcessorSpec, std::__1::allocator<o2::framework::DataProcessorSpec> > const&, WorkflowInfo const&, std::__1::vector<o2::framework::DataProcessorInfo, std::__1::allocator<o2::framework::DataProcessorInfo> > const&, o2::framework::DriverControl&, o2::framework::DriverInfo&, std::__1::vector<o2::framework::DeviceMetricsInfo, std::__1::allocator<o2::framework::DeviceMetricsInfo> >&, boost::program_options::variables_map&, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >) /Users/bvonhall/dev/alice/sw/SOURCES/O2/0_O2_DATAFLOW/0/Framework/Core/src/runDataProcessing.cxx:0
[/Users/bvonhall/dev/alice/sw/osx_x86-64/O2/0_O2_DATAFLOW-1/lib/libO2Framework.dylib] doMain(int, char**, std::__1::vector<o2::framework::DataProcessorSpec, std::__1::allocator<o2::framework::DataProcessorSpec> > const&, std::__1::vector<o2::framework::ChannelConfigurationPolicy, std::__1::allocator<o2::framework::ChannelConfigurationPolicy> > const&, std::__1::vector<o2::framework::CompletionPolicy, std::__1::allocator<o2::framework::CompletionPolicy> > const&, std::__1::vector<o2::framework::DispatchPolicy, std::__1::allocator<o2::framework::DispatchPolicy> > const&, std::__1::vector<o2::framework::ConfigParamSpec, std::__1::allocator<o2::framework::ConfigParamSpec> > const&, o2::framework::ConfigContext&) /Users/bvonhall/dev/alice/sw/SOURCES/O2/0_O2_DATAFLOW/0/Framework/Core/src/runDataProcessing.cxx:0
[/Users/bvonhall/dev/alice/sw/osx_x86-64/QualityControl/knopers8-qc-analysis-1/bin/o2-qc-run-basic] main /Users/bvonhall/dev/alice/sw/osx_x86-64/O2/0_O2_DATAFLOW-1/include/Framework/runDataProcessing.h:0
[/usr/lib/system/libdyld.dylib] start (no debug info)
```
THus I think that the problem is deeper in DPL, right ?

Barthelemy:
It is creating and installing the workflow dumps but I am not sure it is yet ok. 
I see that the path to the config files (e.g. analysisDerived.json) in the workflow dump is correct. The path to `o2-qc` or to the dump itself is not the one installed though. Is that ok ? 
I let you try it.

knopers8:
Thank you. The exec path is probably correct. Dumps generated by Giulio and used in hyperloop also have paths like:
```
        {
            "name": "internal-dpl-clock",
            "executable": "../../../stage/bin/o2-analysis-filter-cf",
            "cmdLineArgs": [
```
so this may work.
The dump is not installed in the sources instead of the installation folder, but this should be easy to fix, will try.

knopers8:
This should OK, let's see if CI complains (when it starts to work).

knopers8:
My bad, your version was actually correct.

PR #538, opened 2020-11-04
[QC-472] Extract advanced topology

Barthelemy:
Goal: be able to plug the QC with a different config file. 
I would like to have that to test the aggregators in particular. 

Also add documentation.


Barthelemy:
@TimoWilken I am not sure what is going on but the tests are red while there are no details. Would you mind having a look when you have time ? thank you

TimoWilken:
Hi @Barthelemy, I believe the build fails because there are conflicts between this PR and the master branch -- "Cannot merge PR into test area" is probably not the best message to show in this case, I'll try to think of a better one.

GitHub shows conflicts in `Framework/CMakeLists.txt`. Could you resolve that, e.g. by `git merge`-ing the latest master into your PR branch and fixing the conflict during that merge?

Barthelemy:
Ok, totally missed the message, sorry about that :) I will merge manually

Barthelemy:
@TimoWilken Hello, sorry to call again, it seems to me that o2-dataflow is not progressing.

Framework/include/QualityControl/AdvancedWorkflow.h
```diff
@@ -0,0 +1,35 @@
...
+///
+
+#ifndef QUALITYCONTROL_AdvancedWorkflow_H
+#define QUALITYCONTROL_AdvancedWorkflow_H
+
+#include <Framework/CompletionPolicyHelpers.h>
+#include <Framework/ConfigContext.h>
+#include <Framework/DataSpecUtils.h>
+#include <Framework/CompletionPolicyHelpers.h>
+#include <Framework/WorkflowSpec.h>
```
knopers8, line 24: You probably don't need most of these in the header.

Framework/src/AdvancedWorkflow.cxx
```diff
@@ -0,0 +1,96 @@
...
+#include <Framework/CompletionPolicyHelpers.h>
+#include <Framework/DataSpecUtils.h>
+#include "QualityControl/QcInfoLogger.h"
+#include <random>
+
+using namespace o2;
+using namespace o2::header;
+using namespace o2::framework;
+using SubSpecificationType = o2::header::DataHeader::SubSpecificationType;
+using SubSpec = o2::header::DataHeader::SubSpecificationType;
```
knopers8, line 28: These two are the same, please pick one.

TimoWilken:
@Barthelemy I'll have a look!

doc/ModulesDevelopment.md
```diff
@@ -410,6 +410,21 @@ __Live detector data__
 
 If the detector is ready and connected to the CRU(s), one can of course start the full data taking workflow, including the SubTimeFrameBuilder and the DPL processing and plug the QC onto it.
 
+## A more advanced example
+
+A more complete example is available. The config file is called [advanced.json](../Framework/advanced.json). The workflow is made of 3 sources, intermediate processing steps, 3 sinks and a Dispatcher connecting two QC tasks to a number of these steps and 2 checks:
```
knopers8, line 415: Please consider adding a sentence saying that the topology doesn't mean to represent any particular physics processing, it is just an example with multiple data processors.

knopers8 (commented):
All good, I have just a few minor suggestions.

Barthelemy:
Thank you , I have addressed all comments.

knopers8 (approved):

Barthelemy:
The error on macOS is weird: `[ERROR] SEVERE: Device QC-TASK-RUNNER-skeletonTask (7203) had at least one message above severity ERROR: Unhandled std::exception reached the top of main: failed setting ZMQ_EVENTS, reason: Interrupted system call, device shutting down.`

I will trigger a rebuild

Barthelemy:
@knopers8 any idea why `testWorkflow` is failing ? I cannot reproduce the problem on my mac and I don't see how my change can affect this test.

knopers8:
No idea, but I've noticed that @ktf has been reworking integration with ZeroMQ in DPL, so it might be related.

ktf:
Yes, it's definitely related to some of the reworking I was doing. Where do you get it?

Barthelemy:
we get it in testWorkflow on Mac but only in this PR and I am not able to reproduce it on my Mac.
It happens when we quit: `        pctx.services().get<ControlService>().readyToQuit(QuitRequest::All);`

Barthelemy:
seems to be fixed

ktf:
@MichaelLettrich could you check what is going on with the macOS builder?

pzhristov:
I another pull request I saw that the brew updated python to 3.9 and then the Python-modules-list was not OK, for example there was no matching version of tensorflow. A quick solution is to link python@3.8. If we want to use python@3.9, I have a patch that needs some refinement.

MichaelLettrich:
@pzhristov @ktf indeed the macOS machines have been updated to python 3.9 via brew yesterday.  Do we need to downgrade them to 3.8?

pzhristov:
@MichaelLettrich No need to downgrade, simply link python@3.8 
`brew link --force --overwrite python@3.8`

PR #539, opened 2020-11-05
[EMCAL-524] apply on-the-fly calibration for the Bad Channel Map and Time Calibration

cterrevo:


Modules/EMCAL/src/DigitsQcTask.cxx
```diff
@@ -84,13 +103,23 @@ void DigitsQcTask::startOfActivity(Activity& /*activity*/)
 void DigitsQcTask::startOfCycle()
 {
   QcInfoLogger::GetInstance() << "startOfCycle" << AliceO2::InfoLogger::InfoLogger::endm;
+  std::map<std::string, std::string> metadata;
+  mBadChannelMap = retrieveConditionAny<o2::emcal::BadChannelMap>("EMC/BadChannelMap", metadata);
+  if (!mBadChannelMap)
```
Barthelemy, line 108: we recommend the use of curly braces even when there is a single line.

Barthelemy (approved):
Good for me, just consider adding curly braces in single line "if".

PR #541, opened 2020-11-06
[QC-459] Add InputRecord.h include to all tasks 

Barthelemy:
Needed by O2-1794, ie. forward declaring InputRecord to avoid problems with generating dictionaries.

### DIFF
--- a/Modules/EMCAL/src/DigitsQcTask.cxx
+++ b/Modules/EMCAL/src/DigitsQcTask.cxx
@@ -14,7 +14,7 @@
 #include "EMCAL/DigitsQcTask.h"
 #include "DataFormatsEMCAL/Cell.h"
 #include "EMCALBase/Geometry.h"
-#include "Framework/InputRecord.h"
+#include <Framework/InputRecord.h>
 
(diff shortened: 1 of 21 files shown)


Barthelemy:
@ktf With these changes, the change you propose in O2 (forward declaring InputRecord) should work.

Modules/EMCAL/src/DigitsQcTask.cxx
```diff
@@ -14,6 +14,7 @@
 #include "EMCAL/DigitsQcTask.h"
 #include "DataFormatsEMCAL/Cell.h"
 #include "EMCALBase/Geometry.h"
+#include "Framework/InputRecord.h"
```
knopers8, line 17:
```suggestion
#include <Framework/InputRecord.h>
```
Perhaps <> braces to let the compiler find the header faster, as it is should look for it in other project.

Barthelemy: absolutely, fixing all of htem

knopers8 (approved):

PR #542, opened 2020-11-09
ITS Fhr task optimization

JianLIUhep:
- ITS hits are stored as std::unordered_map<unsigned int, int>
- passing custom parameters via .json 
- completed OB functions
- THnSparse objects created on staves instead of HIC/chip

### DIFF
--- a/Modules/ITS/include/ITS/ITSFhrTask.h
+++ b/Modules/ITS/include/ITS/ITSFhrTask.h
@@ -32,7 +32,6 @@ class TH1F;
 class TH2;
 
 using namespace o2::quality_control::core;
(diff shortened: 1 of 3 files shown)


Barthelemy (approved):
I cannot judge on the logic but the code looks ok. 
We should perhaps consider how to let users' code send metrics to the Monitoring system. It seems to me that you are trying to monitor the performance of your task.

JianLIUhep:
Yes, it would be helpful if we can look into the performance of the sub-detector tasks through the Monitoring.

Barthelemy:
https://alice.its.cern.ch/jira/browse/QC-474

PR #543, opened 2020-11-10
[QC-476] Add option to save locally objects in a ROOT file

Barthelemy:


Framework/src/TaskRunner.cxx
```diff
@@ -485,4 +489,14 @@ int TaskRunner::publish(DataAllocator& outputs)
   return objectsPublished;
 }
 
+void TaskRunner::saveToFile()
+{
+  if(!mTaskConfig.saveToFile.empty()) {
+    ILOG(Debug, Support) << "Save data to file " << mTaskConfig.saveToFile << ENDM;
+    TFile f(mTaskConfig.saveToFile.c_str(), "RECREATE");
+    mObjectsManager->getNonOwningArray()->Write();
+    f.Close();
```
knopers8: I am not sure, but maybe it would be better to save the raw objects instead of the array of MonitorObjects? I have two reasons in mind:
1. It should be easier to browse and share with others, no need to load QualityControlTypes.
2. They are saved like that in QCDB.

Barthelemy: indeed, I have taken the lazy path here. I checked that it worked with the TBrowser when starting root from the QC environment.

Ok, I will do it.

Barthelemy:
ok, I save the encapsulated TObjects inside the MonitorObjects

Framework/src/TaskRunner.cxx
```diff
@@ -485,4 +489,16 @@ int TaskRunner::publish(DataAllocator& outputs)
...
 
+void TaskRunner::saveToFile()
+{
+  if (!mTaskConfig.saveToFile.empty()) {
+    ILOG(Debug, Support) << "Save data to file " << mTaskConfig.saveToFile << ENDM;
+    TFile f(mTaskConfig.saveToFile.c_str(), "RECREATE");
+    for (size_t i = 0; i < mObjectsManager->getNumberPublishedObjects(); i++) {
+      mObjectsManager->getMonitorObject(i)->getObject()->Write();
+    }
+    f.Close();
```
knopers8, line 500: But will that work if there is more than one MO? won't they overwrite each other?

Barthelemy: No, because ROOT saves the objects under a name (Key) which corresponds to the TObject name.

knopers8: thanks

knopers8 (approved):

PR #545, opened 2020-11-10
TPC: Add task for receiving TPCTrackingQA from external

davidrohr:
This adds the TPC JSON configuration to receive TPC Tracking QA data as external input.
The second variant to process the QA in a real QC task will come in a second PR.

### DIFF
--- a/Modules/TPC/CMakeLists.txt
+++ b/Modules/TPC/CMakeLists.txt
@@ -82,4 +82,5 @@ install(FILES run/tpcQCPID_sampled.json
               run/tpcQCSimpleTrending.json
               run/tpcQCTrendingClusters.json
               run/tpcQCClusters_direct.json
+              run/tpcQCTrackingFromExternal_direct.json
         DESTINATION etc)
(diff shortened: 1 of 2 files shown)


davidrohr:
@Barthelemy : Following up on the e-mail thread: I have truncated and filled qcg-test.cern.ch with this PR and AliceO2Group/AliceO2#4810, looks good.
Perhaps we can just address 1 point before merging:
- I have to use `External-1` in the query `"query": "External-1:TPC/TRACKINGQA/0"`, otherwise it does not work. Any idea why?
I think anything else is OK for now and can be improved later.

Barthelemy:
@davidrohr That must be a bug. I can have a look this afternoon.

davidrohr:
@Barthelemy : In fact what I wrote is wrong, I can change the name, just had to do it in 2 places, which I didn't see before :)
After I have renamed the external task it works. Perhaps one could still improve the error message :)
From my side this is ready to be merged.

Barthelemy:
Ok, good to know, I was indeed not able to reproduce it. I am approving, and later merging. I will see what I can do for the error message.

Barthelemy (approved):

Barthelemy:
@TimoWilken It seems that build/QualityControl/o2 has been building for almost a day ? would you mind checking it ?

TimoWilken:
@Barthelemy I'll have a look!

PR #550, opened 2020-11-16
First HMPID QC implementation

gvolpe79:


Modules/HMPID/src/HmpidTask.cxx
```diff
@@ -0,0 +1,247 @@
...
+void HmpidTask::startOfCycle()
+{
+  ILOG(Info) << "startOfCycle" << ENDM;
+}
+
+void HmpidTask::monitorData(o2::framework::ProcessingContext& ctx)
+{
+  nCycles++;
+
+  Printf("******************IN HMPID MONITOR DATA*********************************************");
```
Barthelemy, line 103: Same remark about logging. The output will be lost if you print it instead of sending it to either the infologger or fairlogger.

Barthelemy:
all good for me, could you just reformat ? or do you need for that ?

Barthelemy (approved):

Barthelemy:
Thank you, it is approved and the tests will start now.

PR #553, opened 2020-11-17
Add TPC Tracking QA Task

davidrohr:
@wiechula @stheckel @tklemenz : This adds the standalone TPC QA task to QC (needs AliceO2Group/AliceO2#4846)
- I can get the reconstructed tracks via `o2-qc-run-tpctrackreader`
- And I am reading the tpc track references from the ROOT file.
In addition, I would need:
- The track MC labels (can I get them also from `o2-qc-run-tpctrackreader` somehow).
- The cluster MC labels (I need to cut on how many clusters are found in the TPC to define some findable criterion, and I think for this I need the MC labels). The O2 workflows usually just initialize a RootTreeReader to read tpc clusters + mc labels internally. Is there already a way I can easily provide them for a QC task?

### DIFF
--- a/Modules/TPC/CMakeLists.txt
+++ b/Modules/TPC/CMakeLists.txt
@@ -3,6 +3,7 @@
 add_library(QcTPC)
 
 target_sources(QcTPC PRIVATE src/PID.cxx
+                             src/Tracking.cxx
                              src/Tracks.cxx
                              src/PIDClusterCheck.cxx
                              src/TrackClusterCheck.cxx
@@ -24,6 +25,7 @@ target_link_libraries(QcTPC
 
 add_root_dictionary(QcTPC
                     HEADERS include/TPC/PID.h
+                            include/TPC/Tracking.h
                             include/TPC/Tracks.h
                             include/TPC/PIDClusterCheck.h
                             include/TPC/TrackClusterCheck.h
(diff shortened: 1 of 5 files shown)


Barthelemy (approved):
Most of the code is in O2, what I see here is ok.

Barthelemy:
I wait for @wiechula @stheckel @tklemenz  to comment and give approval.

davidrohr:
The code here is OK, but it canot be used yet. We need:
- Add the components that read track and cluster MC labels.
- Change the nullptrs her to point to the correct MC data.
- Add a json config file to run this task and read data from the new input components.

Barthelemy:
OK, I propose to "WIP" it then.

davidrohr:
OK, just wanted to see whether it passes the CI or if I still have to do something, therefore no WIP :)

tklemenz:
Hi @davidrohr, unfortunately I think I cannot help you much, since we haven't had a look at MC labels, yet. 

Concerning the trackreader: If this is the way you want to go instead of taking the data directly from the reco-workflow I think it should be possible to add another output (TPCTracksMCTruth) to the reader. How exactly this would look like I do not know but feel free to modify the reader. 

Concerning the cluster MC labels: Afaik there is nothing implemented in the QC so far to easily provide MC labels for the clusters. At least not from our side or anyone else that I know of.

davidrohr:
@tklemenz : Yes, I think we should modify the o2-qc-run-tpctrackreader to provide the mc track labels as a second output.
For the clusters, I guess we should implement an o2-qc-run-tpcclusterreader, which can read and ship the clusters and the cluster mc labels.
In principle, we could even do this inside O2 not inside QC because I think it might be helpful in general to have a device that can provide clusters / tracks as input.
We can discuss this during our call next Wednesday.

davidrohr:
@Barthelemy : This is now fully working and ready to be merged.

PR #555, opened 2020-11-18
Barth comments implemented

gvolpe79:


Barthelemy:
Hi Giacomo, 
Do I understand correctly that these are the changes I requested for the other PR ? #550 
If so, you should commit them to the same branch on top of the rest. 
Can you try ? 
Thank you

gvolpe79:
Hi Barth,

  yes, they are! I tried to do that, but I did not succeed, I do not know why! So then I decided to create a new branch !
I will try to understand what is the problem for the next time! Is really needed to use the same branch?

Th problem is the following:

I follow the procedure here https://github.com/AliceO2Group/QualityControl/blob/master/doc/ModulesDevelopment.md
I do

git clone https://github.com/gvolpe79/QualityControl.git
git checkout -b hmpid-qc
git push --set-upstream origin hmpid-qc (here the problem, it seems I have to pull before but..)

git pull (doesn't work)
git pull hmpid-qc (doesn't work)

It seems that the branch is not recognized....do you have any suggestion. What shall I do to pull the branch?!

Thanks, cheers.

Giacomo.

------------------------
Giacomo Volpe, Ph.D.
Assistant Professor - Dipartimento Interateneo di Fisica ‘M. Merlin’ and Sezione INFN
Via Giovanni Amendola, 173
Bari, ITALY
contacts:
Giacomo.Volpe@cern.ch, Giacomo.Volpe@uniba.it, Giacomo.Volpe@ba.infn.it
tel. +39 5442385
------------------------
________________________________

Barthelemy:
Hi, 

Git can be confusing. 

I will update the doc once we understand what is going on.

The way I would expect it to go is 

1. git clone (...)
2. git checkout -b hmpid-qc
3. git push (...) to push the branch to the server
4. do your changes, keep adding and committing, also pushing
5. create a PR
6. [PR under review -> changes are requested]
7. Go back to your code, which should still be there !
8. More commits, push them
9. PR is automatically updated. 

So basically  you should just add your new commits corresponding to what I requested on top of the other commits in the same branch. 

Have you removed the local directory ? what does it say when you check out your branch ?

PR #560, opened 2020-11-26
[QC-489] Make sure aggregators are run in the right order

Barthelemy:


Framework/src/AggregatorRunner.cxx
```diff
@@ -200,17 +201,93 @@ void AggregatorRunner::initAggregators()
...
+  // One could build a proper tree (e.g. with boost Graph) and then apply a complex algorithm to order
+  // the nodes and find cycles.
+  // Instead this implementation goes through the aggregators and for each checks whether
+  // there are no dependencies or if they are all fulfilled. If it is the case the aggregator
+  // is moved at the end of the resulting vector.
+  // In case we have looped through all the remaining aggregators and nothing has been done,
+  // ie. no aggregator has its dependencies fulfilled, we stop and raise an error.
+  // This means that there is a cycle or that one of the dependencies does not exist.
+  // Note that by "fulfilled" we mean that all the sources of an aggregator are already
+  // in the result vector.
```
knopers8, line 256: I am not sure if that will always work, looking at the code.
Imagine the following scenario:
Two independent check+aggregators chains:
C1 -> A1 -> A2 -> A3.
C2 -> A4 -> A5
The aggregators are put in the following order:
A5,A4,A3,A2,A1
Then `originals` in each iteration will look as follows:
A5,A4,A3,A2,A1
A5,A3,A2
A3
`modificationLastIteration` will be set to true after pushing A5 to `toBeMoved`, thus the loop will still move A2, but then it will exit and A3 won't be handled.
Though I am not 100% sure I am correct.

knopers8 (commented):
Thank you, I like the idea, though I am not sure it will work for each case, please see my comment to the algorithm

Framework/src/AggregatorRunner.cxx
knopers8: C1 -> A1 -> A2
C2 -> A3
and
A3,A2,A1
actually represent the same idea but simpler.

Barthelemy: In your second example, the `originals` are as follow

- A1, A2, A3 -> originals is not empty: loop 
  - 2 items have no deps and are removed: A1 and A3
- A2 -> originals is not empty: loop
  - A2 has its deps fulfilled, it is removed
- <none> -> empty and true: loop
- <none> -

> empty and false: stop

It is suboptimal because we always loop one too many times. 
I might totally be missing the point :)

Barthelemy: I will try it to see

knopers8: A1 and A3 will be indeed removed, but modificationLastIteration will be set to true by chance, so the while loop will break, won't it?

knopers8: ah, sorry, my mistake. for some reason i thought that there was **!** before `modificationLastIteration` in the while loop check. I need a weekend

Barthelemy: it works in both cases

knopers8: thank you for cross-checking

Barthelemy:
Merging because failures are unrelated.

PR #565, opened 2020-12-01
fix macos warnings in TOF

Barthelemy:
@njacazio @ktf


Barthelemy:
merging because it blocks the rest

PR #567, opened 2020-12-03
fix links in the QC analysis doc

knopers8:


knopers8:
simple fix in the documentation, no need for CI, merging

PR #571, opened 2020-12-07
Enable code format check via GitHub Actions

Barthelemy:


Barthelemy:
tested with a badly formatted file and it worked.

.github/workflows/format.yml
```diff
@@ -0,0 +1,26 @@
...
+      - uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+      - name: Run clang-foramt on changed files
+        run: |
+          set -x
+          git fetch origin ${{ github.event.pull_request.base.ref }}
+          git fetch origin pull/${{ github.event.pull_request.number }}/head:${{ github.event.pull_request.head.ref }}
+          BASE_COMMIT=$(git rev-parse ${{ github.event.pull_request.base.sha }})
+          COMMIT_FILES=$(git diff --name-only ${BASE_COMMIT} | grep -i -v LinkDef)
```
knopers8, line 18: What about CMakeLists, bash scripts and python files. With clang-format behave?

.github/workflows/format.yml
```diff
@@ -0,0 +1,26 @@
...
+          RESULT_OUTPUT=$(git-clang-format-8 --commit ${BASE_COMMIT} --diff --binary $(which clang-format-8) ${COMMIT_FILES})
+          if [ "$RESULT_OUTPUT" == "no modified files to format" ] || [ "$RESULT_OUTPUT" == "clang-format did not modify any files" ]; then
```
knopers8, line 20: What if the text changes with a future version? Maybe it would be better to check the exit code? (if it is actually non-zero when there are no files to format)

knopers8 (commented):
I am not very proficient in this. It looks good, I just have two small doubts, see the comments.

.github/workflows/format.yml
Barthelemy: correct, good point, I am adding a restriction to *.cxx and *.h only, excluding LinkDef

.github/workflows/format.yml
Barthelemy: Unfortunately the tool returns the same value in both cases. :(

Barthelemy:
Merging because the mac build is always broken.

PR #572, opened 2020-12-09
 ITS: diagnostic info inspection and adding QC check for ITSFeeTask

JianLIUhep:
- ITSFeeTask verified with the ITS outer barrel at 100kHz trigger rate (datasampling fraction is 100%). readout -> stfb -> o2-dpl-raw-proxy -> o2-qc
- Add ITSFeeCheck
- Bug fixes for cluster and track checks

### DIFF
--- a/Modules/ITS/include/ITS/ITSFeeTask.h
+++ b/Modules/ITS/include/ITS/ITSFeeTask.h
@@ -33,18 +33,23 @@ namespace o2::quality_control_modules::its
 /// \brief ITS FEE task aiming at 100% online data integrity checking
 class ITSFeeTask final : public TaskInterface
 {
-
   struct GBTDdw { //GBT diagnostic word
-
-    union GBTBits {
-      struct payloadBits {
-        uint64_t flags2 : 36; /// 0:35  not defined yet
-        uint64_t flags1 : 36; /// 36:71  not defined yet
-        uint64_t id : 8;      /// 72:79  0xe0; Header Status Word (HSW) identifier
-      } payload;
-
-      uint8_t data8[16]; // 80 bits GBT word +  padding to 128 bits
-    } ddwBits;
+    union {
+      uint64_t word0 = 0x0;
+      struct {
+        uint64_t laneStatus : 56;
+        uint16_t zero0 : 8;
+      } laneBits;
+    };
+    union {
+      uint64_t word1 = 0x0;
+      struct {
+        uint8_t flag1 : 4;
+        uint8_t index : 4;
+        uint8_t id : 8;
+        uint64_t padding : 48;
+      } indexBits;
+    };
   };
 
  public:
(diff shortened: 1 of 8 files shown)


Modules/ITS/include/ITS/ITSFeeTask.h
```diff
@@ -33,18 +33,23 @@ namespace o2::quality_control_modules::its
 /// \brief ITS FEE task aiming at 100% online data integrity checking
 class ITSFeeTask final : public TaskInterface
 {
-
   struct GBTDdw { //GBT diagnostic word
```
Barthelemy, line 36: Coding Guidelines recommendation: avoid abbreviations. https://rawgit.com/AliceO2Group/CodingGuidelines/master/naming_formatting.html#General_Naming_Rules

Modules/ITS/include/ITS/ITSFeeTask.h
```diff
@@ -33,18 +33,23 @@ namespace o2::quality_control_modules::its
...
-    union GBTBits {
-      struct payloadBits {
-        uint64_t flags2 : 36; /// 0:35  not defined yet
-        uint64_t flags1 : 36; /// 36:71  not defined yet
-        uint64_t id : 8;      /// 72:79  0xe0; Header Status Word (HSW) identifier
-      } payload;
-
-      uint8_t data8[16]; // 80 bits GBT word +  padding to 128 bits
-    } ddwBits;
+    union {
```
Barthelemy, line 37: Out of curiosity, is this not available in O2 or another package ?

Modules/ITS/src/ITSFeeTask.cxx
```diff
@@ -149,22 +218,14 @@ void ITSFeeTask::monitorData(o2::framework::ProcessingContext& ctx)
...
+void ITSFeeTask::getRunNumber()
 {
-  std::ifstream configFile("Config/ConfigFakeRateFee.dat");
-  for (int ilayer = 0; ilayer < NLayer; ilayer++) {
-    configFile >> mEnableLayers[ilayer];
-    if (mEnableLayers[ilayer]) {
-      ILOG(Info) << "Enable layer : " << ilayer << ENDM;
-    }
-  }
+  std::ifstream configFile("Config/ConfigFee.dat");
```
Barthelemy, line 226: Could you make it slightly more resistant to the very probable case when the file is not there ?

Barthelemy (changes requested):
Very good. I have 2 questions and 1 request for change. Thank you for considering them.

Modules/ITS/include/ITS/ITSFeeTask.h
JianLIUhep: done. hope the amended variable name meets the requirement.

Modules/ITS/src/ITSFeeTask.cxx
JianLIUhep: Added the checking for the config file and run number file during the commissioning.

Modules/ITS/include/ITS/ITSFeeTask.h
JianLIUhep: No, nothing implemented in O2 yet concerning the diagnostic GBT words. But we should think about moving this light-weight decoding algorithm to the O2, e.g.  AliceO2/Detectors/ITSMFT/ITS/QC.  Adding them directly to the current ITS raw data decoder might not be a good idea in terms of CPU/RAM consumption. To be discussed.

Barthelemy (approved):

Barthelemy:
Merging as the Mac build is broken.

PR #573, opened 2020-12-11
Make sure Travis is no longer used

awegrzyn:


Barthelemy:
Merging because mac builds are always broken.

PR #576, opened 2020-12-16
[QC-509] Clear warnings

Barthelemy:


Barthelemy:
@JianLIUhep

JianLIUhep:
Thank you @Barthelemy. It seems the same fix should also be carried out for ITSTrackCheck.cxx and ITSFeeCheck.cxx.

JianLIUhep:
And the **++iter** for ITSFhrCheck.cxx

Barthelemy:
@JianLIUhep Yes, correct, I am going to do that.

PR #581, opened 2021-01-06
o2-data-dump: use AliceO2::DebugGUI

ktf:
This replaces the internal version with the common one, so that we
can merge https://github.com/alisw/alidist/pull/2728 with the Metal
support.


ktf:
@Barthelemy could you it give a try? I got it to compile just fine, but I've not had chance to try it out. It will also speed up the build a bit, I guess.

Barthelemy:
@ktf ok, compiling on CC7 . Thank you

Barthelemy (approved):
tested and working on mac

Barthelemy:
-42000 lines is always something :)

ktf:
cool thanks!

PR #584, opened 2021-01-11
[QC-515] Fix analysis json generation in Example and stop polluting src dir

Barthelemy:


Modules/Example/CMakeLists.txt
```diff
@@ -25,16 +25,10 @@ install(TARGETS QcExample
...
-# the config files being already installed in the right place.
-configure_file("etc/analysisDirect.json" "${CMAKE_INSTALL_PREFIX}/etc/analysisDirect.json")
-configure_file("etc/analysisDerived.json" "${CMAKE_INSTALL_PREFIX}/etc/analysisDerived.json")
-
 # ---- Workflows for analysis ----
 
-o2_add_qc_workflow(WORKFLOW_NAME o2-qc-example-analysis-direct CONFIG_FILE_PATH ${CMAKE_INSTALL_PREFIX}/etc/analysisDirect.json)
-o2_add_qc_workflow(WORKFLOW_NAME o2-qc-example-analysis-derived CONFIG_FILE_PATH ${CMAKE_INSTALL_PREFIX}/etc/analysisDerived.json)
+o2_add_qc_workflow(WORKFLOW_NAME o2-qc-example-analysis-direct CONFIG_FILE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/etc/analysisDirect.json)
+o2_add_qc_workflow(WORKFLOW_NAME o2-qc-example-analysis-derived CONFIG_FILE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/etc/analysisDerived.json)
```
knopers8, line 31: The DPL dumps should point to the configuration files in the installation paths, not the ones in source dir.

Barthelemy:
@TimoWilken @ktf It seems that the o2 build failed without producing a log.

Modules/Example/CMakeLists.txt
Barthelemy: ah right, that was the thing. It needs to point to the correct path to generate a proper json. Fixing

davidrohr (approved):
For reference, all problems are fixed for me now with this.

PR #585, opened 2021-01-11
[EMCAL-524,EMCAL-526] Plots for different trigger classes

cterrevo:
Create and fill QA plots for Calib and Physics Triggers in Raw and Digit
tasks

### DIFF
--- a/Modules/EMCAL/src/DigitsQcTask.cxx
+++ b/Modules/EMCAL/src/DigitsQcTask.cxx
@@ -17,6 +17,7 @@
 #include "EMCALCalib/BadChannelMap.h"
 #include "EMCALCalib/TimeCalibrationParams.h"
 #include <Framework/InputRecord.h>
+#include <CommonConstants/Triggers.h>
 
 namespace o2
 {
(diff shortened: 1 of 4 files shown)


Modules/EMCAL/src/DigitsQcTask.cxx
```diff
@@ -142,21 +179,34 @@ void DigitsQcTask::monitorData(o2::framework::ProcessingContext& ctx)
     //gsl::span<const o2::emcal::Digit> eventdigits(digitcontainer.data() + trg.getFirstEntry(), trg.getNumberOfObjects());
     gsl::span<const o2::emcal::Cell> eventdigits(digitcontainer.data() + trg.getFirstEntry(), trg.getNumberOfObjects());
 
+    //trigger type
+    auto triggertype = trg.getTriggerBits();
+    bool isPhysTrigger = triggertype & o2::trigger::PhT, isCalibTrigger = triggertype & o2::trigger::Cal;
+    std::string trgClass;
+    if (isPhysTrigger)
```
Barthelemy, line 186: we recommend to always add curvy braces.

Barthelemy (commented):
Ok, just 1 comment.

Barthelemy (approved):

Barthelemy:
@ktf @TimoWilken I think that there is something wrong with the mac builder.

TimoWilken:
@Barthelemy looking into it!

TimoWilken:
@Barthelemy I think I've found the issue - the build timed out, very probably because of a corrupt git repo for some dependencies. I've fixed it for now and I'll have a look at how to prevent this in future. In the meantime, I've marked the failed builds as pending so they run again.

Modules/EMCAL/src/DigitsQcTask.cxx
mfasDa: OK, we take this as recommendation :)

mfasDa:
To make clear: We need a change in O2, because we need to propagate the trigger bits in one of our headers in order to distinguish different triggers. The change is merged into dev there.

We tested the change with the data we just collected the past few days with the expected setup for the VS test and we are ready with this. It is important for us to get this commit merged before the MW1.

Barthelemy:
@mfasDa There are real errors in the o2 build that must be fixed, otherwise it would break everything. 
For the mac build, I am not sure why it fails if the change in O2 is there. Or is it just an old build ? 

I am not sure that it is possible to have it in time for MW1 to be honest. We need to have the CI pass here, create a new version of QC before the FLP suite is locked.

mfasDa:
Should be closed as the PR is split into a PR for the raw task https://github.com/AliceO2Group/QualityControl/pull/590 which does not rely on changes in O2 and should be handled with priority, and a commit for the cell task https://github.com/AliceO2Group/QualityControl/pull/589 which relies on changes in O2.

PR #586, opened 2021-01-13
Add noise analysis to QC

njacazio:
* Add basic histos for noise analysis

* Add Time Window as a parameter in json

* Add function for noise map

* Insert noise threshold parameter in json

* Update name of tasks in TOF config


Barthelemy:
@njacazio could you check why the macos build is breaking ?

njacazio:
@Barthelemy ok, it should be fixed now, I also added the multinode json files. Let's wait for the tests!

Barthelemy (approved):

PR #587, opened 2021-01-13
[QC-516] Do not crash if config file URI is not valid

Barthelemy:
And document a bit better.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,7 +13,7 @@ include(AddQCWorkflow)
 # ---- Project ----
 
 project(QualityControl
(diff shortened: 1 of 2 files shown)


knopers8 (approved):
It would be also nicer if Configuration returned a clear error instead of crashing.

Barthelemy:
It does not crash, it throws a generic runtime_error. On certain systems (Linux) this results in a segfault if not catched, for some reasons. I would be in favour of the Configuration raising a specific exception to help us catch it, probably at the level of the DS which could then decide how to proceed or bail out.

knopers8:
Oh, shame on me then. I didn't expect that it wouldn't be catched, DPL usually catches everything nicely.

PR #590, opened 2021-01-14
[EMCAL-524] Plots for different trigger classes

mfasDa:
- Create and fill QA plots for Calib and Physics Triggers in Raw task
- ADC range of raw amplitude projection increased in order to cover the
  range of LED events


mfasDa:
@Barthelemy This is the commit in the raw task which does not depend on changes in O2. It should be handled with high priority

Barthelemy:
@TimoWilken would it be possible to know where in the queue this PR is with regards to the macOS build ?

TimoWilken:
@Barthelemy it's building now. There was still a problem with some of the large git repos on the CI server, I've disabled `git gc` for now, so it doesn't keep repacking them.

mfasDa:
Build error unrelated

Barthelemy (approved):

PR #588, opened 2021-01-13
[MFT-Module] Updating the MFT digit QC Task

tomas-herman:
Adding overview histograms to the MFT digit QC task, adding pixel and chip hit maps, adding hot pixel data producer for testing.

PR #592, opened 2021-01-17
[MFT-5] Updating the Cluster QC task to work with the new itsmft Comp Clusters 

tomas-herman:
This PR already contains the changes in https://github.com/AliceO2Group/QualityControl/pull/588 which is stuck on unknown error at this time. So it also shows the changes from that PR as new.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,7 +13,7 @@ include(AddQCWorkflow)
 # ---- Project ----
 
 project(QualityControl
-        VERSION 1.7.0 #  TODO update this automatically when there are new releases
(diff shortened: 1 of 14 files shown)


tomas-herman:
@Barthelemy the o2 test is currently in an error state but for the past 4 days it was successful so I would say it is a spurious error.

Barthelemy (approved):
Looks ok, I don't review the logic of modules code, only the general adherence to the coding guidelines and obvious mistakes.

Barthelemy:
@tomas-herman Agreed, force merging

tomas-herman:
Understood, thank you.

PR #589, opened 2021-01-14
[EMCAL-526] Plots for different trigger classes

mfasDa:
Create and fill QA plots for Calib and Physics Triggers in Digit task

PR #593, opened 2021-01-19
Use new TPC cluster reader helper to avoid copy&pasted code

davidrohr:
no functionall change, just getting rid of some copy&pasted code, tested already and histograms are published to the qcg-test website in the same way as before.

### DIFF
--- a/Modules/TPC/src/Tracking.cxx
+++ b/Modules/TPC/src/Tracking.cxx
@@ -28,10 +28,8 @@
 #include "SimulationDataFormat/MCCompLabel.h"
 #include "SimulationDataFormat/MCTruthContainer.h"
 #include "SimulationDataFormat/ConstMCTruthContainer.h"
-#include "DataFormatsTPC/TPCSectorHeader.h"
-#include "DataFormatsTPC/ClusterGroupAttribute.h"
 #include "DataFormatsTPC/ClusterNative.h"
-#include "DataFormatsTPC/ClusterNativeHelper.h"
+#include "DataFormatsTPC/WorkflowHelper.h"
 #include "TPCQC/Helpers.h"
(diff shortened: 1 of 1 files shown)


davidrohr:
@Barthelemy @TimoWilken : What is QC/o2-dataflow tested against? Is it o2 dev branch? I am wondering how it can still fail, since the problem in O2 is fixed, and I could compile O2 and QC with my patch with o2-dataflow defaults locally.

Barthelemy:
I am not sure. Timo, could you check ? I suspect that it is against the alidist defined version of O2 which would be incorrect.

TimoWilken:
@davidrohr @Barthelemy it's tested against what's defined in the master branch of alidist, so right now it's v21.01.

davidrohr:
ok, so this is the reason why it is failing, since the tag lacks the latest fixes in O2/dev. Shouldn't it be checked against dev?

Barthelemy:
I agree, I would prefer if it was tested against dev.

TimoWilken:
See: https://github.com/alisw/ali-bot/pull/931

davidrohr:
@Barthelemy : Something is stange here. Both dataflow CIs have been green during the day, but now they are failing again somewhere else.

Barthelemy:
I don't even find where it is failing... Could it be the same as in https://github.com/AliceO2Group/QualityControl/pull/589#issuecomment-764744483 ? @TimoWilken @ktf any idea ?

davidrohr:
Now all CIs are failing without any indication what is going wrong @Barthelemy @TimoWilken

Barthelemy:
@ktf @TimoWilken Could you check ?

davidrohr:
@Barthelemy : It is finally green :)
Could you merge it?

PR #597, opened 2021-01-27
TOF: Drawing the diagnostics per crate

njacazio:
- Add RDH information

### DIFF
--- a/Modules/TOF/src/TaskRaw.cxx
+++ b/Modules/TOF/src/TaskRaw.cxx
@@ -45,6 +45,11 @@ void RawDataDecoder::decode()
   DecoderBase::run();
 }
 
+void RawDataDecoder::rdhHandler(const o2::header::RAWDataHeader* rdh)
+{
+  mRDHCounter[rdh->feeId & 0xFF].Count(0);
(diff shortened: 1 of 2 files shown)


Modules/TOF/src/TaskRaw.cxx
```diff
@@ -45,6 +45,11 @@ void RawDataDecoder::decode()
   DecoderBase::run();
 }
 
+void RawDataDecoder::rdhHandler(const o2::header::RAWDataHeader* rdh)
+{
+  mRDHCounter[rdh->feeId & 0xFF].Count(0);
```
njacazio, line 50: @preghenella I check the feeID like in the compressor. Then I fill the crate map with it in line https://github.com/AliceO2Group/QualityControl/pull/597/files#diff-85ba64596620595e54f7320c6e7707eaa1e851b0830a237a42f2aa56294e7696R434

Barthelemy (approved):

Barthelemy:
The failures in macos are false positives. For o2-dataflow I am checking what is wrong, for sure nothing related to your code.

PR #599, opened 2021-02-02
Disable failing tests until we understand what is going on

Barthelemy:


Barthelemy:
@TimoWilken @ktf the builds `o2` and `o2-dataflow` failed without logs. Force merging

PR #602, opened 2021-02-03
Script to create patch

Barthelemy:


knopers8 (approved):
Thank you, this is very nice.
In August it might break because the user+password authentication will be deprecated on github. Unless the ssh keys for git are globally registered on a machine, I don't remember.

Barthelemy:
you are correct about the change of authentication. I think that it should still be able to have the keys stored at the system level. We shall see and adapt the script.

Barthelemy:
Also, I would like to to do the same to release a new version but it is less urgent.

PR #603, opened 2021-02-04
[QC-532] Catch exceptions before they reach the DPL service manager.

Barthelemy:


Barthelemy:
If we let the exceptions bubble up, we lose them and it makes debugging far more difficult as we just see a crash.

knopers8 (approved):
Ideally, I think that DPL should handle them properly, but at this point it seems that it is better to rely on ourselves.

PR #605, opened 2021-02-04
[QC-518] Add a Merger even when there is only one local Task

knopers8:
Less efficient, but more predictable.


Barthelemy (approved):
It actually simplifies the code.

PR #610, opened 2021-02-09
[QC-539] Add missing level to infologger calls

Barthelemy:


Barthelemy:
@TimoWilken @ktf any idea why the checks are red again ?

Barthelemy:
@TimoWilken @ktf Could you check why the builds are borken ? it has been weeks

ktf:
@TimoWilken is it actually building? I do not see any progress since this morning.

TimoWilken:
@ktf it is now, I restarted it again this afternoon. There was a problem when I restarted this morning, which I only noticed later.

Barthelemy:

PR #611, opened 2021-02-09
Add first version of QC task for TPC calibration data + more output for TPC cluster QC task

tklemenz:
This adds:

- a first version of a QC task for TPC calibration data. For now it takes the most recent pedestal and noise CalPad object from the CCDB and publishes the data in 1D and 2D histograms.
- nicer output for the TPC cluster data.

The calibration data task can be run by simply calling  
`o2-qc --config json:/${QUALITYCONTROL_ROOT}/etc/tpcQCCalibration.json`

### DIFF
--- a/Modules/TPC/CMakeLists.txt
+++ b/Modules/TPC/CMakeLists.txt
@@ -8,7 +8,9 @@ target_sources(QcTPC PRIVATE src/PID.cxx
                              src/PIDClusterCheck.cxx
                              src/TrackClusterCheck.cxx
                              src/ROCReductor.cxx
-                             src/Clusters.cxx)
+                             src/Clusters.cxx
+                             src/CalDetPublisher.cxx
+                             src/Utility.cxx)
 
 target_include_directories(
   QcTPC
@@ -31,6 +33,8 @@ add_root_dictionary(QcTPC
                             include/TPC/TrackClusterCheck.h
                             include/TPC/ROCReductor.h
                             include/TPC/Clusters.h
+                            include/TPC/CalDetPublisher.h
+                            include/TPC/Utility.h
                     LINKDEF include/TPC/LinkDef.h
                     BASENAME QcTPC)
 
@@ -86,4 +90,5 @@ install(FILES run/tpcQCPID_sampled.json
               run/tpcQCTrendingClusters.json
               run/tpcQCClusters_direct.json
               run/tpcQCTrackingFromExternal_direct.json
+              run/tpcQCCalDetPublisher.json
         DESTINATION etc)
--- a/Modules/TPC/src/Clusters.cxx
+++ b/Modules/TPC/src/Clusters.cxx
@@ -13,40 +13,15 @@
 /// \author Thomas Klemenz
 ///
 
-#include "Framework/PartRef.h"
-#include "Framework/WorkflowSpec.h" // o2::framework::mergeInputs
-#include "Framework/DataRefUtils.h"
-#include "Framework/DataSpecUtils.h"
-#include "Framework/ControlService.h"
-#include "Framework/ConfigParamRegistry.h"
-#include "Framework/InputRecordWalker.h"
-#include "DataFormatsTPC/TPCSectorHeader.h"
-
-#include <bitset>
-
-// root includes
-#include <TCanvas.h>
-#include <TH1.h>
-#include <TH2.h>
-
 // O2 includes
 #include "Framework/ProcessingContext.h"
-#include "DataFormatsTPC/Defs.h"
-//#include "TPCQC/Helpers.h"
-#include "TPCBase/Painter.h"
 #include "DataFormatsTPC/ClusterNative.h"
-#include "DataFormatsTPC/ClusterNativeHelper.h"
-#include "SimulationDataFormat/MCTruthContainer.h"
-#include "DataFormatsTPC/Constants.h"
+#include "TPCBase/Painter.h"
 
 // QC includes
 #include "QualityControl/QcInfoLogger.h"
 #include "TPC/Clusters.h"
-
-using namespace o2::framework;
-using namespace o2::header;
-using namespace o2::tpc;
-using namespace o2::dataformats;
+#include "TPC/Utility.h"
 
 namespace o2::quality_control_modules::tpc
 {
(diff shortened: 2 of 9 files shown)


Modules/TPC/src/Calibration.cxx
```diff
@@ -0,0 +1,89 @@
...
+  o2::tpc::painter::makeSummaryCanvases(pedestal, 300, 0, 0, true, &vecPtrPed);
+
+  //==========Noise===========
+  auto& noise = o2::tpc::CDBInterface::instance().getNoise();
+
+  std::vector<TCanvas*> vecPtrNoise;
+  for (auto& canv : mNoiseCanvasVec) {
+    vecPtrNoise.emplace_back(canv.get());
+  }
+  o2::tpc::painter::makeSummaryCanvases(noise, 300, 0, 0, true, &vecPtrNoise);
```
wiechula, line 75: Just for the fun of it and to save some copy'n'paste you could try to do
```suggestion
auto toVector = [](auto& input) { 
  std::vector<TCanvas*> output;
  for (auto& in : input) {
    output.emplace_back(in.get());
  }
  return output;
}

  //========Pedestals==========
  auto& pedestal = o2::tpc::CDBInterface::instance().getPedestals();
  o2::tpc::painter::makeSummaryCanvases(pedestal, 300, 0, 0, true, &toVector(mPedestalCanvasVec));

  //==========Noise===========
  auto& noise = o2::tpc::CDBInterface::instance().getNoise();
  o2::tpc::painter::makeSummaryCanvases(noise, 300, 0, 0, true, &toVector(mNoiseCanvasVec));
```

and perhaps even put the lambda above in some helper header you can also include in the cluster thing.

Modules/TPC/src/Clusters.cxx
```diff
@@ -69,39 +66,64 @@ void Clusters::initialize(o2::framework::InitContext& /*ctx*/)
...
+    auto canvas = canvasPtr.get();
+    getObjectsManager()->startPublishing(canvas);
+    getObjectsManager()->addMetadata(canvas->GetName(), "custom", "34");
+  }
+
+  for (auto& canvasPtr : mTimeBinCanvasVec) {
+    auto canvas = canvasPtr.get();
+    getObjectsManager()->startPublishing(canvas);
+    getObjectsManager()->addMetadata(canvas->GetName(), "custom", "34");
   }
```
wiechula, line 127: This is a bit ugly since it is so much copy'n'paste. Could also be made nicer using a lamba. E.g.
```suggestion
addAndPublish = [](auto& canVec, std::vector<std::string_view> canvNames) {
  for (const auto& canvName : canvNames) {
    canvVec.emplace_back(std::make_unique<TCanvas>(canvName.data()));
    auto canvas = canVec.back().get();
    getObjectsManager()->startPublishing(canvas);
    getObjectsManager()->addMetadata(canvas->GetName(), "custom", "34");
  }
}

  addAndPublish(mNClustersCanvasVec, {"c_Sides_N_Clusters", "c_ROCs_N_Clusters_1D", "c_ROCs_N_Clusters_2D"});
  addAndPublish(mQMaxCanvasVec, {"c_Sides_Q_Max", "c_ROCs_Q_Max_1D", "c_ROCs_Q_Max_2D"});
  addAndPublish(mQTotCanvasVec, {"c_Sides_Q_Tot", "c_ROCs_Q_Tot_1D", "c_ROCs_Q_Tot_2D"});
  addAndPublish(mSigmaTimeCanvasVec, {"c_Sides_Sigma_Time", "c_ROCs_Sigma_Time_1D", "c_ROCs_Sigma_Time_2D"});
  addAndPublish(mSigmaPadCanvasVec, {"c_Sides_Sigma_Pad", "c_ROCs_Sigma_Pad_1D", "c_ROCs_Sigma_Pad_2D"});
  addAndPublish(mTimeBinCanvasVec, {"c_Sides_Time_Bin", "c_ROCs_Time_Bin_1D", "c_ROCs_Time_Bin_2D"});
```

Same can be done in `Calibration.cxx` above. perhaps also by putting the lambda in some common space.

Modules/TPC/src/Calibration.cxx
```diff
@@ -0,0 +1,89 @@
...
+
+//root includes
+#include "TCanvas.h"
+
+using namespace o2::quality_control::postprocessing;
+
+namespace o2::quality_control_modules::tpc
+{
+
+void Calibration::initialize(Trigger, framework::ServiceRegistry&)
```
wiechula, line 32: See comment in 
`void Clusters::initialize(o2::framework::InitContext& /*ctx*/)`

Modules/TPC/src/Clusters.cxx
```diff
@@ -199,17 +217,41 @@ void Clusters::monitorData(o2::framework::ProcessingContext& ctx)
 
   mQCClusters.analyse();
 
-  int histIter = 0;
-  for (auto& calPadWrapper : mWrapperVector) {
-    auto& histA = *mHistoVector[histIter].get();
-    auto& histC = *mHistoVector[histIter + 1].get();
-    const auto& calPad = calPadWrapper.getObj();
+  std::vector<TCanvas*> vecPtrNClusters;
```
wiechula, line 220: See comment in 
`void Calibration::update(Trigger t, framework::ServiceRegistry&)`

Modules/TPC/CMakeLists.txt
```diff
@@ -8,7 +8,8 @@ target_sources(QcTPC PRIVATE src/PID.cxx
                              src/PIDClusterCheck.cxx
                              src/TrackClusterCheck.cxx
                              src/ROCReductor.cxx
-                             src/Clusters.cxx)
+                             src/Clusters.cxx
+                             src/Calibration.cxx)
```
wiechula, line 12: No sure how good the name calibration is. Perhaps `CalDetPublisher` or so?

wiechula (changes requested):
Hi @tklemenz ,
very nice! I have a few suggestions to make the code a bit shorter and more readable.

Modules/TPC/src/Calibration.cxx
tklemenz: Hi @wiechula, I agree on removing the copy&paste. However, the suggested solution is not working. The lambda gives back a temporary instance of output and the makeSummaryCanvases function complains about `taking address of temporary`. If I place line 59 outside the lambda and let it take a reference to the vector as an argument it still fails with the same error.  I am trying around a bit to fix it.

Modules/TPC/src/Clusters.cxx
tklemenz: This one works with a few minor changes.

Modules/TPC/CMakeLists.txt
tklemenz: I agree, the name is not the best. However, I went with the "naming convention" we had so far in QC. Just name the tasks after what they are looking at. Since the code is in QC it should be clear that this does not calibrate anything. But again, I agree that especially in this case the name is not ideal.

Modules/TPC/src/Clusters.cxx
tklemenz: What would you suggest as a common space? Just put a Utility.h in Modules/TPC/include/TPC or go even more common?

Modules/TPC/src/Calibration.cxx
wiechula: Ok I didn't try this, but make sense. You could do
```c++
auto canvPedestal = toVector(mPedestalCanvasVec);
o2::tpc::painter::makeSummaryCanvases(pedestal, 300, 0, 0, true, canvPedestal);
```
without a reference. Should be `std::move` by the compiler, so no big performance issue.

Modules/TPC/src/Clusters.cxx
wiechula: Yes, `Utiltiy.h` in TPC sounds good.

Modules/TPC/CMakeLists.txt
wiechula: Ok, depends on what it will all look at. At the moment it only looks at CalDet Calibration objects. If we want to use it for other stuff as well the name could stay. But I think it would be more natural to have it only look at the CalDet stuff, so Pulser, Pedestal/Noise, Laser Central Electrode.

tklemenz: True, I changed it to CalDetPublisher

Modules/TPC/src/Calibration.cxx
tklemenz: That's also (almost) what I found. :)

tklemenz:
Hi @wiechula, I implemented the suggested changes and in addition added the possibility to define the published metadata in the addAndPublish function via a std::map. I think this is quite handy. By default it uses key custom and value 42.

Modules/TPC/CMakeLists.txt
tklemenz: Then I should also change the path in the QCG, right? Or should we leave it at qc/TPC/MO/Calibration?

wiechula (approved):
@tklemenz , thanks. Much nicer!

tklemenz:
Just in case you did not see the comment I made above concerning the renaming to CalDetPublisher, I quote myself: "Then I should also change the path in the QCG, right? Or should we leave it at qc/TPC/MO/Calibration?"

Another thing: We were talking about something that handles the conversion of the clusters from DPL input to ClusterNativeAccess so that a bunch of copy&paste can be removed for everyone taking clusters from the DPL. I am currently trying around making this a simple function in the utility file. It takes the InputRecord from the ProcessingContext and returns a ClusterNativeAccess struct that can be looped over afterwards. Should we include the function + the implied changes in the Cluster task in this PR if I can make it run quickly?

wiechula:
How about `qc/TPC/MO/PadCalibration`

If it can be added quickly, why not. Otherwise I would make it in another PR.

tklemenz:
Added a `clusterHandler` function to the Utility functions. I simply put what was in the Cluster task and made it that function. Tbh I do not know enough about the cluster handling to be sure that this cannot be done in a better way but it works. It also does not handle MC info at the moment but I don't think we need this now. 

In addition all headers are cleaned up and I added doxygen documentation. Could you please check if what I commented for the lambdas in Utility.h makes sense?

Everything is now in one commit instead of two because with the clusterHandler the commits were convoluted. But that doesn't matter, I guess.

tklemenz:
Almost forgot one minor change in the Cluster task. I changed the destructor from "empty" in the source file to `default` in the header. I think this is much nicer.

wiechula (approved):
@tklemenz , thanks for the changes. Looks fine. Please remove the [WIP] that it can be merged.

tklemenz:
This needs a small change in the `CDBInterface` to include  the getAnyObject function. I will do the PR.

wiechula (approved):
@tklemenz , looks fine for me

tklemenz:
I changed `getCalPad` accordingly in the task.

Barthelemy (approved):

tklemenz:
@wiechula Should I push again to trigger the checks now that the needed PR in O2 is merged?

wiechula:
@tklemenz , yes you can try. Otherwise we'll need to ping the experts to see if manual interaction is needed to trigger a new o2 reference tag.

tklemenz:
Both failing checks are unrelated to this PR.

wiechula:
@Barthelemy , it seems the failures are unrelated. Can this be merged or how should we continue?

knopers8:
@wiechula Barth is on holidays until the end of the week.
Indeed the failing tests seem to be unrelated, I will merge.

wiechula:
Thanks @knopers8 !

PR #612, opened 2021-02-09
[MFT-7] MFT Digit QC Task for the MW2

tomas-herman:
The PR contains updates to the Digit QC task which were supposed to be used during the MW2. It consists of series of .json files which should be used to test different aspects of the workflows.


Barthelemy (approved):
I have to see, because the version of QC for the FLP Suite is v1.10, already released. We can probably patch it with your change or simply copy the config files on the FLPs.

PR #613, opened 2021-02-10
[O2-2018] Allow usage of Microsoft GSL v2 and v3

aphecetche:
See https://github.com/AliceO2Group/AliceO2/pull/5424 for the rationale.
(but note that even without GSL changes' incentive, this PR actually only simplifies the code ;-) )


Barthelemy (approved):

knopers8:
@TimoWilken @ktf I don't understand why the o2 and o2-dataflow failed, but there are no errors and failed tests. Could you have a look please?

TimoWilken:
@knopers8 it seems that the build succeeded, and then later failed without logs, so the "success" logs didn't get overwritten. I've found successful builds of this PR and commit for both red checks in the logs, so I've marked them as green.

knopers8:
@TimoWilken Thanks a lot!

PR #614, opened 2021-02-11
[QC-543] Fix TestWorkflow crash on Mac

Barthelemy:
Add a guard against a null pointer. 

Yet, the underlying problem is not yet understood.


knopers8 (approved):

Barthelemy:

PR #615, opened 2021-02-11
[OCONF-301] conventions

Barthelemy:


Barthelemy:
@TimoWilken It seems that many builds randomly fail in QC. Could you have a look ?

knopers8 (changes requested):
There are several unchanged `QualityControl` library names in Framework/CMakeLists.txt, which makes the build fail on my machine. The mac build log actually contains the message:
```
CMake Error at Framework/CMakeLists.txt:86 (target_link_libraries):
  Cannot specify link libraries for target "QualityControl" which is not
  built by this project.
  ```

Barthelemy:
I had messed up the commits I pushed.

Barthelemy:
@ktf any idea why we still get this error with FairLogger ?

ktf:
No, not really. Maybe that test does not use the head of alidist? However, clearly a bunch of stuff in O2 should use `Framework/Logger.h` rather than "FairLogger.h" now that we know.

Barthelemy:
Failure is unrelated, merging.

PR #616, opened 2021-02-15
Small fixes

aphecetche:
About the DebugGUI change : here I assume that QC can work without it, otherwise may be the `find_package(GLFW)` should specify `REQUIRED` (or the existence of the target DebugGUI be mandatory) ? 
Going a step further :  why is there an explicit find_package(GLFW) at all ? GLFW is only a transitive dependency, isn't it ?

### DIFF
--- a/Framework/CMakeLists.txt
+++ b/Framework/CMakeLists.txt
@@ -94,7 +94,6 @@ target_link_libraries(QualityControl
                              AliceO2::Monitoring
                              AliceO2::Configuration
                              AliceO2::Occ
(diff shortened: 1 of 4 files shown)


Barthelemy:
@aphecetche The `DebugGui` stuff dates from a time where we had no transitive dependencies. Your changes are ok, we could try without the find_package but I don't think it hurts for the time being. 

@TimoWilken @ktf Could you check why the builders are breaking ? thank you

Barthelemy (approved):

Barthelemy:

PR #618, opened 2021-02-19
Create README for the TRD

tdietel:


knopers8:
I will force-merge, because it is only documentation.
@TimoWilken Could you please have a look why we didn't receive check results for the three builds for almost a week? It might be related to the fact, that only an .md file was added.

TimoWilken:
@knopers8 the CI didn't run because @tdietel hadn't committed to this repository before. In that case, someone needs to give the PR an approving review on GitHub for the CI to start. This arrangement is so that "evil" PRs can't steal secrets from the CI environment.

knopers8:
@TimoWilken Thanks. I was actually aware of this mechanism in general (and it's good it's there), but apparently I didn't notice that @tdietel hadn't committed here before. My bad.

tdietel:
Thanks for the merge, and looking into the CI and notification. We might have more people from the TRD commit to QC, we will ping you by email when this happens.

PR #619, opened 2021-02-24
TOF: add slot name, add slot for RDH in part. slot plot

njacazio:
- Code cleanup and minor improvements
- Add RDH read in the partecipating slot plot
- Add slot labels in histograms
- Move Participating slot histo to task
- Separate counters and histos with different naming conventions i.e. mHisto.. and mCounter..
- Update documentation
- Minor performance fixes to counter class

### DIFF
--- a/Modules/TOF/include/Base/Counter.h
+++ b/Modules/TOF/include/Base/Counter.h
@@ -44,10 +44,13 @@ class Counter
   ~Counter() = default;
 
   /// Functions to increment a counter
-  /// @param v Index in the counter array to increment, @param weight weight to add to the array element
-  void Add(const unsigned int& v, const unsigned int& weight);
+  /// @param index Index in the counter array to increment
+  /// @param weight weight to add to the array element
+  void Add(const unsigned int& index, const uint32_t& weight);
 
-  void Count(const unsigned int& v);
+  /// Functions to count a single event
+  /// @param index Index in the counter array to increment by one
+  void Count(const unsigned int& index) { Add(index, 1); }
 
   /// Function to reset counters to zero
   void Reset();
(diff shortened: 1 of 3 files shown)


Modules/TOF/include/Base/Counter.h
```diff
@@ -44,9 +44,12 @@ class Counter
   ~Counter() = default;
 
   /// Functions to increment a counter
-  /// @param v Index in the counter array to increment, @param weight weight to add to the array element
-  void Add(const unsigned int& v, const unsigned int& weight);
+  /// @param v Index in the counter array to increment
+  /// @param weight weight to add to the array element
+  void Add(const unsigned int& v, const uint32_t& weight);
```
Barthelemy, line 49: why not change both to uint32_t ? `v` is not really clear as a name, perhaps you could change it.

Modules/TOF/include/Base/Counter.h
```diff
@@ -97,7 +100,7 @@ class Counter
 // #define ENABLE_PRINT_HISTOGRAMS_MODE // Flag used to enable more printing and more debug
 
 template <const unsigned int size, const char* labels[size]>
-void Counter<size, labels>::Add(const unsigned int& v, const unsigned int& weight)
+void Counter<size, labels>::Add(const unsigned int& v, const uint32_t& weight)
```
Barthelemy, line 103: same

Barthelemy (approved):
Just a very small comment

Modules/TOF/include/Base/Counter.h
njacazio: I agree, I also realized it while opening the PR, let me change this

njacazio: For the type I used unsigned int because this is mostly used in for loops, it can be templated of course, @Barthelemy what do you think?

Barthelemy: @njacazio Yes, that makes sense.

PR #621, opened 2021-02-25
[QC-545] Fix memory leaks associated with Monitor Objects

knopers8:
When Tasks publish MonitorObjects, they do not own underlying objects - the user algorithm does. However, after they are published, received and deserialized, they should own these objects as there is no one else to delete them.


Barthelemy (approved):
Good catch, you are totally correct.

Barthelemy:

PR #622, opened 2021-02-26
Add information about switching detector in flp suite

Barthelemy:


knopers8 (approved):

Barthelemy:

PR #627, opened 2021-03-03
TOF: Improve Counter class

njacazio:
- Add unit test of counter


Barthelemy:
one of the two failed checks is legit

njacazio:
@Barthelemy should be ok now

Barthelemy (approved):

Barthelemy:
Errors are false positives.

PR #628, opened 2021-03-06
Correct FairLogger usage

ktf:
Use fairlogger/Logger.h as reccomended in O2-2082


Barthelemy:
@ktf thank you. I see that now the errors reported are due to instances of the same problem in O2. I will make a PR there.

Barthelemy:
https://github.com/AliceO2Group/AliceO2/pull/5650

PR #636, opened 2021-03-11
Mid

gtaillepied:
MID folder moved under MUON
added rawDataChecker and UserLogicChecker workflows


Barthelemy:
I tried to fix the conflict but I think that you need to rebase.

aphecetche:
@gtaillepied would also probably be better to split that PR in 2 : one trivial with just the directory change (MID->MUON/MID) and another with the actual code additions.

gtaillepied:
I removed the code additions and cleaned up the history a bit. Now this is only moving the MID folder in MUON.

Barthelemy (approved):

dstocco:
@gtaillepied the Checker complains about conflicting files. Probably you need to update the master and rebase your branch to it. You will have to resolve the conflicts manually though...

PR #642, opened 2021-03-15
[WIP] [QC-558] Fix the facility in IL

Barthelemy:
Make sure that we set the facility after all the constructors have been called.

### DIFF
--- a/Framework/src/AggregatorRunner.cxx
+++ b/Framework/src/AggregatorRunner.cxx
@@ -89,6 +89,7 @@ std::string AggregatorRunner::createAggregatorRunnerName()
 
 void AggregatorRunner::init(framework::InitContext&)
 {
+  ILOG_INST.setFacility("Aggregator");
   try {
     initDatabase();
     initMonitoring();
--- a/Framework/src/CheckRunner.cxx
+++ b/Framework/src/CheckRunner.cxx
@@ -135,7 +135,6 @@ CheckRunner::CheckRunner(std::vector<Check> checks, std::string configurationSou
     mTotalNumberQOStored(0),
     mTotalNumberMOStored(0)
 {
-  ILOG_INST.setFacility("Check");
   try {
     mConfigFile = ConfigurationFactory::getConfiguration(configurationSource);
   } catch (...) {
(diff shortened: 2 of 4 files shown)


Barthelemy:
@TimoWilken The clang format test is failing on the prerequisite. Could you have a look ?

Barthelemy:
@ktf

Barthelemy:
@ktf @TimoWilken also `o2` build failed without logs

TimoWilken:
@Barthelemy looking into O2; the clang-format failure has already been reported to GitHub.

Barthelemy:
@TimoWilken I don't undersatnd why `o2` test is red.

TimoWilken:
I've cleaned out the previous build and restarted the o2 check. Hopefully that should fix the problem.

knopers8:
Oh, the old fairlogger problems from O2:
```
/mnt/mesos/sandbox/sandbox/qualitycontrol/sw/SOURCES/O2/build_QualityControl_o2/0/Detectors/ZDC/base/src/ModuleConfig.cxx:13:10: fatal error: FairLogger.h: No such file or directory
 #include <FairLogger.h>
          ^~~~~~~~~~~~~~
compilation terminated.
gmake[2]: *** [Detectors/ZDC/base/CMakeFiles/O2lib-ZDCBase.dir/src/ModuleConfig.cxx.o] Error 1
[  2%] Built target O2test-datacompression-Fifo
/mnt/mesos/sandbox/sandbox/qualitycontrol/sw/SOURCES/O2/build_QualityControl_o2/0/Detectors/ZDC/base/src/Geometry.cxx:12:10: fatal error: FairLogger.h: No such file or directory
 #include "FairLogger.h"
          ^~~~~~~~~~~~~~
          ```

knopers8 (approved):
Indeed, that makes sense.

I am actually wondering if we shouldn't even set the device name there, so we can differentiate between the tasks and checkers.

Barthelemy:
> I am actually wondering if we shouldn't even set the device name there, so we can differentiate between the tasks and checkers.

Yes, good idea. We lack a field actually. Because "facility" is `Task`, `Check` but we need to set a "process name"

Barthelemy:
Closing in favour of https://github.com/AliceO2Group/QualityControl/pull/646

PR #646, opened 2021-03-19
[QC-560] Configuration of infologger

Barthelemy:
- [x] doc for config
- [x] name of facility

### DIFF
--- a/Framework/basic-aggregator.json
+++ b/Framework/basic-aggregator.json
@@ -22,8 +22,8 @@
         "url": "ccdb-test.cern.ch:8080"
       },
       "infologger": {                     "": "Configuration of the Infologger (optional).",
-        "filterDiscardDebug": "0",        "": "Set to 1 to discard debug and trace messages (default: 0)",
-        "filterDiscardLevel": "21",        "": "Message at this level or above are discarded (default: 21 - Trace)"
+        "filterDiscardDebug": "false",    "": "Set to 1 to discard debug and trace messages (default: false)",
+        "filterDiscardLevel": "21",       "": "Message at this level or above are discarded (default: 21 - Trace)"
       }
     },
     "tasks": {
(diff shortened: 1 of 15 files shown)


Framework/basic-aggregator.json
```diff
@@ -20,6 +20,10 @@
       },
       "conditionDB": {
         "url": "ccdb-test.cern.ch:8080"
+      },
+      "infologger": {                     "": "Configuration of the Infologger (optional).",
+        "filterDiscardDebug": "0",        "": "Set to 1 to discard debug and trace messages (default: 0)",
```
knopers8, line 25: Why 0/1, while in the rest of the config file we use true/false?

Framework/include/QualityControl/QcInfoLogger.h
```diff
@@ -18,6 +18,7 @@
 
 #include <InfoLogger/InfoLogger.hxx>
 #include <InfoLogger/InfoLoggerMacros.hxx>
+#include <boost/property_tree/ptree.hpp>
```
knopers8, line 21: Please use `ptree_pwd` in the header and the full `ptree` in the implementation file. We use `QcInfoLogger` everywhere, so it could have a noticeable impact on the build time.

knopers8 (changes requested):
Thank you, it should be very helpful. I would ask you to address my two comments.

Framework/basic-aggregator.json
Barthelemy: It was because I took it directly from the infologger's parameters. You are right, it should be true/false.

Framework/include/QualityControl/QcInfoLogger.h
Barthelemy: correct

knopers8 (approved):

PR #647, opened 2021-03-21
[MCH] [MRRTF-108] Prepare for digit move

aphecetche:
We will be moving Digit.h from MCHBase to DataFormatsMCH in AliceO2. 
This PR hopefully makes the QC immune to that forthcoming change.

### DIFF
--- a/Modules/MUON/MCH/CMakeLists.txt
+++ b/Modules/MUON/MCH/CMakeLists.txt
@@ -1,5 +1,6 @@
 set(MODULE_NAME "O2QcMuonChambers")
 
+
 # ---- Files ----
 
 set(SRCS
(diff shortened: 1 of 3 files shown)


aphecetche:
@aferrero2707 this is to avoid breaking the QC compilation when we'll change the home of our MCH Digit.h (in a forthcoming PR in AliceO2)

Barthelemy (changes requested):
The file Modules/Example/README.md seems to have been automatically added. Could you remove it ? 
Otherwise, all good !

aphecetche:
@Barthelemy have no idea how this empty Modules/Example/README.md ended up there, but it should be fixed now.

Barthelemy (approved):

aphecetche:
@Barthelemy all green

PR #648, opened 2021-03-22
Decrease level and severity of some recurrent logs

knopers8:
To my experience during MW3, some of the logs were a bit too omnipresent. I also updated some logs to use the usual macro.


Barthelemy (approved):
Thank you, very good.

Barthelemy:
The failed test is a lie.

PR #649, opened 2021-03-24
[MFT-8] MFT Track QC Task update

tomas-herman:
Adding new histograms for the track task


Barthelemy:
@TimoWilken @ktf Could you check why `o2-dataflow` is failing ?

tomas-herman:
The O2+QC versions from which I started developing this PR are the same as in https://github.com/AliceO2Group/QualityControl/pull/650 and https://github.com/AliceO2Group/QualityControl/pull/651 which are merged successfully. Hence I am pretty sure the problem is a spontaneous error on the side of the check.

Barthelemy:
I agree, force merging.

PR #656, opened 2021-03-26
[MCH] temporarily disable pedestals processing

aferrero2707:
The code that processes the input from the calibrator is temporarily
removed due to the class names changes in AliceO2Group/AliceO2#5772

The code will be re-introduced once the O2 PR is merged.


aferrero2707:
Closing as a better solution has been introduced in AliceO2Group/AliceO2#5772.

PR #658, opened 2021-03-30
[QC-565] Set run number on objects

Barthelemy:


Barthelemy:
There seems to be room for improvement with regards to the *Runners family as they share quite some functionalities and code. To be seen.

Framework/src/CcdbDatabase.cxx
```diff
@@ -181,6 +181,7 @@ void CcdbDatabase::storeMO(std::shared_ptr<const o2::quality_control::core::Moni
   metadata["qc_detector_name"] = mo->getDetectorName();
   metadata["qc_task_name"] = mo->getTaskName();
   metadata["ObjectType"] = mo->getObject()->IsA()->GetName(); // ObjectType says TObject and not MonitorObject due to a quirk in the API. Once fixed, remove this.
+  metadata["Run"] = std::to_string(mo->getRunNumber());
```
knopers8, line 184: Why "Run" instead of "RunNumber"? If we add more fields like "RunType", "RunDuration", then it will be inconsistent.

Framework/src/MonitorObject.cxx
```diff
@@ -111,4 +113,14 @@ void MonitorObject::setDescription(const string& description)
...
 }
 
+int MonitorObject::getRunNumber() const
+{
+  return mRunNumber;
+}
+
+void MonitorObject::setRunNumber(int runNumber)
+{
+  MonitorObject::mRunNumber = runNumber;
```
knopers8, line 123: Why explicitly?

knopers8 (commented):
Nice, thank you. Please see my two questions.

Framework/src/CcdbDatabase.cxx
Barthelemy: That is a very good point. I had asked Costin about it but I think that we overlooked this problem. I will indeed move to "RunNumber".

Framework/src/MonitorObject.cxx
Barthelemy: because CLion decided so ;) it is auto-generated. I will fix it.

Barthelemy:
I will also fix the tests that apparently started failing.

Framework/script/RepoCleaner/README.md
```diff
@@ -21,7 +21,7 @@ There can be any number of these rules. The order is important as we use the fir
 - `delay`: the duration in minutes of the grace period during which an object is not removed, even if it matches the above path. 
 - `policy`: the name of a policy to apply on the matching objects. Here are the currently available policies (full description in the corresponding files):
    - `1_per_hour`: keep the first and extend its validity to 1 hour, remove everything in the next hour, repeat.
-   - `1_per_run`: requires the "run" metadata to be set. Keep only the most recent version of an object for a given run. 
+   - `1_per_run`: requires the "Run" or "RunNumber" metadata to be set. Keep only the most recent version of an object for a given run.
```
knopers8, line 24: Should I understand that supporting the original ITS workaround is no longer needed?

Framework/src/TaskRunner.cxx
```diff
@@ -413,14 +409,14 @@ void TaskRunner::startOfActivity()
...
 
-  // We take the run number as set from the FairMQ options if it is there, otherwise the one from the config file
-  int run = mRunNumber > 0 ? mRunNumber : mConfigFile->get<int>("qc.config.Activity.number");
-  Activity activity(run,
+  // Start activity in module's stask and update objectsManager
+  Activity activity(mRunNumber,
                     mConfigFile->get<int>("qc.config.Activity.type"));
   ILOG(Info, Ops) << "Starting run " << mRunNumber << ENDM;
-  mCollector->setRunNumber(run);
+  mCollector->setRunNumber(mRunNumber);
```
knopers8, line 416: I don't see a `start()` method which would compute the run number, similarly to `CheckRunner` and `AggregatorRunner`. How is it set for `TaskRunner`?

Modules/Skeleton/src/SkeletonTask.cxx
```diff
@@ -51,7 +51,7 @@ void SkeletonTask::initialize(o2::framework::InitContext& /*ctx*/)
 
 void SkeletonTask::startOfActivity(Activity& activity)
 {
-  ILOG(Info, Support) << "startOfActivity" << activity.mId << ENDM;
+  ILOG(Info, Support) << "SkeletonTask: startOfActivity " << activity.mId << ENDM;
```
knopers8, line 54: Please leave it as it was or update all messages in SkeletonTask consequently.

Framework/script/RepoCleaner/README.md
Barthelemy: I still support the "Run" metadata that they use. The doc was wrong to say "run".

Framework/src/TaskRunner.cxx
Barthelemy: It is there but I extracted the computation of the run number to `computeRunNumber`.

Modules/Skeleton/src/SkeletonTask.cxx
Barthelemy: I agree changing it back.

Framework/src/TaskRunner.cxx
knopers8: Yes, but I can't find any usage of `computeRunNumber` in `TaskRunner`. Maybe I'm blind though

Barthelemy: There was already a computeRunNumber in TaskRunner from when I did my tests to get the run number. 
I actally forgot to use the runnerUtils's method so I modified the PR along this line. 
Thanks

knopers8: ah, I see, this is why i couldn't find it in the diff.

knopers8 (approved):

PR #662, opened 2021-04-01
CPV Pedestal Task & Check 

sevdokim:
CPV Pedestal Task & Check created. Task takes digits and digit trigger records as input and produces Pedestal Monitoring objects: 1D historgrams of amplitudes for every channel (~24K histos total) which are used to produce pedestal maps for modules, count pedestal sigmas and so on. Check is performed to see how many bad channels of different reasons we have. Overall detector quality is produced (1 QO) then. Configurable are channel badness criteria and tolerated numbers of bad channels of different types. Config is read from json file. See Modules/CPV/etc/read-raw-from-file/pedestal-task-no-sampling.json for example. And also Modules/CPV/etc/readme for run recipe.


sevdokim:
Hello! Can you please clarify what is a problem with clang-format checking failure? 
I see "exit code 1" in output but only diff commands ran there. Where from is this exit code coming?
Thank you!

Modules/CPV/src/PedestalTask.cxx
```diff
@@ -0,0 +1,483 @@
...
+#include "DataFormatsCPV/TriggerRecord.h"
+
+namespace o2::quality_control_modules::cpv
+{
+
+PedestalTask::~PedestalTask()
+{
+  for (int i = 0; i < kNHist1D; i++) {
+    if (mHist1D[i]) {
+      mHist1D[i]->Delete();
```
Barthelemy, line 34: is there a particular reason not to use `delete mHist1D[i]` ?

Barthelemy (approved):
I did not read in details the logic of the task and check but what I read was ok.

Modules/CPV/src/PedestalTask.cxx
sevdokim: no, there is no reason.

sevdokim:
Dear Barth, there are some errors during build which I don't understand. Could you please tell me what went wrong this time?

Barthelemy:
They are spurrious https://alice.its.cern.ch/jira/browse/O2-2214

I am waiting for the macos test

Modules/CPV/src/PedestalTask.cxx
sevdokim: I tried to understand the difference between histo->Delete() and delete histo and they both call destructor but first case contains additional check IsOnHeap(). Whatever this check is I prefer to get rid of it just to be 100% sure histo is deleted. In future commits I will replace `mHist1D[i]->Delete()` by `delete mHist1D[i]` as you suggested.

PR #663, opened 2021-04-07
[QC-570] script to release

Barthelemy:


Framework/script/release.sh
```diff
@@ -0,0 +1,99 @@
...
+fi
+#git push
+
+
+read -p "Go to JIRA and release this version : https://alice.its.cern.ch/jira/projects/QC?selectedItem=com.atlassian.jira.jira-projects-plugin%3Arelease-page&status=released-unreleased
+Press any key to continue" -n 1 -r
+echo
+
+
+echo "Go to Github https://github.com/AliceO2Group/QualityControl/releases/new?tag=v${newVersion}&title=v${newVersion}&body=$(cat ${scriptDir}/../../doc/ReleaseNotesTemplate.md |jq -sRr @uri)"
```
knopers8, line 88: I didn't have `jq` on Ubuntu. Could you check if it is installed at the beginning?

Framework/script/release.sh
```diff
@@ -0,0 +1,99 @@
...
+git commit -m "${newVersion}"
+
+
+read -p "Ready to push, please confirm [y|n]" -n 1 -r
+echo
+if [[ ! $REPLY =~ ^[Yy]$ ]]
+then
+    exit 3
+fi
+#git push
```
knopers8, line 80: I tihnk the push is actually missing here

knopers8 (commented):
Nice, I like it! Please have a look at my minor suggestions.

Framework/script/release.sh
Barthelemy: ah yes, I did not want to mess up too much during development. Thank you

Framework/script/release.sh
Barthelemy: done

knopers8 (approved):

PR #668, opened 2021-04-14
[QC-571] Remove the default ServiceDiscovery URL (consul-test)

Barthelemy:


Barthelemy:
We want to avoid that the PR checks and anything run by a user ping the DNS for consul-test.

knopers8 (approved):

PR #670, opened 2021-04-15
[QC-415] Set detector field in infologger

Barthelemy:
Set detector field in infologger in TaskRunner. It does not make sense to do it in the other Runners.
The IL contexts are also stored in the QCInfoLogger to reuse them when the config is changed.

### DIFF
--- a/Framework/include/QualityControl/QcInfoLogger.h
+++ b/Framework/include/QualityControl/QcInfoLogger.h
@@ -69,7 +69,7 @@ class QcInfoLogger : public AliceO2::InfoLogger::InfoLogger
   QcInfoLogger(const QcInfoLogger&) = delete;
 
   // remember the contexts
-  AliceO2::InfoLogger::InfoLoggerContext* mContext = nullptr;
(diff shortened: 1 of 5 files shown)


Framework/include/QualityControl/QcInfoLogger.h
```diff
@@ -66,6 +67,10 @@ class QcInfoLogger : public AliceO2::InfoLogger::InfoLogger
   // Disallow copying
   QcInfoLogger& operator=(const QcInfoLogger&) = delete;
   QcInfoLogger(const QcInfoLogger&) = delete;
+
+  // remember the contexts
+  AliceO2::InfoLogger::InfoLoggerContext* mContext = nullptr;
+  AliceO2::InfoLogger::InfoLoggerContext* mDplContext = nullptr;
```
knopers8, line 73: How about making our context a `shared_ptr` so it is deleted?
I would leave the DPL context as it is though, because we only want to observe it, I assume.

Barthelemy: agreed, on it

knopers8 (approved):

PR #677, opened 2021-04-19
[QC-496] Allow Checks to associate (many) Reasons with Qualities

knopers8:
It also makes `QualitiesToTRFCollectionConverter` extract FlagReasons during the conversion.


Barthelemy (approved):
Nothing to say.

PR #680, opened 2021-04-22
[QC-537] Multinode setup test

knopers8:
DO NOT MERGE. I have to give it a few tries and see if it fails when it should.

EDIT: it should work now


Barthelemy (approved):
very nice

PR #681, opened 2021-04-22
Fixes for GCC 10.2

ktf:


Barthelemy (approved):

ktf:
This is required otherwise the lambdas defined in the header might produce duplicate symbols. Maybe a tactically places `static` would also work, but since there is no reason to use lambdas like that I think this PR is cleaner.

PR #687, opened 2021-04-27
[QC-488] Aggregator which selects the worst Quality of all

knopers8:


Barthelemy (approved):
nothing to say, perfect

PR #690, opened 2021-04-30
getter of ITS/MFT dictionary file got more explicit name

shahor02:


Barthelemy:
Hi Ruben, any idea what these build errors are ?

shahor02:
no idea, but it should not be related to PR: lot of framework tests fail and the full log contains plenty of `ERROR: ld.so: object '"libSegFault.so' from LD_PRELOAD cannot be preloaded: ignored.` messages.
Pinging to @ktf and @TimoWilken

Barthelemy (approved):

Barthelemy:
`o2-dataflow` broken until we move to gcc10, ignoring.

PR #691, opened 2021-05-06
[O2-2018] No need to link to ms_gsl...

aphecetche:
... which is to be renamed to Microsoft.GSL::GSL anyway

### DIFF
--- a/Framework/CMakeLists.txt
+++ b/Framework/CMakeLists.txt
@@ -97,7 +97,6 @@ target_link_libraries(O2QualityControl
                              O2::Framework
                              O2::CCDB
                              O2::MCHBase
-                             ms_gsl::ms_gsl
                              O2QualityControlTypes
                              O2::Mergers
                              O2::DataSampling


aphecetche:
For the record, this one is needed by https://github.com/AliceO2Group/AliceO2/pull/6075

aphecetche:
@TimoWilken any idea what are the timeout failures for build/QualityControl/o2 ?

TimoWilken:
I don't know, sorry. I've restarted it for now, let me know if that doesn't fix it.

aphecetche:
@TimoWilken the error is unfortunately still present. @Barthelemy maybe this is something you've already observed ?

aphecetche:
@TimoWilken @Barthelemy don't know what changed but now it's green, ready to be merged 😄

Barthelemy:
I am not sure but it feels like it is an issue with the compiler version and the c++ standard version in Common. @ktf what do you think ?

Barthelemy (approved):

Barthelemy:
It is this PR that was merged although `o2-dataflow` was red: https://github.com/AliceO2Group/AliceO2/pull/6102

shahor02:
sorry for that, I think the dataflow in the original PR was green at some stage, then I had some incremental changes in other files. The boost::fs was creating problems, so the change is needed. 
@ktf do you understand why in some cases it fails to find the header?

Barthelemy:
`o2-dataflow` is broken until we move to gcc10, ignoring the check failure. 
See https://github.com/AliceO2Group/AliceO2/pull/6102

PR #692, opened 2021-05-06
[QC-577] [QC-578] Allow for more than one query in multinode sampling policy

knopers8:
It also gives a more meaningful name for the proxies (sampling policy name instead of binding).

Needs https://github.com/AliceO2Group/AliceO2/pull/6072


Barthelemy (approved):
The name is better indeed, thank you.

Barthelemy:
o2-dataflow is broken until we move to gcc10, ignoring the check failure.

PR #693, opened 2021-05-07
ITS FEE task fix DPLRawParser warning message

zla2019:
Add a filter to avoid DPLRawParser warning message


Barthelemy:
o2-dataflow is broken until we move to gcc10, ignoring the check failure.

PR #694, opened 2021-05-10
TPC: rawDataQA

tklemenz:
Add a QC task for TPC raw digits

### DIFF
--- a/Modules/TPC/CMakeLists.txt
+++ b/Modules/TPC/CMakeLists.txt
@@ -10,7 +10,8 @@ target_sources(O2QcTPC PRIVATE src/PID.cxx
                              src/ROCReductor.cxx
                              src/Clusters.cxx
                              src/CalDetPublisher.cxx
-                             src/Utility.cxx)
+                             src/Utility.cxx
+                             src/RawDigits.cxx)
 
 target_include_directories(
   O2QcTPC
@@ -21,7 +22,8 @@ target_include_directories(
 
 target_link_libraries(O2QcTPC
                       PUBLIC O2QualityControl
-                             O2::TPCQC)
+                             O2::TPCQC
+                             O2::TPCWorkflow)
 
 
 
@@ -35,6 +37,7 @@ add_root_dictionary(O2QcTPC
                             include/TPC/Clusters.h
                             include/TPC/CalDetPublisher.h
                             include/TPC/Utility.h
+                            include/TPC/RawDigits.h
                     LINKDEF include/TPC/LinkDef.h
                     BASENAME O2QcTPC)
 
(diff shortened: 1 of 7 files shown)


tklemenz:
The `o2-dataflow` build error seems unrelated to the PR.

Barthelemy:
Indeed, the `o2-dataflow` is due to something else (we need to move to gcc10). The `macOS` is strange as there is nothing. Let's see.

tklemenz:
Is gcc10 the solution? I am getting complaints about `#include <filesystem>` on my local machine when compiling the latest `O2 dev` but `QualityControl master` builds without error.

Barthelemy:
Do you have gcc 10 ? 
The probem here is that `o2-dataflow` uses gcc 7 and filesystem was still in experimental back then.

Barthelemy:
Then this is why you have this issue. If you use the defaults `o2` then it will force gcc10. It is not yet hte case with `o2-dataflow` and it provokes this error. We are working on it: https://github.com/alisw/alidist/pull/3027

tklemenz:
This is weird, because on my local machine I do use `--defaults o2`.
`aliBuild build QualityControl --defaults o2`

Barthelemy:
Are you sure that your alidist is up to date ? Because `defaults-o2.sh` reads: 
```
overrides:
  GCC-Toolchain:
     version: "v10.2.0-alice2"
     tag: "v10.2.0-alice2"
```

tklemenz:
Yes, this is exactly what I see in my `defaults-o2.sh`, too.
```
overrides:
  GCC-Toolchain:
     version: "v10.2.0-alice2"
     tag: "v10.2.0-alice2"
```

stheckel:
Hi,
I observe the same behaviour as Thomas concerning the gcc: I have aliBuild version 1.8.0 and alidist/O2/QC (master or dev) from Friday (May 7) morning. Compiling O2 and QC with
`aliBuild build QualityControl --defaults o2`
gcc is not compiled, but taken from the system, and the system version is 7.5.0. At least, when loading QC (and hence O2) and typing `gcc --version` I get the 7.5.0.
As Thomas I am on Ubuntu 18.04.

tklemenz:
FYI: I changed the required version of gcc in `gcc-toolchain.sh` and it worked.
```
case $REQUESTED_VERSION in
  v10*) MIN_GCC_VERSION=100200 ;;
  *) MIN_GCC_VERSION=100200 ;;
esac
```

Barthelemy:
@ktf Is it expected that gcc 7 is picked from the system ? should we not enforce gcc 10 ?

ktf:
@Barthelemy alibuild will pick up GCC 10.2  from the system only if such a version is requested in GCC-ToolChain (i.e. not for aliphysics).

Barthelemy:
the macOS build seems spurrious., merging.

tklemenz:
Sure, I can open a ticket and yes, I do not use aliphysics.

PR #696, opened 2021-05-11
QC Framework should not depend on MCHBase ;-)

aphecetche:


Barthelemy:
@knopers8 There is a `multi_node test` failure

PR #702, opened 2021-05-17
[QC-583] Make sure we kill the processes in the multinode test

Barthelemy:


Barthelemy:
@ktf @TimoWilken There is now an error with testMemoryResources.cxx

ktf:
I think there is an issue with the whole test here, and it's actually the cause of many zombies we see. Since you are not waiting for the process in background to finish, there are cases in which the parent script terminates without waiting for all the children, which then leads to zombies. Can you add a wait statement for all the processes in background and disable the test until that is done?

Barthelemy:
I'll let @knopers8 comment as he wrote the test. 

I see that there is a `fg` to make sure that the script waits for the `o2-qc-run-producer` to finish and the `o2-qc` is not put in the background. To me it seems that the script waits that the children are done before finishing. 
Also, we don't see this issue on our build, development and CI machines, nor on any other platform. I am not able to reproduce it on the same build machine when running alibuild by hand. 

I can disable the test but I am not sure it is a general issue with the whole test.

ktf:
are you running the build inside docker? The underlying problem I think is that docker does not reap zombies until the container is gone (since a docker container does not have an init / systemd subsystem).

Barthelemy:
no, I just connect there and build. 

I still fail to understand why there are zombies. The script must not exit before the processes are done. Could the script be killed for some reason and leave behind these processes ? it should really not as they are not even in the background.

Barthelemy:
mmm actually it is true that the script could be killed (by ctest for example) and leave the processes behind.

PR #706, opened 2021-05-19
ITS: bug fixes for ML decoding

JianLIUhep:
- bug fixes for ML decoding
- temporarily reverted to the previous method for querying TF ID

### DIFF
--- a/Modules/ITS/src/ITSFhrTask.cxx
+++ b/Modules/ITS/src/ITSFhrTask.cxx
@@ -327,7 +327,7 @@ void ITSFhrTask::monitorData(o2::framework::ProcessingContext& ctx)
   std::chrono::time_point<std::chrono::high_resolution_clock> end;
   int difference;
   start = std::chrono::high_resolution_clock::now();
-
+  /* temporary disable this 
   //get TF id by dataorigin and datadescription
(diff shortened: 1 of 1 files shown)


Barthelemy:
@sawenzel Any idea what this test `o2sim_G3_checklogs` is ? it is failing

Modules/ITS/src/ITSFhrTask.cxx
```diff
@@ -327,15 +327,15 @@ void ITSFhrTask::monitorData(o2::framework::ProcessingContext& ctx)
   std::chrono::time_point<std::chrono::high_resolution_clock> end;
   int difference;
   start = std::chrono::high_resolution_clock::now();
-
+  /* temporary disable this
```
Barthelemy, line 330: if it is temporary you could also drive it via a configuration parameter to avoid having to release a new version when you want to re-enable it. This is fine with me though.

Barthelemy (approved):

Modules/ITS/src/ITSFhrTask.cxx
JianLIUhep: Thanks for your comment. This temporarily disabled part is only for handling the STFid in this task. There is also a minor bug but not break the task. I am not the code owner, so I would let @zla2019 fix it later. He is quite busy this week.  
The main fix in this PR is to make it correctly fill the hits into histograms for the middle layers of ITS, which block our tests in MW5.

Barthelemy:
understood, test failure unrelated, merging

PR #710, opened 2021-05-27
Add doc about sending monitoring metrics of Dispatcher

Barthelemy:


Barthelemy:
the build error is plain wrong.

knopers8 (approved):

PR #714, opened 2021-05-28
Fix ML and TF id bug. Add two parameters to define fired pixel check cut and how to get TF id

zla2019:
"HitNumberCut" to define the cut of fired pixel check.
"GetTFFromBinding" to define how to get TF id ("0" get TF id by binding, "1" get TF id by data description and data source).


Barthelemy:
the error is genuine in the macos build is genuine.

zla2019:
Thanks a lot!! I am going to modify it.

Barthelemy (approved):

PR #715, opened 2021-05-31
[QC-589] Reduce the amount of configuration needed for mutlinode setups

knopers8:
The changes aren't so drastic as it looks, most of the diff is due to change of indentation. Now the `remoteMachine`, `remotePort` in Tasks, and `port` in DS policies are optional. For `localMachines`, I would first wait until we are done with the switch to pushed configuration, but anyway I am not sure if we can completely avoid it.

This PR will be followed by another one to O2/DataSampling, but first we need this PR and a tag to accept `std::optional<uint16_t>` from `DataSampling::PortForPolicy`.


knopers8:
We shouldn't merge it before the multinode test is reenabled.

Barthelemy:
`testInfrastructureGenerator` fails

Framework/src/InfrastructureGenerator.cxx
```diff
@@ -423,7 +453,7 @@ vector<OutputSpec> InfrastructureGenerator::generateCheckRunners(framework::Work
         break;
       }
     }
-    if (!isStored) {
+    if (!isStored) { // fixme: statement is always true
```
Barthelemy, line 456: right...

Framework/src/InfrastructureGenerator.cxx
```diff
@@ -47,6 +47,9 @@ using SubSpec = o2::header::DataHeader::SubSpecificationType;
 namespace o2::quality_control::core
 {
 
+const char* defaultRemotePort = "36543";
```
Barthelemy, line 50: why not string?

Barthelemy (approved):
All clear and good, just one tiny comment

Framework/src/InfrastructureGenerator.cxx
knopers8: no reason. I will change it to string if the checks fail anyway, otherwise I wouldn't block this with running the checks again.

Barthelemy: indeed

PR #717, opened 2021-06-01
Check for TOF DRM Diagnostics

pranjalsarma:
Hi @njacazio, 

PR for DRM Diagnostics is created. Please have look.

Best regards,
Pranjal

### DIFF
--- a/Modules/TOF/src/CheckDRMDiagnostics.cxx
+++ b/Modules/TOF/src/CheckDRMDiagnostics.cxx
@@ -74,7 +74,7 @@ void CheckDRMDiagnostics::beautify(std::shared_ptr<MonitorObject> mo, Quality ch
     if (checkResult == Quality::Bad) {
       ILOG(Info, Support) << "Quality::Bad, Error bins with content >0";
       msg->Clear();
-      msg->AddText("Error bins with content >0");
+      msg->AddText("DRM reporting error words");
       msg->SetFillColor(kGreen);
       //
       //      h->SetFillColor(kGreen);
@@ -82,7 +82,7 @@ void CheckDRMDiagnostics::beautify(std::shared_ptr<MonitorObject> mo, Quality ch
       ILOG(Info, Support) << "Quality::Null, No Error bins with content >0";
       msg->SetTextColor(kWhite);
       msg->SetFillColor(kBlack);
-      msg->AddText("No Error bins with content >0");
+      msg->AddText("DRM words: OK!");
     }
   } else
     ILOG(Error, Support) << "Did not get correct histo from " << mo->GetName();
--- a/Modules/TOF/src/CheckDRMDiagnostics.cxx
+++ b/Modules/TOF/src/CheckDRMDiagnostics.cxx
@@ -43,20 +43,19 @@ Quality CheckDRMDiagnostics::check(std::map<std::string, std::shared_ptr<Monitor
     (void)moName;
     if (mo->getName() == "DRMCounter") {
       auto* h = dynamic_cast<TH2F*>(mo->getObject());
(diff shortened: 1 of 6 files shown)


Modules/TOF/src/CheckDRMDiagnostics.cxx
```diff
@@ -0,0 +1,92 @@
...
+  for (auto& [moName, mo] : *moMap) {
+    (void)moName;
+    if (mo->getName() == "DRMCounter") {
+      auto* h = dynamic_cast<TH2F*>(mo->getObject());
+//      if (h->GetEntries() == 0) {
+//        result = Quality::Medium;
+
+	for (int i = 1; i < h->GetNbinsX(); i++) {
+	for (int j = 1; j < h->GetNbinsY(); j++) {
+          const float content = h->GetBinContent(i, j);
```
njacazio, line 51: @pranjalsarma perfect! Here we can check that the histogram does not have bins with errors inside. The meaning of each bin should be this one: https://github.com/AliceO2Group/AliceO2/blob/dev/DataFormats/Detectors/TOF/include/DataFormatsTOF/CompressedDataFormat.h#L118-L134 I think that we can skip the first bin (HAS_DATA) because this is fine.

Modules/TOF/src/CheckDRMDiagnostics.cxx
```diff
@@ -0,0 +1,92 @@
...
+    auto* h = dynamic_cast<TH1F*>(mo->getObject());
+    TPaveText* msg = new TPaveText(0.5, 0.5, 0.9, 0.75, "NDC");
+    h->GetListOfFunctions()->Add(msg);
+    msg->Draw();
+    msg->SetName(Form("%s_msg", mo->GetName()));
+
+    if (checkResult == Quality::Bad) {
+      ILOG(Info, Support) << "Quality::Bad, Error bins with content >0";
+      msg->Clear();
+      msg->AddText("Error bins with content >0");
```
njacazio, line 78: I would adjust the message accordingly, with something e.g. "DRM reporting some error words"

njacazio (commented):
Hi @pranjalsarma, great work indeed! we just would need to specify the checks and we are set!

Modules/TOF/toffull.json
```diff
@@ -78,6 +78,22 @@
...
+        "className": "o2::quality_control_modules::tof::CheckDRMDiagnostics",
+        "moduleName": "QcTOF",
+        "detectorName": "TOF",
+        "policy": "OnAny",
+        "dataSource": [
+          {
+            "type": "Task",
+            "name": "TaskRaw",
+            "MOs": [
+              "RDHCounterCrate0"
```
njacazio, line 92: This should be the `DRMCounter` no?

Modules/TOF/tofraw.json
```diff
@@ -61,6 +61,22 @@
...
+        "className": "o2::quality_control_modules::tof::CheckDRMDiagnostics",
+        "moduleName": "QcTOF",
+        "detectorName": "TOF",
+        "policy": "OnAny",
+        "dataSource": [
+          {
+            "type": "Task",
+            "name": "TaskRaw",
+            "MOs": [
+              "RDHCounterCrate0"
```
njacazio, line 75: Idem, This should be the name of the histogram that you are processing in the checker

Modules/TOF/toffull.json
pranjalsarma: In the CheckDiagnostics.cxx file it is given as RDH.

Barthelemy (approved):

Barthelemy:
Spurrious check errors, merging.

PR #718, opened 2021-06-01
Do not use CompletionPolicy::InputSet alias

ktf:
Will be soon removed to reduce header visibiltiy.


ktf:
Required by https://github.com/AliceO2Group/AliceO2/pull/6293

Barthelemy:
merging because the errors are not related.

PR #722, opened 2021-06-04
[QC-593] Add callbacks for checkRunner and aggregatorRunner

Barthelemy:


Framework/src/TaskRunner.cxx
```diff
@@ -436,6 +437,7 @@ void TaskRunner::startCycle()
 
 void TaskRunner::finishCycle(DataAllocator& outputs)
 {
+  ILOG(Debug, Ops) << "Finish cycle " << mCycleNumber << " in " << mTaskConfig.taskName << ENDM;
```
knopers8, line 440: No need to say where, "Facility" fields says that. In standalone, the device name is at the beginning of the line.

knopers8 (approved):

Framework/src/TaskRunner.cxx
Barthelemy: correct, fixed.

Barthelemy:
Spurrious check errors, merging.

PR #723, opened 2021-06-07
initial ft0 calibration task with checker

Kavaldrin:


Modules/FT0/CMakeLists.txt
```diff
@@ -74,3 +78,17 @@ foreach(test ${TEST_SRCS})
   set_tests_properties(${test_name} PROPERTIES TIMEOUT 20)
 endforeach()
 
+# ----- Configs -------
+install(FILES
+        ft0-calibration-config.json
+        ft0-multinode-config.json
+        ft0-postprocessing-config.json
+        ft0-basic-config.json
```
Barthelemy, line 86: are you sure that you need these files installed ? as I understand them they are used when running manually. At P2, they will be unused as the configuration files are stored in Consul.

Modules/FT0/include/FT0/CalibrationTask.h
```diff
@@ -0,0 +1,66 @@
...
+ private:
+  static constexpr int CHANNEL_TIME_HISTOGRAM_RANGE = 200;
+  static constexpr const char* CCDB_PARAM_KEY = "CCDBUrl";
+
+  // Object which will be published
+  std::unique_ptr<TH1F> mNotCalibratedChannelTimeHistogram;
+  std::unique_ptr<TH1F> mCalibratedChannelTimeHistogram;
+  std::unique_ptr<TGraph> mChannelTimeCalibrationObjectGraph;
+  std::unique_ptr<TH2F> mCalibratedTimePerChannelHistogram;
+  std::unique_ptr<TH2F> mNotCalibratedTimePerChannelHistogram;
```
Barthelemy, line 60: In general you should avoid raw pointers. However, for reasons beyond us, we need to pass a raw pointer when publishing the object. The risk here is that the pointer is freed by the unique_ptr while the publishing entity is still alive and trying to use it. 

It is not a big risk. You can keep it like that but I want you to be aware of it.

Barthelemy (approved):

Barthelemy:
I merge this PR as the comments are minor. Please consider them nevertheless for a futre PR.

Modules/FT0/include/FT0/CalibrationTask.h
Kavaldrin: But you mean `o2::ft0::FT0ChannelTimeCalibrationObject* mCurrentChannelTimeCalibrationObject`? It's managed by BasicCCDBManager, it will not be published or etc (it is used only to read values).

Modules/FT0/CMakeLists.txt
Kavaldrin: Well I wanted to make more user friendly path to them, using {QUALITYCONTROL_ROOT}.

Barthelemy: Fair enough, please prefix them with `o2-qc-` because they will end up in the common `etc`.

Modules/FT0/include/FT0/CalibrationTask.h
Barthelemy: No, I am talking about mCalibratedChannelTimeHistogram and other objects that you publish to the QC.

PR #725, opened 2021-06-07
MFT: large update of MFT QC tasks (naming, conventions, mapping)

Katka19:
* updated digit reader and digit task
 
* added chip mapping

 * large update of names and conventions


Barthelemy:
it is a *very* large PR. I will need some time to review it.

Barthelemy (approved):
I must admit that I did not look into the details. Nothing catched my eyes. 

In the future, I would ask that  the PRs are smaller and comprise only 1 type of changes if possible.

Katka19:
Thank you. We understand, and will try to avoid such large PRs in the future.

PR #730, opened 2021-06-16
MCH: Implementation of mergeable Occupancy

seperrin:
Optimisation of histogram filling (removal of time-consuming inverse mapping)
Implementation of mergeable Occupancy plots (using a custom class inheriting MergeInterface)


Barthelemy (approved):
all good

PR #731, opened 2021-06-16
ITS: add the combined json file for the cluster and track tasks on EPN

JianLIUhep:


Modules/ITS/itsEPN.json
```diff
@@ -0,0 +1,127 @@
...
+        "dataSource": {
+          "type": "dataSamplingPolicy",
+          "name": "compclus"
+        },
+        "location": "remote",
+        "taskParameters": {
+          "layer": "1111111",
+          "clusterDictionaryPath": "/home/epn/odc/files/ITSdictionary.bin",
+          "runNumberPath": "/home/its/QC/workdir/infiles/RunNumber.dat",
+          "geomPath": "/home/epn/odc/files/o2sim_geometry.root",
```
Barthelemy, line 43: is it fine that they are in different homes ?

Barthelemy (approved):

Barthelemy:
I merge as I expect that the answer to my question is "it is fine"

Modules/ITS/itsEPN.json
JianLIUhep: Yes, it's fine. The `runNumberPath` will become obsolete and be removed soon. We will fetch it from ECS then. Thank you.

PR #739, opened 2021-06-23
Increase the timeout values in the multinode test

knopers8:
The previously used values were still not enough sometimes. The workflows would take even 15 seconds to start.


knopers8:
failure on o2-dataflow is unrelated, merging and releasing.

PR #741, opened 2021-06-23
HWerror plot fixed

peressounko:


peressounko:
test build/QualityControl/o2  failed for reasons not related to PHOS. Can it be restarted or skipped?

Barthelemy (approved):

PR #744, opened 2021-06-24
Fix stdout repo cleaner

Barthelemy:
All output was going to stderr.


knopers8 (approved):
interesting, i wouldn't expect it to be the default behaviour.

PR #746, opened 2021-06-26
TPC CalDetPublisher: fetch CalPad in std::unordered_map from CCDB + add…

tklemenz:
… ZS calibration file check

### DIFF
--- a/Modules/TPC/include/TPC/CalDetPublisher.h
+++ b/Modules/TPC/include/TPC/CalDetPublisher.h
@@ -20,16 +20,25 @@
 #include "QualityControl/PostProcessingInterface.h"
 
 #include <boost/property_tree/ptree_fwd.hpp>
+#include <map>
 
 class TCanvas;
+class TPaveText;
 
 namespace o2::quality_control_modules::tpc
 {
 
 /// \brief Valid CalDet objects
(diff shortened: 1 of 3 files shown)


tklemenz:
This needs [this PR in O2](https://github.com/AliceO2Group/AliceO2/pull/6531) to have the templated `getCalPad` function.

Barthelemy:
It seems that there is a problem with circular changes here. This PR depends on another PR in O2 which depends on this one. 

One option is to force merge on both sides and prepare a new QC version quickly to avoid breaking every PR on O2 and having an unstable master in QC. 

Another option, is to make the change backward compatible. In the specific case, you could keep the old methods (`getCalPad`,  but mark them as deprecated. This way, you can merge the PR in O2 and then the one in QC. 

I prefer the second one but understand that it is more work for you as you have to remove the methods in another iteration.

tklemenz:
Hi Barth, I just discussed with Jens already that I will make a few changes to make it backward compatible.

Barthelemy:
thank you

tklemenz:
@Barthelemy The PR in O2 was just merged. Will the tests rerun on their own or do they need to be triggered?

Barthelemy:
They restart occasionally

Barthelemy:
alidoctor error in mac unrelated, merging.

PR #750, opened 2021-06-29
[EMCAL-526] histos for max number of hits,FECid

cterrevo:
Histos to count the max number of hit channels and correspondent FECid, and RMS,MEAN,MIN,MAX amplitude of EMCAL


mfasDa:
Failures unrelated. @Barthelemy @knopers8 could you please merge this?

Barthelemy (approved):

PR #757, opened 2021-07-02
TPC Utility tmp workaround: remove fatal from clusterHandler

tklemenz:
This is a temporary workaround to get rid of a fatal leading to crashes of the cluster workflow.

Once the underlying problem is fixed, this will be put back in.

### DIFF
--- a/Modules/TPC/src/Utility.cxx
+++ b/Modules/TPC/src/Utility.cxx
@@ -108,12 +108,12 @@ o2::tpc::ClusterNativeAccess clusterHandler(o2::framework::InputRecord& input)
       }
       continue;
     }
-    if ((validInputs & sectorMask).any()) {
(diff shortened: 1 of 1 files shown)


knopers8 (approved):

Barthelemy:
@TimoWilken is everything ok on `alientest06` ? all tests are taking a very large time and some are timing out.

TimoWilken:
Hm, it seems that lots of tests and a compilation fail because of out-of-memory errors. Because they were all running on alientest06, it's hard to know who the culprit is, but I suspect is the increase in the number of jobs in https://github.com/alisw/ali-bot/commit/97e6b42c839e66f20e55cb8a58ae3b6873eb413b.

The macOS test had an old problem, I've restarted it.

tklemenz:
The issue was actually solved at another point meanwhile. I will close the PR.

PR #766, opened 2021-07-15
Fix moving window

knopers8:
@tklemenz This should fix the problem.


knopers8:
Errors unrelated, merging and releasing, since this might affect not only setups with moving window.

PR #767, opened 2021-07-15
MFT: fix for digit checker

tomas-herman:
A quick fix for a bug in the digit checker.


knopers8:
Failures unrelated, merging.

PR #768, opened 2021-07-15
TPC: configurable histogram ranges

tklemenz:
Make output histogram ranges of Cluster and RawDigits task configurable via json file.

### DIFF
--- a/Modules/TPC/include/TPC/Utility.h
+++ b/Modules/TPC/include/TPC/Utility.h
@@ -40,6 +40,18 @@ void addAndPublish(std::shared_ptr<o2::quality_control::core::ObjectsManager> ob
 /// \return std::vector<TCanvas*>
 std::vector<TCanvas*> toVector(std::vector<std::unique_ptr<TCanvas>>& input);
 
+/// \brief Fills std::vector<std::unique_ptr<TCanvas>> with data from calDet
+/// This is a convenience function to call o2::tpc::painter::makeSummaryCanvases in QC tasks to visualize the content of a CalDet object.
+/// \param calDet Object to be displayed in the canvases
+/// \param canvases Vector containing three std::unique_ptr<TCanvas>, will be filled
+/// \param params Information about the ranges of the histograms that will be drawn on the canvases. The params can be set via 'taskParameters' in the config file of corresponding the task.
+/// \param paramName Name of the observable that is stored in calDet
+void fillCanvases(const o2::tpc::CalDet<float>& calDet, std::vector<std::unique_ptr<TCanvas>>& canvases, const std::unordered_map<std::string, std::string>& params, const std::string paramName);
+
+/// \brief Clears all canvases
+/// \param canvases Contains the canvases that will be cleared
+void clearCanvases(std::vector<std::unique_ptr<TCanvas>>& canvases);
+
 /// \brief Converts CLUSTERNATIVE from InputRecord to ClusterNativeAccess
 /// Convenience funtion to make native clusters accessible when receiving them from the DPL
 /// \param input InputReconrd from the ProcessingContext
--- a/Modules/TPC/run/tpcQCClusters_direct.json
+++ b/Modules/TPC/run/tpcQCClusters_direct.json
@@ -28,30 +28,22 @@
         "className": "o2::quality_control_modules::tpc::Clusters",
         "moduleName": "QcTPC",
         "detectorName": "TPC",
-        "cycleDurationSeconds": "60",
+        "cycleDurationSeconds": "10",
         "maxNumberCycles": "-1",
+        "resetAfterCycles": "5",
         "dataSource": {
           "type": "direct",
           "query" : "input:TPC/CLUSTERNATIVE"
         },
         "taskParameters": {
-          "myOwnKey": "myOwnValue"
+          "NClustersNBins": "100",  "NClustersXMin": "0", "NClustersXMax": "100",
+          "QmaxNBins":      "200",  "QmaxXMin":      "0", "QmaxXMax":      "200",
+          "QtotNBins":      "600",  "QtotXMin":      "0", "QtotXMax":      "600",
+          "SigmaPadNBins":  "200",  "SigmaPadXMin":  "0", "SigmaPadXMax":  "2",
+          "SigmaTimeNBins": "200",  "SigmaTimeXMin": "0", "SigmaTimeXMax": "2",
+          "TimeBinNBins":   "1000", "TimeBinXMin":   "0", "TimeBinXMax":   "100000"
         },
         "location": "remote"
-      }
-    },
-    "checks": {
-      "QcCheckClusters": {
-        "active": "true",
-        "className": "o2::quality_control_modules::skeleton::SkeletonCheck",
-        "moduleName": "QcSkeleton",
-        "policy": "OnAny",
-        "detectorName": "TPC",
-        "dataSource": [{
-          "type": "Task",
-          "name": "Clusters",
-          "MOs": ["example"]
-        }]
       }
     }
   },
--- a/Modules/TPC/src/Clusters.cxx
+++ b/Modules/TPC/src/Clusters.cxx
@@ -50,7 +50,6 @@ void Clusters::initialize(o2::framework::InitContext& /*ctx*/)
 
   for (auto& wrapper : mWrapperVector) {
     getObjectsManager()->startPublishing(&wrapper);
-    getObjectsManager()->addMetadata(wrapper.getObj()->getName().data(), "custom", "87");
   }
 }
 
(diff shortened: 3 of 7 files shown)


Modules/TPC/src/Clusters.cxx
```diff
@@ -113,7 +130,31 @@ void Clusters::reset()
 {
   // clean all the monitor objects here
 
-  QcInfoLogger::GetInstance() << "Resetting the histogram" << AliceO2::InfoLogger::InfoLogger::endm;
+  QcInfoLogger::GetInstance() << "Resetting the canvases" << AliceO2::InfoLogger::InfoLogger::endm;
+
```
wiechula, line 134: Also this is quite some copy and paste. Perhaps this could also be simplified at some point.

Modules/TPC/src/RawDigits.cxx
```diff
@@ -97,6 +105,18 @@ void RawDigits::reset()
   // clean all the monitor objects here
 
   QcInfoLogger::GetInstance() << "Resetting the histogram" << AliceO2::InfoLogger::InfoLogger::endm;
+
+  for (const auto& canvas : mNRawDigitsCanvasVec) {
```
wiechula, line 109: same as above.

wiechula (changes requested):
@tklemenz please have a look at my comments.

Modules/TPC/src/Clusters.cxx
```diff
@@ -81,22 +80,40 @@ void Clusters::monitorData(o2::framework::ProcessingContext& ctx)
   mQCClusters.analyse();
 
   auto vecPtrNClusters = toVector(mNClustersCanvasVec);
-  o2::tpc::painter::makeSummaryCanvases(mQCClusters.getNClusters(), 300, 0, 0, true, &vecPtrNClusters);
+  o2::tpc::painter::makeSummaryCanvases(mQCClusters.getNClusters(),
```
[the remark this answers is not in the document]
tklemenz: Good point about the copy&paste. I implemented the function.
I could make the CalDet name match the `paramName`. However, due to the "abuse" of the o2 part of the cluster task to become the RawDigits task there would be one confusing name for the `taskParameters` of the RawDigits. `NRawDigitsNBins`, `NRawDigitsXMin` and `NRawDigitsXMax` would need to be called `N_Clusters_NBin`, etc... To avoid this we would need to add a specific CalPad member for `N_RawDigits` to `O2/Detectors/TPC/qc/include/TPCQC/Clusters.h` which is filled in `o2::tpc::qc::Clusters::fillADCValue`. Then the `o2::tpc::qc::Clusters` member of the qc task would always have one empty CalDet member. `N_Clusters` if it is used for raw digits or `N_RawDigits` if it is used for clusters. 
I don't know if we would want that.

Modules/TPC/src/Clusters.cxx
tklemenz: I added `clearCanvases`.

Modules/TPC/include/TPC/Utility.h
```diff
@@ -40,6 +40,17 @@ void addAndPublish(std::shared_ptr<o2::quality_control::core::ObjectsManager> ob
 /// \return std::vector<TCanvas*>
 std::vector<TCanvas*> toVector(std::vector<std::unique_ptr<TCanvas>>& input);
 
+/// \brief Fills std::vector<std::unique_ptr<TCanvas>> with data from calDet
+/// \param calDet Object to be displayed in the canvases
+/// \param canvases Vector containing three std::unique_ptr<TCanvas>, will be filled
+/// \param params Information about the ranges of the histograms that will be drawn on the canvases
+/// \param paramName Name of the observable that is stored in calDet
+void fillCanvases(const o2::tpc::CalDet<float>& calDet, std::vector<std::unique_ptr<TCanvas>>& canvases, std::unordered_map<std::string, std::string>& params, const std::string paramName);
```
wiechula, line 48:
```suggestion
void fillCanvases(const o2::tpc::CalDet<float>& calDet, std::vector<std::unique_ptr<TCanvas>>& canvases, const std::unordered_map<std::string, std::string>& params, const std::string paramName);
```

Modules/TPC/src/Utility.cxx
```diff
@@ -53,6 +55,22 @@ std::vector<TCanvas*> toVector(std::vector<std::unique_ptr<TCanvas>>& input)
...
+  o2::tpc::painter::makeSummaryCanvases(calDet,
+                                        std::stoi(params[(paramName + "NBins")]), std::stof(params[(paramName + "XMin")]), std::stof(params[(paramName + "XMax")]),
+                                        true,
+                                        &vecPtr);
+}
+
+void clearCanvases(std::vector<std::unique_ptr<TCanvas>>& canvases)
+{
+  for (const auto& canvas : canvases) {
+    canvas.get()->Clear();
```
wiechula, line 70: Is the get really needed?
```suggestion
    canvas->Clear();
```

Modules/TPC/src/Utility.cxx
```diff
@@ -53,6 +55,22 @@ std::vector<TCanvas*> toVector(std::vector<std::unique_ptr<TCanvas>>& input)
   return output;
 }
 
+void fillCanvases(const o2::tpc::CalDet<float>& calDet, std::vector<std::unique_ptr<TCanvas>>& canvases, std::unordered_map<std::string, std::string>& params, const std::string paramName)
+{
+  auto vecPtr = toVector(canvases);
```
wiechula, line 60: Perhaps the canvases should be cleared directly before filling?
```suggestion
  auto vecPtr = toVector(canvases);
  clearCanvases(canvases);
```

Modules/TPC/src/Utility.cxx
```diff
@@ -53,6 +55,22 @@ std::vector<TCanvas*> toVector(std::vector<std::unique_ptr<TCanvas>>& input)
   return output;
 }
 
+void fillCanvases(const o2::tpc::CalDet<float>& calDet, std::vector<std::unique_ptr<TCanvas>>& canvases, std::unordered_map<std::string, std::string>& params, const std::string paramName)
```
wiechula, line 58:
```suggestion
void fillCanvases(const o2::tpc::CalDet<float>& calDet, std::vector<std::unique_ptr<TCanvas>>& canvases, const std::unordered_map<std::string, std::string>& params, const std::string paramName)
```

wiechula (changes requested):
minor suggestions

Modules/TPC/include/TPC/Utility.h
tklemenz: I tried this in the first place but it leads to
````
error: passing 'const std::unordered_map<std::__cxx11::basic_string<char>, std::__cxx11::b
asic_string<char> >' as 'this' argument discards qualifiers [-fpermissive]
   62 |                                         std::stoi(params[(paramName + "NBins")]), std::stof(params[(paramName + "XMin")]), std::stof(params[(paramName + "XMax")]),
      |                                                                               ^
````

File: Modules/TPC/src/Utility.cxx (hunk 6 above)
Comment (tklemenz) [reply Modules/TPC/src/Utility.cxx:60]: We can do this but `o2::tpc::painter::makeSummaryCanvases` does this anyway in lines 337ff.

File: Modules/TPC/src/Utility.cxx (hunk 5 above)
Comment (tklemenz) [reply Modules/TPC/src/Utility.cxx:70]: Indeed not. Smart pointers are smarter than I thought. ;)

File: Modules/TPC/include/TPC/Utility.h (hunk 4 above)
Comment (wiechula) [reply Modules/TPC/include/TPC/Utility.h:48]: Ok, fine.

File: Modules/TPC/src/Utility.cxx (hunk 6 above)
Comment (wiechula) [reply Modules/TPC/src/Utility.cxx:60]: Very clever ;-) Then we don't need it here, of course.

File: Modules/TPC/include/TPC/Utility.h (hunk 4 above)
Comment (tklemenz) [reply Modules/TPC/include/TPC/Utility.h:48]: I will try something like this. A check for missing parameters makes sense anyway.

File: Modules/TPC/src/Utility.cxx [hunk 8]
```diff
@@ -53,6 +55,36 @@ std::vector<TCanvas*> toVector(std::vector<std::unique_ptr<TCanvas>>& input)
...
+  const auto last = params.end();
+  const auto itNbins = params.find(parNBins);
+  const auto itXmin = params.find(parXMin);
+  const auto itXmax = params.find(parXmax);
+  if ((itNbins == last) || (itXmin == last) || (itXmax == last)) {
+    LOGP(warning, "missing parameter {}, {} or {}, falling back to auto scaling", parNBins, parXMin, parXmax);
+  } else {
+    nbins = std::stoi(itNbins->second);
+    xmin = std::stof(itXmin->second);
+    xmax = std::stof(itXmax->second);
```
wiechula, line 75: Make naming consistent
```suggestion
  const std::string parXMax = paramName + "XMax";
  int nbins = 300;
  float xmin = 0;
  float xmax = 0;
  const auto last = params.end();
  const auto itNBins = params.find(parNBins);
  const auto itXMin = params.find(parXMin);
  const auto itXMax = params.find(parXMax);
  if ((itNBins == last) || (itXMin == last) || (itXMax == last)) {
    LOGP(warning, "missing parameter {}, {} or {}, falling back to auto scaling", parNBins, parXMin, parXMax);
  } else {
    nbins = std::stoi(itNBins->second);
    xmin = std::stof(itXMin->second);
    xmax = std::stof(itXMax->second);
```
Please also extend the warning message about how to fix this in the json file

wiechula (changes requested):

Modules/TPC/run/tpcQCClusters_direct.json
```diff
@@ -28,31 +28,23 @@
         "className": "o2::quality_control_modules::tpc::Clusters",
         "moduleName": "QcTPC",
         "detectorName": "TPC",
-        "cycleDurationSeconds": "60",
+        "cycleDurationSeconds": "10",
```
Barthelemy, line 31: Are you sure about this one ? publishing every 10 seconds is quite a lot.

Barthelemy (commented):
Good for me, I would just ask to reconsider the monitor cycle of 10 seconds.

Modules/TPC/run/tpcQCClusters_direct.json
tklemenz: Yes, this is for testing only anyway so it doesn't really matter.

Barthelemy: ok, thank you

Barthelemy (approved):

PR #769, opened 2021-07-15
More debug info for CCDBDatabaseTest

knopers8:


knopers8:
Hi @TimoWilken, @ktf, there are no logs on `build/QualityControl/o2 `, could you have a look please?

TimoWilken:
That's a very strange error:
```
++ source /mnt/mesos/sandbox/sandbox/qualitycontrol/sw/slc7_x86-64/alibuild-recipe-tools/0.2.2-1/etc/profile.d/init.sh
/mnt/mesos/sandbox/sandbox/qualitycontrol/sw/slc7_x86-64/AliEn-CAs/v1-30/etc/profile.d/init.sh: line 2: /mnt/mesos/sandbox/sandbox/qualitycontrol/sw/slc7_x86-64/alibuild-recipe-tools/0.2.2-1/etc/profile.d/init.sh: No such file or directory
```
What makes it weirder is that if I fetch the tarball for AliEn-CAs v1-30 from the S3 remote, it correctly contains the line
```bash
[ -z ${ALIBUILD_RECIPE_TOOLS_REVISION+x} ] && source ${WORK_DIR}/${ALIBUILD_ARCH_PREFIX}/alibuild-recipe-tools/0.2.2-5/etc/profile.d/init.sh
```
(Note the `-5` revision instead of `-1` as above!)

I've fetched the correct tarball from S3 onto the one in the builder's store and deleted its `sw/slc7_x86-64/AliEn-CAs/`. Hopefully that'll fix it. I'm not sure how this could happen, though.

knopers8:
Thanks!

PR #771, opened 2021-07-19
[QC-619] Force the facility name to be less than 32 characters

Barthelemy:


Framework/include/QualityControl/runnerUtils.h
```diff
@@ -81,6 +81,20 @@ inline int computeRunNumber(const framework::ServiceRegistry& services, const bo
   return run;
 }
 
+/**
+ * If the string is longer than the requested length, truncate it.
+ * @param toBeTruncated
+ * @return toBeTruncated truncated to `length` characters
+ */
+inline std::string truncate(std::string toBeTruncated, size_t length)
```
knopers8, line 89: Doesn't `string::substr()` do exactly that? It should survive length() > length

knopers8 (approved):

Framework/include/QualityControl/runnerUtils.h
Barthelemy: I thought not but let me give it a try !

knopers8: Returns a substring [pos, pos+count). If the requested substring extends past the end of the string, or if count == npos, the returned substring is [pos, size()). 
from https://en.cppreference.com/w/cpp/string/basic_string/substr

Barthelemy: Dammit, you were right !
Fixed;

PR #772, opened 2021-07-20
fix the version name in the last step

Barthelemy:


knopers8 (approved):
Ah yes, I've seen this but I thought i did something wrong.

PR #774, opened 2021-07-20
[QC-623] Repo cleaner in parallel

Barthelemy:


Barthelemy:
@TimoWilken I am not sure but I am wondering if there is a problem with the check QualityControl/o2. 

The timeout seems to affect stuff connecting to ccdb-test. It does not happen on other checks and platforms.

Barthelemy:
error unrealted, merging.

PR #777, opened 2021-07-22
Tune the tests timeouts

Barthelemy:


Barthelemy:
The mac build error seems unrelated.

PR #781, opened 2021-07-23
[EMCAL-524] fix binnings for ampl. and time histos

cterrevo:


mfasDa (approved):
Looks good. Should be merged. We use the actual amount of bins in order to reduce the in-memory size a bit. Local tests with the profiler however indicate that we need to drastically reduce the size of the output objects as we spend 99% of the time the task consums in sending the histograms via shmem transport.

Barthelemy:
error unrelated, merging.

Barthelemy:
@TimoWilken Could you check why  build/QualityControl/o2  failed ?

PR #785, opened 2021-07-26
[EMCAL-704] Add suport of cell container in multiple subtimeframes

mfasDa:
In case the reconstruction runs on the FLP the cells
might come from two different cell containers from
the two subtimeframes. In this case the cell containers
might not be merged, but instead the TriggerRecords
are merged to have two different RangeReferences
with a subspecification, called Subevent. The event
loop iterates over the CombinedEvent containing
the interaction record, the trigger type and the list of
subevents. In order to identify the cell container based
on the subspecification a map is built at the beginning
of the monitorData function.

### DIFF
--- a/Modules/EMCAL/src/DigitsQcTask.cxx
+++ b/Modules/EMCAL/src/DigitsQcTask.cxx
@@ -27,7 +27,11 @@
 #include "EMCALBase/Geometry.h"
 #include "EMCALCalib/BadChannelMap.h"
 #include "EMCALCalib/TimeCalibrationParams.h"
+#include <Framework/ConcreteDataMatcher.h>
+#include <Framework/DataRefUtils.h>
+#include <Framework/DataSpecUtils.h>
 #include <Framework/InputRecord.h>
+#include <Framework/InputRecordWalker.h>
 #include <CommonConstants/Triggers.h>
 
 namespace o2
@@ -98,22 +102,36 @@ void DigitsQcTask::monitorData(o2::framework::ProcessingContext& ctx)
     }
   }
 
-  // Get payload and loop over digits
-  auto digitcontainer = ctx.inputs().get<gsl::span<o2::emcal::Cell>>("emcal-digits"); //it was emcal::Digit
-  auto triggerrecords = ctx.inputs().get<gsl::span<o2::emcal::TriggerRecord>>("emcal-triggerecords");
+  // Handling of inputs from multiple subevents (multiple FLPs)
+  // Build maps of trigger records and cells according to the subspecification
+  // and combine trigger records from different maps into a single map of range
+  // references and subspecifications
+  std::vector<framework::InputSpec> cellInputs{ { "cellfilter", framework::ConcreteDataTypeMatcher(header::gDataOriginEMC, "CELLS") } },
+    triggerRecordInputs{ { "triggerrecordfilter", framework::ConcreteDataTypeMatcher(header::gDataOriginEMC, "CELLSTRGR") } };
+  std::unordered_map<int, gsl::span<const o2::emcal::Cell>> cellSubEvents;
+  std::unordered_map<int, gsl::span<const o2::emcal::TriggerRecord>> triggerRecordSubevents;
+
+  for (const auto& celldata : framework::InputRecordWalker(ctx.inputs(), cellInputs)) {
+    int subspecification = framework::DataRefUtils::getHeader<header::DataHeader*>(celldata)->subSpecification;
+    cellSubEvents[subspecification] = ctx.inputs().get<gsl::span<o2::emcal::Cell>>(celldata);
+  }
+  for (const auto& trgrecorddata : framework::InputRecordWalker(ctx.inputs(), triggerRecordInputs)) {
+    int subspecification = framework::DataRefUtils::getHeader<header::DataHeader*>(trgrecorddata)->subSpecification;
+    triggerRecordSubevents[subspecification] = ctx.inputs().get<gsl::span<o2::emcal::TriggerRecord>>(trgrecorddata);
+  }
+
+  auto combinedEvents = buildCombinedEvents(triggerRecordSubevents);
 
   //  QcInfoLogger::GetInstance() << "Received " << digitcontainer.size() << " digits " << AliceO2::InfoLogger::InfoLogger::endm;
   int eventcounter = 0;
-  for (auto trg : triggerrecords) {
+  for (auto trg : combinedEvents) {
     if (!trg.getNumberOfObjects()) {
       continue;
     }
-
     QcInfoLogger::GetInstance() << QcInfoLogger::Debug << "Next event " << eventcounter << " has " << trg.getNumberOfObjects() << " digits" << QcInfoLogger::endm;
-    gsl::span<const o2::emcal::Cell> eventdigits(digitcontainer.data() + trg.getFirstEntry(), trg.getNumberOfObjects());
 
     //trigger type
-    auto triggertype = trg.getTriggerBits();
+    auto triggertype = trg.mTriggerType;
     bool isPhysTrigger = triggertype & o2::trigger::PhT, isCalibTrigger = triggertype & o2::trigger::Cal;
     std::string trgClass;
     if (isPhysTrigger) {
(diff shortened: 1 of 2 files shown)


mfasDa:
Implementation according to suggestion in https://github.com/AliceO2Group/AliceO2/pull/6725

Modules/EMCAL/src/DigitsQcTask.cxx
```diff
@@ -98,22 +102,46 @@ void DigitsQcTask::monitorData(o2::framework::ProcessingContext& ctx)
...
+
+  for (const auto& triggerrecdata : framework::InputRecordWalker(ctx.inputs(), triggerRecordInputs)) {
+    int ntriggerrecords = framework::DataRefUtils::getPayloadSize(triggerrecdata) / sizeof(o2::emcal::TriggerRecord);
+    int subspecification = 0;
+    auto subspecval = framework::DataSpecUtils::getOptionalSubSpec(*triggerrecdata.spec);
+    if (subspecval.has_value())
+      subspecification = *subspecval;
+    auto data = gsl::span<const o2::emcal::TriggerRecord>(reinterpret_cast<o2::emcal::TriggerRecord const*>(triggerrecdata.payload), ntriggerrecords);
+    triggerRecordSubevents[subspecification] = data;
+  }
```
shahor02, line 131: I think the following should work:
````
for (const auto& celldata : framework::InputRecordWalker(ctx.inputs(), cellInputs)) {
  int subspecification = framework::DataRefUtils::getHeader<header::DataHeader*>(ref)->subSpecification;
  cellSubEvents[subspecification] = ctx.inputs().get<gsl::span<const o2::emcal::Cell>>(celldata);
}
for (const auto& celldata : framework::InputRecordWalker(ctx.inputs(), triggerRecordInputs)) {
  int subspecification = framework::DataRefUtils::getHeader<header::DataHeader*>(ref)->subSpecification;
  triggerRecordSubevents[subspecification] = ctx.inputs().get<gsl::span<const o2::emcal::TriggerRecord>>(celldata);
}
````

BTW, `unordered_map` should work faster, unless you really need sorting over subspecs.

Comment (mfasDa) [reply Modules/EMCAL/src/DigitsQcTask.cxx:131]: Hi @shahor02,

Thanks for your suggestions! I will implement them accordingly. Sure, I don't need the subspecification to be sorted, unordered_map is sufficient.

One question: In the line getting the subspecification the header is taken from "ref", what is ref in this case? Where do I get the header from? I would guess the  iterator celldata - is this correct?

Comment (shahor02) [reply Modules/EMCAL/src/DigitsQcTask.cxx:131]: Right, it should be the iterator, i.e. the `celldata`, sorry, was copy/pasting from other code.

Comment (mfasDa) [reply Modules/EMCAL/src/DigitsQcTask.cxx:131]: Still the code does not work because ctx.inputs().get returns std::unique_ptr which I think I cannot pass that easy to a map. I need to copy the gsl::span before adding to the unordered_map, but as it is only a reference over a container the copy should be a fast operation

**Comment by mfasDa** [comment]
A further remark: With the commit I also disable two histograms (cell time vs. abs. ID) from sending for two reasons:

- We run into crashes when sending the histograms with the message that we exceed the shared memory buffer offline (more than 1.5 GB). Interesting that in run2 for the HLT this was no problem - we had the same histograms there  - while for run3 QC it is
- When running the time profiler we see that we currently spend 99.9% in streaming, sending and postprocessing of the output and less than 0.1% in monitorData.

Ideally we would like to put the histograms back, for which we need to check internally how we can reduce the size and still have meaningful output, but in order to do further commissioning tests for the moment we need to disable them. Also I am wondering whether other detectors don't have similar needs to monitor stuff per channel, and all barrel detectors should have far more channels than the calorimeters.

File: Modules/EMCAL/src/DigitsQcTask.cxx (hunk 1 above)
Comment (shahor02) [reply Modules/EMCAL/src/DigitsQcTask.cxx]: @mfasDa The `ctx.inputs().get<gsl::span<const o2::emcal::Cell>>(celldata)` should return a `gsl::span<const o2::emcal::Cell>`, see https://github.com/AliceO2Group/AliceO2/tree/dev/Framework/Core#creating-outputs---the-dataallocator-api
Do you get problem in compilation?

Comment (mfasDa) [reply Modules/EMCAL/src/DigitsQcTask.cxx]:
```
2021-07-27@10:44:34:DEBUG:QualityControl:QualityControl:aliceo2: /Users/markusfasel/alice/sw/SOURCES/QualityControl/aliceo2/0/Modules/EMCAL/src/DigitsQcTask.cxx:125:46: error: no viable overloaded '='
2021-07-27@10:44:34:DEBUG:QualityControl:QualityControl:aliceo2:     triggerRecordSubevents[subspecification] = ctx.inputs().get<gsl::span<const o2::emcal::TriggerRecord>>(trgrecorddata);
2021-07-27@10:44:34:DEBUG:QualityControl:QualityControl:aliceo2:     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ ^ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
2021-07-27@10:44:34:DEBUG:QualityControl:QualityControl:aliceo2: /Users/markusfasel/alice/sw/osx_x86-64/ms_gsl/3.1.0-local1/include/gsl/span:539:21: note: candidate function not viable: no known conversion from 'std::__1::unique_ptr<const gsl::span<const o2::emcal::TriggerRecord, 18446744073709551615>, o2::framework::InputRecord::Deleter<const gsl::span<const o2::emcal::TriggerRecord, 18446744073709551615>>>' to 'const gsl::span<const o2::emcal::TriggerRecord, 18446744073709551615>' for 1st argument
2021-07-27@10:44:34:DEBUG:QualityControl:QualityControl:aliceo2:     constexpr span& operator=(const span& other) noexcept = default;
2021-07-27@10:44:34:DEBUG:QualityControl:QualityControl:aliceo2:                     ^
2021-07-27@10:44:34:DEBUG:QualityControl:QualityControl:aliceo2: 1 error generated.
```

shahor02:
@mfasDa 1.5 GB histos looks an overkill and sending root objects goes via serialization hence expensice. Are these huge histos used for analysis or visualization? I don't see point in visualization of histo with such amount of channels, can't you process it to lower number of bins or even dimension?
Note that there is a simple messeagable histo class https://github.com/AliceO2Group/AliceO2/blob/dev/DataFormats/common/include/CommonDataFormat/FlatHisto2D.h and https://github.com/AliceO2Group/AliceO2/blob/dev/DataFormats/common/include/CommonDataFormat/FlatHisto1D.h

Modules/EMCAL/src/DigitsQcTask.cxx
shahor02: please use `cellSubEvents[subspecification] = ctx.inputs().get<gsl::span<o2::emcal::Cell>>(celldata);` (instead of `span<const gsl::span<o2::emcal::Cell>`, the latter will be returned)

mfasDa:
@shahor02 The main reason for having such a granularity online is basically to find frontend cards which loose configuration, this happened from time to time in run2, they usually showed up in the time distribution loosing the main peak but instead showing distinct time peak at periodic time, something that can be caught in a checker component. Once the on-call knows about them (card ID) they can be reconfigured via the DCS and don't need to be masked. For the purpose we are internally discussing whether we can reduce the granularity from cellID to FEC ID - this would mean ~40 cards / supermodule - as usually all cells in such frontend cards behave similarly. Still we need a sufficient time resolution to distinguish the pattern.

Modules/EMCAL/src/DigitsQcTask.cxx
mfasDa: Not the span was const but the EMCAL Cell, but indeed removing the const from the Cell solves the problem. I will however keep the cells constant within the task as they should be read-only.

shahor02: I meant `ctx.inputs().get<gsl::span<X>>(..)` returns `gsl::span<const X>`, so the LHS should be of that type.

knopers8 (approved):

PR #790, opened 2021-07-29
[EMCAL-524] removing Low Gain histos

cterrevo:


mfasDa (approved):
We discussed internally that we drop the low gain digits for the moment (high energy). This will reduce the amount of 2D histograms by factor 2. At least offline we see with this that we no longer saturate the shared memory buffer. With this we could give it another try on the EPN+merger. In case it is still not sufficient with having the high-gain only there are some more options in reducing the range and consequently also the amount of bins in the observable direction.

knopers8 (approved):

PR #792, opened 2021-07-31
[EMCAL-524] Make histograms vs cell ID optional

mfasDa:
- Move filling of all hisstograms to a separate function
  fillHistograms of DigitsHistograms, this makes
  handling of optional histograms a lot easier
- Use lamdba functions for fill/reset/delete of
  optional histograms
- Disable all 2D histograms except occupancy for
  the LED events
- Add functions for easy access to task parameters

### DIFF
--- a/Modules/EMCAL/include/EMCAL/DigitsQcTask.h
+++ b/Modules/EMCAL/include/EMCAL/DigitsQcTask.h
@@ -20,6 +20,7 @@
 #include "QualityControl/TaskInterface.h"
 #include <array>
 #include <unordered_map>
+#include <string_view>
 #include <gsl/span>
 #include <CCDB/TObjectWrapper.h>
 #include <TProfile2D.h>
@@ -40,6 +41,7 @@ namespace emcal
 class Geometry;
 class BadChannelMap;
 class TimeCalibrationParams;
+class Cell;
 } // namespace emcal
 
 namespace quality_control_modules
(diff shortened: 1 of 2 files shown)


Modules/EMCAL/include/EMCAL/DigitsQcTask.h
```diff
@@ -95,9 +106,12 @@ class DigitsQcTask final : public TaskInterface
   void endOfActivity(Activity& activity) override;
   void reset() override;
 
-  void setThreshold(Double_t threshold) { mCellThreshold = threshold; }
   void setEndOfPayloadCheck(Bool_t doCheck) { mDoEndOfPayloadCheck = doCheck; }
 
+  bool hasConfigValue(const std::string_view key);
+  std::string getConfigValue(const std::string_view key);
+  std::string getConfigValueLower(const std::string_view key);
```
mfasDa, line 113: @knopers8 @Barthelemy Could those functions be pushed upwards to the TaskInterface?

Actually I am thinking about something similar to what is done for the Options by the framework having variant data types for int, float, bool and string, and a template function casting to the required type. (Would remove the need for the lambdas I am using to extract bool and float from the string myself)

mfasDa:
Tests work fine - @knopers8 @Barthelemy please merge when the build checks are done

Modules/EMCAL/include/EMCAL/DigitsQcTask.h
knopers8: I created a ticket: https://alice.its.cern.ch/jira/browse/QC-632

Modules/EMCAL/src/DigitsQcTask.cxx
```diff
@@ -369,35 +493,25 @@ void DigitsQcTask::DigitsHistograms::clean()
...
-    delete mIntegratedOccupancy;
-
-  if (mnumberEvents)
-    delete mnumberEvents;
+  auto cleanOptional = [](TObject* hist) {
+    if (hist)
+      delete hist;
+  };
+
+  cleanOptional(mDigitTime);
```
knopers8, line 501: FYI, deleting nullptr has no effect:
https://stackoverflow.com/questions/6731331/is-it-still-safe-to-delete-nullptr-in-c0x

knopers8 (approved):

PR #793, opened 2021-08-02
Implementation of TOF matching QC

chiarazampolli:
intermediate commit

TOF Matching QC for {ITS-TPC} or {ITS-TPC and TPC}

clang-format + remove unused variable

### DIFF
--- a/Modules/TOF/include/TOF/TOFMatchedTracks.h
+++ b/Modules/TOF/include/TOF/TOFMatchedTracks.h
@@ -22,14 +22,14 @@
 #include "DataFormatsGlobalTracking/RecoContainer.h"
 #include "ReconstructionDataFormats/MatchInfoTOFReco.h"
 
-class TH1D;
+class TH1F;
+
+namespace o2::quality_control_modules::tof
+{
 
 using namespace o2::quality_control::core;
 using GID = o2::dataformats::GlobalTrackID;
 using trkType = o2::dataformats::MatchInfoTOFReco::TrackType;
-
-namespace o2::quality_control_modules::tof
-{
 
 /// \brief Example Quality Control DPL Task
 /// \author My Name
@@ -64,12 +64,12 @@ class TOFMatchedTracks final : public TaskInterface
 
   bool mUseMC = false;
   bool mVerbose = false;
-  TH1D* mInTracksPt[trkType::SIZE];
-  TH1D* mInTracksEta[trkType::SIZE];
-  TH1D* mMatchedTracksPt[trkType::SIZE];
-  TH1D* mMatchedTracksEta[trkType::SIZE];
-  TH1D* mEffPt[trkType::SIZE];
-  TH1D* mEffEta[trkType::SIZE];
+  TH1F* mInTracksPt[trkType::SIZE] = {};
+  TH1F* mInTracksEta[trkType::SIZE] = {};
+  TH1F* mMatchedTracksPt[trkType::SIZE] = {};
+  TH1F* mMatchedTracksEta[trkType::SIZE] = {};
+  TH1F* mEffPt[trkType::SIZE] = {};
+  TH1F* mEffEta[trkType::SIZE] = {};
 };
 
 } // namespace o2::quality_control_modules::tof
--- a/Modules/TOF/include/TOF/TOFMatchedTracks.h
+++ b/Modules/TOF/include/TOF/TOFMatchedTracks.h
@@ -1,0 +1,77 @@
+// Copyright 2019-2020 CERN and copyright holders of ALICE O2.
+// See https://alice-o2.web.cern.ch/copyright for details of the copyright holders.
+// All rights not expressly granted are reserved.
+//
+// This software is distributed under the terms of the GNU General Public
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+///
+/// \file   TOFMatchedTracks.h
+/// \author My Name
+///
+
+#ifndef QC_MODULE_TOF_TOFTOFMATCHEDTRACKS_H
+#define QC_MODULE_TOF_TOFTOFMATCHEDTRACKS_H
+
+#include "QualityControl/TaskInterface.h"
+
+#include "DataFormatsGlobalTracking/RecoContainer.h"
+#include "ReconstructionDataFormats/MatchInfoTOFReco.h"
+
+class TH1D;
+
+using namespace o2::quality_control::core;
+using GID = o2::dataformats::GlobalTrackID;
+using trkType = o2::dataformats::MatchInfoTOFReco::TrackType;
+
+namespace o2::quality_control_modules::tof
+{
+
+/// \brief Example Quality Control DPL Task
+/// \author My Name
+class TOFMatchedTracks final : public TaskInterface
+{
+ public:
+  /// \brief Constructor
+  TOFMatchedTracks() = default;
+  /// Destructor
+  ~TOFMatchedTracks() override;
+
+  // Definition of the methods for the template method pattern
+  void initialize(o2::framework::InitContext& ctx) override;
+  void startOfActivity(Activity& activity) override;
+  void startOfCycle() override;
+  void monitorData(o2::framework::ProcessingContext& ctx) override;
+  void endOfCycle() override;
+  void endOfActivity(Activity& activity) override;
+  void reset() override;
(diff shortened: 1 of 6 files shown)


chiarazampolli:
@noferini , @njacazio , FYI. 

@shahor02 : if you could also take a look, it won't hurt for sure :-) 

Please, note that, unless I misunderstood or mis-implemented something, there seems to be an issue when one relates the GlobalTrackID's to the TOF matching info. This was spotted because in the eta distribution of unconstrained tracks (those matched to TPC only), in my simulation, I got one entry in one bin in the numerator without any corresponding entry in the denominator, which cannot be since the matched tracks are a subsample of the TPC-only ones. This should be checked in O2, in the MatchTOF class. As a consequence, the efficiency plots that result from this task might be uncorrect.

As discussed, the MC information is at present not used.

njacazio (approved):

chiarazampolli:
Hello @TimoWilken , 

Was the build killed due to memory? Can you check?

Thanks, 

Chiara

TimoWilken:
@chiarazampolli, I think it was killed due to memory, but it's been rebuilt now (and this time, it succeeded). I could reduce the number of jobs to prevent OOMs, but I'd rather avoid it until this becomes a problem as it slows that check down.

chiarazampolli:
Thanks @TimoWilken !

Modules/TOF/include/TOF/TOFMatchedTracks.h
```diff
@@ -0,0 +1,77 @@
...
+  gsl::span<const o2::dataformats::MatchInfoTOF> mITSTPCTOFMatches;
+
+  bool mUseMC = false;
+  bool mVerbose = false;
+  TH1D* mInTracksPt[trkType::SIZE];
+  TH1D* mInTracksEta[trkType::SIZE];
+  TH1D* mMatchedTracksPt[trkType::SIZE];
+  TH1D* mMatchedTracksEta[trkType::SIZE];
+  TH1D* mEffPt[trkType::SIZE];
+  TH1D* mEffEta[trkType::SIZE];
```
shahor02, line 72: it is safer to init them with {}, in the destructor you are deleting them all in loop.

Modules/TOF/src/TOFMatchedTracks.cxx
```diff
@@ -0,0 +1,242 @@
...
+  if ((mSrc[GID::Source::TPCTOF] == 1 && mSrc[GID::Source::TPC] == 0) || (mSrc[GID::Source::TPCTOF] == 0 && mSrc[GID::Source::TPC] == 1)) {
+    ILOG(Fatal, Support) << "Check the requested sources: TPCTOF = " << mSrc[GID::Source::TPCTOF] << ", TPC = " << mSrc[GID::Source::TPC] << ENDM;
+  }
+
+  if ((mSrc[GID::Source::ITSTPCTOF] == 1 && mSrc[GID::Source::ITSTPC] == 0) || (mSrc[GID::Source::ITSTPCTOF] == 0 && mSrc[GID::Source::ITSTPC] == 1)) {
+    ILOG(Fatal, Support) << "Check the requested sources: ITSTPCTOF = " << mSrc[GID::Source::ITSTPCTOF] << ", ITSTPC = " << mSrc[GID::Source::ITSTPC] << ENDM;
+  }
+
+  for (int i = 0; i < trkType::SIZE; ++i) {
+    mInTracksPt[i] = new TH1D(Form("mInTracksPt_%d", i), Form("mInTracksPt (trkType %d); Pt; counts", i), 100, 0.f, 20.f);
```
shahor02, line 80: do you really need TH1D?

Modules/TOF/src/TOFMatchedTracks.cxx
```diff
@@ -0,0 +1,242 @@
...
+        }
+      }
+      this->mInTracksPt[trkType::UNCONS]->Fill(trk.getPt());
+      this->mInTracksEta[trkType::UNCONS]->Fill(trk.getEta());
+    }
+    // In case of ITS-TPC-TOF, the ITS-TPC tracks contain also the ITS-TPC-TOF
+    if constexpr (isTPCITSTrack<decltype(trk)>()) {
+      this->mInTracksPt[trkType::CONSTR]->Fill(trk.getPt());
+      this->mInTracksEta[trkType::CONSTR]->Fill(trk.getEta());
+    }
```
shahor02, line 171: since you account the reference "pure" TPC tracks with some selections, for the consistency, the same cuts should be applied to the TPC tracks used in the matches.
This can access it via: 
````
auto tpcTr = this->mRecoCont.getTPCTrack( this->mRecoCont.getTPCContributorGID( GID)  );
````

File: Modules/TOF/src/TOFMatchedTracks.cxx (hunk 2 above)
Comment (chiarazampolli) [reply Modules/TOF/src/TOFMatchedTracks.cxx:80]: I changed it, Initially it was TH1F, then I moved to TH1D thinking that maybe we want to integrate over longer times, but this is then a different way to implement, in the merger (which I have not yet tried myself). So let's use TH1F to start with, it makes sense.

File: Modules/TOF/include/TOF/TOFMatchedTracks.h (hunk 1 above)
Comment (chiarazampolli) [reply Modules/TOF/include/TOF/TOFMatchedTracks.h:72]: done

File: Modules/TOF/include/TOF/TOFMatchedTracks.h [hunk 4]
```diff
@@ -0,0 +1,77 @@
...
+#include "QualityControl/TaskInterface.h"
+
+#include "DataFormatsGlobalTracking/RecoContainer.h"
+#include "ReconstructionDataFormats/MatchInfoTOFReco.h"
+
+class TH1D;
+
+using namespace o2::quality_control::core;
+using GID = o2::dataformats::GlobalTrackID;
+using trkType = o2::dataformats::MatchInfoTOFReco::TrackType;
```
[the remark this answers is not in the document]
chiarazampolli: done

Modules/TOF/src/TOFMatchedTracks.cxx
chiarazampolli: I am not sure I understand where you want to use this line. Maybe this can be taken care of by @njacazio / @noferini while I am away? Since it does not sound like a critical comment, if this PR is merged (I think we need @Barthelemy  or @knopers8 for this), they could do the change you propose in another one. If it is not merged, they can open a PR on my branch.

shahor02: I mean: if you want to normalize to TPC track passing certain cuts, this should be done for all TPC tracks, not only this which failed matching. So, one should get the TPC tracks contributing to TOF-matched tracks (with the line above) and accept the match only if it passes the same cuts. Sure, it can be done later, but I cannot merge in QC...

chiarazampolli: Ah, yes, sure. But the part of the selections is not really implemented now, it was only to show how to access the track in the case of TPCTOF. 
I think that only @Barthelemy  and probably @knopers8 can merge...

Modules/TOF/include/TOF/TOFMatchedTracks.h
```diff
@@ -0,0 +1,77 @@
...
+// This software is distributed under the terms of the GNU General Public
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+///
+/// \file   TOFMatchedTracks.h
+/// \author My Name
```
knopers8, line 14: Please put your name.

Modules/TOF/include/TOF/TOFMatchedTracks.h
```diff
@@ -0,0 +1,77 @@
...
+
+namespace o2::quality_control_modules::tof
+{
+
+using namespace o2::quality_control::core;
+using GID = o2::dataformats::GlobalTrackID;
+using trkType = o2::dataformats::MatchInfoTOFReco::TrackType;
+
+/// \brief Example Quality Control DPL Task
+/// \author My Name
```
knopers8, line 35: please put your name

Modules/TOF/src/TOFMatchedTracks.cxx
njacazio: Hi @chiarazampolli I will handle it

knopers8 (approved):
There are two trivial documentation fixes needed, but they can wait. Merging, since @chiarazampolli  is anyway on holidays.

njacazio:
Hi @knopers8 I will handle the changes in the documentation thanks a lot!

PR #796, opened 2021-08-05
MFT: Update the QCDB url from test to production.

tomas-herman:


Barthelemy (approved):

Barthelemy:
are you sure that you want to point to production database from a config file that will most likely be used outside P2 ? I am not sure we want to have all kind of test objects in the QCDB.

tomas-herman:
Hi Barth, we have started to test the MFT cluster QC task on EPN and from our understanding the json file used for this should be the one present in O2 and therefore it must contain the path to the production database. And because we are not allowed to make changes to existing builds of O2 on the EPN we are using a local copy of the json file located on the EPN at the moment. 

I completely agree that for tests outside P2 we don't want to fill the production QCDB and the plan for that is that we would by hand change the URL in our local builds were we are allowed to make changes.

Please correct me if what I said was wrong or if there is a better solution to our situations.

Barthelemy:
it all makes sense, thank you. Merging now.

PR #798, opened 2021-08-06
[QC-627] Set proper ownership of MO collection after deserialization

knopers8:
This fixes the leaks reported in QC-627. The problem was that `MonitorObjects` in `MonitorObjectCollection` were indeed deleted, but they did not own the contained objects, so those were not.

As far as I could see, ROOT does not offer any method which allows to correct ownership after deserialization, so I added one in https://github.com/AliceO2Group/AliceO2/pull/6811 and made sure it is executed after the received objects are unpacked.
In `CheckRunner` we used to assume that `TObjArray` is always non-owning, I had to fix that.

### DIFF
--- a/Framework/src/MonitorObjectCollection.cxx
+++ b/Framework/src/MonitorObjectCollection.cxx
@@ -15,8 +15,8 @@
 ///
 
 #include "QualityControl/MonitorObjectCollection.h"
-
 #include "QualityControl/MonitorObject.h"
+#include "QualityControl/QcInfoLogger.h"
 
 #include <Mergers/MergerAlgorithm.h>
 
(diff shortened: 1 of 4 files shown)


knopers8:
Needs https://github.com/AliceO2Group/AliceO2/pull/6811

Framework/src/MonitorObjectCollection.cxx
```diff
@@ -51,4 +51,23 @@ void MonitorObjectCollection::merge(mergers::MergeInterface* const other)
...
+    }
+    mo->setIsOwner(true);
+    if (mo->getObject() != nullptr && mo->getObject()->InheritsFrom(TCollection::Class())) {
+      auto collection = dynamic_cast<TCollection*>(mo->getObject());
+      collection->SetOwner(true);
+    }
+  }
+  this->SetOwner(true);
+  delete it;
+}
```
Barthelemy, line 71: ROOT objects ownership at its best.

Barthelemy (approved):

Barthelemy:
test failure unrelated, merging

PR #799, opened 2021-08-06
[QC-630] Pass the host name to Data Sampling

knopers8:
...to allow for creating Data Sampling Policies only on specified machines.

needs https://github.com/AliceO2Group/AliceO2/pull/6812


Barthelemy:
macos failure unrelated merging.

PR #803, opened 2021-08-09
TPC: Update of Tracks task

stheckel:
o Add publishing of ratio plots (neg./pos. tracks)
o Add calling of new function processEndOfCycle

### DIFF
--- a/Modules/TPC/src/Tracks.cxx
+++ b/Modules/TPC/src/Tracks.cxx
@@ -49,6 +49,11 @@ void Tracks::initialize(o2::framework::InitContext& /*ctx*/)
     getObjectsManager()->startPublishing(&hist2);
     getObjectsManager()->addMetadata(hist2.GetName(), "custom", "42");
   }
+
+  for (auto& histR : mQCTracks.getHistogramRatios1D()) {
+    getObjectsManager()->startPublishing(&histR);
+    getObjectsManager()->addMetadata(histR.GetName(), "custom", "51");
+  }
 }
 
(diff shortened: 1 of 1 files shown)


stheckel:
This commits needs the following O2 PR to be merged first:
[https://github.com/AliceO2Group/AliceO2/pull/6824](https://github.com/AliceO2Group/AliceO2/pull/6824)

Modules/TPC/src/Tracks.cxx
```diff
@@ -49,6 +49,11 @@ void Tracks::initialize(o2::framework::InitContext& /*ctx*/)
     getObjectsManager()->startPublishing(&hist2);
     getObjectsManager()->addMetadata(hist2.GetName(), "custom", "42");
   }
+
+  for (auto& histR : mQCTracks.getHistogramRatios1D()) {
+    getObjectsManager()->startPublishing(&histR);
+    getObjectsManager()->addMetadata(histR.GetName(), "custom", "51");
```
wiechula, line 55: Isn't this a leftover from some example? I guess it is not needed. Perhaps this should also be removed everywhere else.
```suggestion
```

wiechula (changes requested):

Modules/TPC/src/Tracks.cxx
stheckel: Yes, I agree, this is not needed.

wiechula (approved):

stheckel:
Thanks, Jens, for approving this one and the corresponding O2 commit. The latter is merged, so I remove the draft status here.

PR #804, opened 2021-08-09
[QC-584] Local QC workflows should appears as indentical to AliECS

knopers8:


knopers8:
This will make us regenerate the workflow templates which use mergers.

Barthelemy (approved):

PR #807, opened 2021-08-11
Remove trailing whitespace in codeowners

mfasDa:


mfasDa:
Even though I am listed as codeowner for the EMCAL module I am not recognised as codeowner (i.e. no review assignment). I suspect the issue is the trailing whitespace after my github username in the CODEOWNERS file.

Barthelemy (approved):

Barthelemy:
I also notice that you don't have write access. This might affect the code ownership. I have fixed it too.

PR #814, opened 2021-08-24
[WIP] Add information about period, pass type and provenance to objects

Barthelemy:


Barthelemy:
The information, including the run number, is now set exclusively by the CheckRunner and not the TaskRunner.

Barthelemy:
The information si only retrieved from the config file for the time being as it is not yet passed by the Control.

Barthelemy:
Also the provenance is not yet used in the path.

Barthelemy:
I thought that it would be a trivial PR. It ends up being more massive than anticipated. I tackled the 3 variables at once. I set them on MOs and QOs in CheckRunner. It is also made available to users through the `Activity` object. Documentation has been updated as well as the tests. The provenance has a default of `qc` because it is crucial for the path, whereas the other variables have no default.

PR #815, opened 2021-08-25
Make sure that we stop the driver if we cannot build the workflow

Barthelemy:


Barthelemy:
https://alice-talk.web.cern.ch/t/error-starting-o2-qc/1064/6

knopers8 (approved):
thanks, now i can see the proper exception again.

PR #816, opened 2021-08-25
[WIP] [QC-615] Bind sample channels on remote machines by default

knopers8:
Needs https://github.com/AliceO2Group/AliceO2/pull/6900


knopers8:
@wiechula FYI, this will make the Data Sampling policies bind on the remote QC machines, so it should let you simplify the configuration files.

PR #820, opened 2021-08-28
FIT: FV0 and FDD modules added

afurs:
QC task for digit visualization


Barthelemy (approved):
Thank you for your PR. I don't review the code of your task itself because I am not in a position to judge what makes sense or not. I noted 3 small points: 
1. Dummy checks are used, they could be removed.
2. Some code is commented out. Could it be removed ? 
3. The use of unique_ptr for the plots published is perhaps pointless as we deal with raw pointer within the framework.

afurs:
> Thank you for your PR. I don't review the code of your task itself because I am not in a position to judge what makes sense or not. I noted 3 small points:
> 
>     1. Dummy checks are used, they could be removed.
> 
>     2. Some code is commented out. Could it be removed ?
> 
>     3. The use of unique_ptr for the plots published is perhaps pointless as we deal with raw pointer within the framework.

Hello @Barthelemy , thanks,
This is first step of FV0/FDD QC implementation, actually it was cloned from FT0 and modified. In next PRs I will take into account your points.

2. Some of this code will be "uncommented" after several modifications we planned in digit structure.

PR #823, opened 2021-08-31
[QC-443] Represent Checks configuration as C++ structures

knopers8:
Includes also PostProcessingTaskSpec and ExternalTaskSpec, but these are only
used by CheckRunners so far.

Not the last PR for this issue...

### DIFF
--- a/Framework/include/QualityControl/CheckConfig.h
+++ b/Framework/include/QualityControl/CheckConfig.h
@@ -20,6 +20,9 @@
 #include <string>
 #include <unordered_map>
 
+#include <Framework/DataProcessorSpec.h>
+#include "QualityControl/UpdatePolicyType.h"
+
 namespace o2::quality_control::checker
 {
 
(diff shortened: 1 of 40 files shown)


Framework/include/QualityControl/CheckConfig.h
```diff
@@ -30,9 +33,12 @@ struct CheckConfig {
   std::string className;
   std::string detectorName = "MISC"; // intended to be the 3 letters code;
   std::unordered_map<std::string, std::string> customParameters = {};
-  std::string policyType = "OnAny";
-  std::vector<std::string> objectNames;
+  UpdatePolicyType policyType = UpdatePolicyType::OnAny;
+  std::vector<std::string> objectNames{}; // fixme: if object names are empty, allObjects are true, consider reducing to one var
```
Barthelemy, line 37: good point

Framework/include/QualityControl/runnerUtils.h
```diff
@@ -121,11 +122,37 @@ inline int computeRunNumber(const framework::ServiceRegistry& services, int defa
...
+
+inline std::string computePassName(const std::string& fallbackPassName = "")
+{
+  std::string passName;
+  passName = fallbackPassName;
+  ILOG(Debug, Devel) << "Period Name returned by computePassName : " << passName << ENDM;
+  return passName;
+}
+
+inline std::string computeProvenance(const std::string& fallbackProvenance = "")
```
Barthelemy, line 148: I think that there is a rebase that is missing because these lines were added in another PR.

Barthelemy (approved):

Framework/include/QualityControl/runnerUtils.h
knopers8: i will check

knopers8: I do not understand what is the problem. For now there are two copies of `compute*` functions until we stop using the ones which require the whole ptree (aggregators still use them).

Barthelemy: oh, sorry, I had misunderstood what was going on. All good !

PR #824, opened 2021-09-03
[WIP][QC-429] Indicate the correct validity interval of QC Objects

knopers8:
I still have to find a way to keep "NewObject" post-processing triggers working.


Barthelemy:
Should we close this PR for the time being ?

PR #827, opened 2021-09-08
[DNM] run ci checks

knopers8:


knopers8:
@TimoWilken @ktf @Barthelemy Maybe I am mistaken, but looking at the logs, I don't see any QC tests being run. Am I interpreting the logs incorrectly, or there is some wrong?

TimoWilken:
You're right, the tests aren't run because CI doesn't set `ALIBUILD_O2_TESTS` (and as far as I can tell, it never has for QC). Should this be enabled?

knopers8:
Well, we used to have QC tests running here, something must have changed, but I am not sure what it was.
If setting `ALIBUILD_O2_TESTS` should fix the problem, then indeed we could enable it.

TimoWilken:
I just realised the culprit is alisw/ali-bot@1a604c186e44e202e1f5a3206b94abaf9d40c3e9 -- I didn't realise other projects relied on the same variable. I'll reenable it for all the QC checks.

knopers8:
Thanks!

Barthelemy:
@knopers8 Thank you for spotting this.

PR #828, opened 2021-09-08
Fix the remote machine in channel address of data sampling proxy

knopers8:
The problem crept in in https://github.com/AliceO2Group/QualityControl/pull/823 , but the CI checks did not run any tests, so we did not see the problem. This fixes the failing test on my setup.


knopers8:
Force-merging this, since the tests are broken anyway now and I confirmed it fixes them on my local setup. It will also unblock releasing v1.29

PR #831, opened 2021-09-14
[EMCAL-526] Add configurations for raw QC used at P2

mfasDa:
- readout-stfb.json: Raw QC from STF builder,
  no remote workflow (only 1 FLP)
- readout-stfb-remote.json: Raw QC from STF builder,
  remote workflow on alio2-cr1-qc02


Barthelemy (approved):

Barthelemy:
@TimoWilken @ktf it seems that the mac builder is stuck

PR #832, opened 2021-09-15
Added new trends for the PID of electrons and MIPs

cimordas:
+ Removed old json for the trending for the clusters


wiechula (approved):

wiechula:
@cimordas , if you only want the change the json files I think it would be better to open a new PR with only these changes to have a clean history. Like this it is a bit of a mess.

PR #837, opened 2021-09-20
Do not warn about missing remotePort if running remotely

knopers8:
(DS port is used then)


Barthelemy (approved):

Barthelemy:
force merging as the mac builders are stuck

PR #840, opened 2021-09-23
[EMCAL-526] Refactoring histogram names

mfasDa:
- Consistent naming scheme:
    + Granularity (Bunch/Channel/SM)
    + Acceptance (Full/Subdetector/Link)
    + RC/Spectrum
- Fix reference for getADC in order to
  avoid copying all raw data in-memory
- Discard supermodules with 0/SHRT_MAX
  in filling min/max ADC per supermodule in
  order to skip supermodules on the FLP
  where the detector is not read out

### DIFF
--- a/Modules/EMCAL/include/EMCAL/RawTask.h
+++ b/Modules/EMCAL/include/EMCAL/RawTask.h
@@ -112,37 +112,37 @@ class RawTask final : public TaskInterface
   TH1* mMessageCounter = nullptr;
   TH1* mNumberOfSuperpagesPerMessage;
   TH1* mNumberOfPagesPerMessage;
-  TH1* mSuperpageCounter = nullptr;                                          ///< Counter for number of superpages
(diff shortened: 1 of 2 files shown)


mfasDa:
@Barthelemy @knopers8 Can you please merge this commit? It is basically renaming variables and histogram names (automatically done with the VSCode refactoring tool) in order to have a consistent and meaningful naming scheme as the old one was to some extend confusing hiding the actual purpose of the histograms.

Barthelemy:
> @Barthelemy @knopers8 Can you please merge this commit?

Hi @mfasDa 
Of course we can and will.
The thing is that we don't work at night and we wait for the tests to pass, or not. If not, we assess the situation. Also, we don't have the manpower allowing us to address PRs in the 2 hours after they are created.

mfasDa:
@Barthelemy Thanks a lot!

I guess there was a misunderstanding - of course there was no hurry! I just wanted to notify you because I realized I cannot self-approve and self-merge my own PRs though I have now merge permissions for EMCAL.

PR #845, opened 2021-09-27
[QC-651] Add class to metadata

Barthelemy:


Barthelemy:
I preferred updating all references to the MO constructor to make sure that we specify the class everywhere.

Barthelemy:
re-wip it because there is something wrong

Barthelemy:
ok, 1 commit was not pushed

Modules/ITS/src/TrendingTaskITSThr.cxx
```diff
@@ -78,6 +78,7 @@ void TrendingTaskITSThr::storeTrend(repository::DatabaseInterface& qcdb)
   ILOG(Info, Support) << "Storing the trend, entries: " << mTrend->GetEntries() << ENDM;
 
   auto mo = std::make_shared<core::MonitorObject>(mTrend.get(), getName(),
+                                                  "o2::quality_control_modules::its::TrendingTaskITSThr",
```
knopers8, line 81: Why not `mConfig.className`?

knopers8 (approved):
It looks good, but please check that one suggestion.

Modules/ITS/src/TrendingTaskITSThr.cxx
Barthelemy: yep

PR #846, opened 2021-09-28
[EMCAL-526] Catch mapping error and display in error count histogram

mfasDa:
Handle mapping error (invalid hardware address) and
geometry errors (cell ID outside range) in error type
histogram.


Barthelemy (approved):

mfasDa:
For the record: This handles misconfigured FECS with misconfigured hardware addresses.

PR #847, opened 2021-09-30
TPC: fixed Cluster and RawDigit tasks

tklemenz:
- normalization to number of clusters did not work properly

### DIFF
--- a/Framework/src/CheckRunner.cxx
+++ b/Framework/src/CheckRunner.cxx
@@ -462,5 +462,4 @@ void CheckRunner::reset()
   }
 }
 
-
 } // namespace o2::quality_control::checker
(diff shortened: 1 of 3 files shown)


tklemenz:
Needs https://github.com/AliceO2Group/AliceO2/pull/7168 before merging.

tklemenz:
Why does the check for clang-format fail even though I ran it?
Is `clang-format-6.0 -style=file -i <filename>` wrong now?

wiechula (approved):

Barthelemy:
@tklemenz clang-format does not report problems on my machine either. Some extra space at the end of CheckRunner ? that's real strange.

tklemenz:
Indeed running clang-format on CheckRunner.cxx changed the file. Apparently it did not conform.
Shall I include the file in this PR or do you want to do a separate one?

Barthelemy:
@tklemenz Please include it in this PR even if it is not consistent with the title. I prefer to know that the change is in
The other errors seems legit ? a change is needed in O2 ?

tklemenz:
Yes as mentioned in an earlier comment one PR in O2 is needed. But no circular dependencies this time. :)

Barthelemy (approved):

PR #848, opened 2021-09-30
MCH: improved naming and location of digits task histograms

aferrero2707:
The DE-specific occupancy histograms are moved into sub-folders, following the convention "STx/DExxx/".
This allows to only have global histograms in the root "QcTaskMCHDigits" folder.
Histogram names have also been shortened for better readability.

### DIFF
--- a/Modules/MUON/MCH/src/PhysicsTaskDigits.cxx
+++ b/Modules/MUON/MCH/src/PhysicsTaskDigits.cxx
@@ -49,6 +49,11 @@ PhysicsTaskDigits::PhysicsTaskDigits() : TaskInterface() {}
 
 PhysicsTaskDigits::~PhysicsTaskDigits() {}
 
+static std::string getHistoPath(int deId)
+{
+  return fmt::format("ST{}/DE{}/", (deId - 100) / 200 + 1, deId);
+}
+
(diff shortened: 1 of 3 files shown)


Barthelemy:
@TimoWilken @ktf Macos build failed without logs

Modules/MUON/MCH/src/PhysicsTaskDigits.cxx
```diff
@@ -49,6 +49,11 @@ PhysicsTaskDigits::PhysicsTaskDigits() : TaskInterface() {}
 
 PhysicsTaskDigits::~PhysicsTaskDigits() {}
 
+static std::string getHistoPath(int deId)
+{
+  return fmt::format("ST{}/DE{}/", (deId - 100) / 200 + 1, deId);
```
Barthelemy, line 54: magic numbers ? :)

Barthelemy (approved):

TimoWilken:
> Macos build failed without logs

It was a temporary failure fetching a dependency, I've restarted the check.

Modules/MUON/MCH/src/PhysicsTaskDigits.cxx
aferrero2707: Yes, but graved in the stone ;-)

PR #849, opened 2021-09-30
MFT: Update of cluster task + minor fixes

tomas-herman:
- Adding histograms to the Cluster task
- Fixing a bug in the layout of digit+cluster chip hit maps
- Setting colz option as default
- Disabling dummy checks
- Minor fixes/updates


tomas-herman:
The error produced by build/QualityControl/o2-dataflow/macOS does not seem to be connected to the files modified in the PR.

Barthelemy:
you are correct the mac error is not related.

Barthelemy (approved):
in general we recommend not commenting out but removing the code as it remains in git but you probably have a good reason for it.

tomas-herman:
Thanks. Yes, the commented part is concerning the checker. It does dummy checks and as we don't want the central shifters to be alerted by them I wanted to switch them of. Originally I did this in the json file but people accidentally change this back so I wanted to comment out the checks themselves. The reason not to delete is because the code is good it just needs meaningful conditions/values to check against and this will be done later.

PR #850, opened 2021-09-30
[QC-443] Isolate Aggregators and Post-processing from configuration

knopers8:
Also fixes [QC-653] (aggregators not created in a remote workflow)


Framework/include/QualityControl/AggregatorRunnerConfig.h
```diff
@@ -0,0 +1,40 @@
...
+
+namespace o2::quality_control::checker
+{
+
+struct AggregatorRunnerConfig {
+  std::unordered_map<std::string, std::string> database;
+  std::string consulUrl{};
+  std::string monitoringUrl{};
+  bool infologgerFilterDiscardDebug = false;
+  int infologgerDiscardLevel = 21;
```
Barthelemy, line 31: We could consider having a parent struct containing the common part of all the *Config struct. It is fine like that, in particular for readability though.

Framework/include/QualityControl/PostProcessingTaskSpec.h
```diff
@@ -21,33 +21,26 @@
...
-      detectorName(std::move(detectorName))
+  PostProcessingTaskSpec(std::string taskName)
+    : taskName(std::move(taskName))
   {
   }
 
-  // basic
+  // Post Processing Tasks configure themselves with a ptree.
+  // While this is a lack of consequence, with respect to other *Specs,
+  // I am afraid it is too late to change it, since also users rely on this (see ITS Trending Task).
```
Barthelemy, line 39: agreed

Barthelemy (approved):
very good. It is a lot more uniform and consistent. thanks!

Framework/include/QualityControl/AggregatorRunnerConfig.h
knopers8: Yes, this would be a nice improvement. That being said, I think we should prioritize moving to push config for now, but we could do this in the next iteration.

knopers8:
@Barthelemy was the mac compiler recently updated or the warning flags modified? I would disagree that warnings about unused functions should be treated as errors:
```
error: unused function 'readSpecEntry' [-Werror,-Wunused-function]
```
Especially that they are actually used...

Barthelemy:
@knopers8 I don't think that the mac was updated but I am not sure. I propose to add an exception for unused-function as to not trigger an error.

knopers8:
@cterrevo Coudl you please try again with updated alidist? This is the code change you specifically need https://github.com/alisw/alidist/pull/3383

PR #855, opened 2021-10-04
[QC-648] Move to pub-sub for the channels between machines

knopers8:


ktf:
Can we hack this in at configuration level? Just to verify that it's indeed the issue we are observing.

knopers8:
We can, but that will take more time to implement.

knopers8:
@ktf If you are relying on the DDS topology files, then I suppose you could modify those channel config arguments by hand, that would be the quickest solution for you, I suppose.

ktf:
We are now trying that with @shahor02.

Barthelemy:
@knopers8  thank you for this very fast move. 

@ironMann Please note the comment of Piotr above

Barthelemy (approved):

ironMann:
Hi. A bunch of previously working FLP-QC workflows aren't doing well since the upgrade today. How soon can this be put in place?

Barthelemy:
Is it related ? what are the symptoms ? I would propose to move the discussion to JIRA or by email.

PR #856, opened 2021-10-04
Add missing reference

Barthelemy:
because it can cause issues on some systems


knopers8 (approved):
What kind of issues, out of curiousity?

Barthelemy:
on Mac, with the latest version of the compiler, it creates a warning (that is considered error)

PR #862, opened 2021-10-05
[QC-592] A facility to store ROOT objects from files to QCDB

knopers8:
This makes it possible to take QAResults.root generated by DPL Analysis Tasks and put them to QCDB.


Framework/src/runUploadRootObjects.cxx
```diff
@@ -0,0 +1,146 @@
...
+{
+  size_t objectsUploaded = 0;
+
+  try {
+    bpo::options_description desc{ "Options" };
+    desc.add_options()                                                                                                              //
+      ("help,h", "Help screen")                                                                                                     //
+      ("input-file", bpo::value<std::string>()->default_value("./QAResults.root"), "Path to the ROOT file with objects to insert.") //
+      ("qcdb-url", bpo::value<std::string>()->default_value("ccdb-test.cern.ch:8080"), "URL to the QCDB.")                          //
+      ("task-name", bpo::value<std::string>(), "Name of the task which the objects belong to. Use / to make directories")           //
```
Barthelemy, line 47: Name of the task to which the objects belong. ...

Framework/src/runUploadRootObjects.cxx
```diff
@@ -0,0 +1,146 @@
...
+      ("detector-code", bpo::value<std::string>()->default_value("TST"), "3-letter detector code. Put AOD for analysis tasks")      //
+      ("validity-start", bpo::value<uint64_t>()->default_value(0), "Start of objects validity in ms since epoch")                   //
+      ("validity-end", bpo::value<uint64_t>()->default_value(0), "End of objects validity in ms since epoch")                       //
```
Barthelemy, line 50: do you actually want default values here, in particular 0 ?

Barthelemy (approved):
very good

Framework/src/runUploadRootObjects.cxx
knopers8: For validity and filters: I don't think there is a good way to retrieve those values from analysis tasks, so I would not bother people too much about it before we can offer a solution.
For run number: indeed we could require an explicitly stated value, I will fix this. If it is multi-run, people should be aware of it and put 0 then.

PR #866, opened 2021-10-05
[EMCAL-527] checkers added for shifter hist+fix2D

cterrevo:


mfasDa (approved):
Trivial fix, but needs to be merged.

PR #880, opened 2021-10-07
[OMON-491] re-enable the process monitoring in task and check runners

Barthelemy:


knopers8:
I don't understand. I see that each DPL process reports CPU and memory usage, including QC. Why is this needed?

Barthelemy:
Indeed, I overlooked that.

PR #883, opened 2021-10-08
FT0/FV0: Add BcOrbitMap for each PM module in DigitQcTask

sbysiak:
If still possible we would love to have this on our FLP for pilot beams.
We discovered the need for these plots during yesterday tests.


jotwinow (commented):
@Barthelemy @knopers8 We would appreciate it it can still go to the release for the pilot beam. These are important check for the FIT FEE which we need to perform. Thanks for the understanding.

Barthelemy (approved):

PR #887, opened 2021-10-11
[QC-666] Make sure to have some Facility 

Barthelemy:
even at an early stage in the spawning of the DPL workflow


knopers8 (approved):
This is already an improvement, but perhaps you might also consider a more hacky way in `o2-qc`, which is to try getting the `--id <device-name` argument in `defineDataProcessing`. I am not 100% it is possible though, since it is a framework-related argument.

Barthelemy:
@ktf the mac error was weird, probably a transient issue.

PR #889, opened 2021-10-11
skip wrong RDH in TOF raw QC

noferini:


Modules/TOF/src/TaskRaw.cxx
```diff
@@ -49,6 +49,12 @@ namespace o2::quality_control_modules::tof
 void RawDataDecoder::rdhHandler(const o2::header::RAWDataHeader* rdh)
 {
   //auto orbit = RDHUtils::getHeartBeatOrbit(rdh);
+  bool isValidRDH = RDHUtils::checkRDH(rdh, false);
+
+  if (!isValidRDH) {
+    LOG(ERROR) << "No valid RDH... skipped";
```
njacazio, line 55: Hi @noferini thanks a lot! If this is happening often then we might overflow the message pipeline. Maybe we want to use DEBUG instead?

PR #895, opened 2021-10-12
[QC-669] Tune down ServiceDiscovery logs

Barthelemy:
ServiceDiscovery logs are now warnings and only output every 10 minutes.


knopers8 (approved):
NIce, I was actually wondering if there is a mechanism like this in IL.

Barthelemy:
@MichaelLettrich @TimoWilken The mac build failed without logs.

Merging as the error is unrealted.

PR #896, opened 2021-10-12
[QC-668]  Tune down modules logging.

Barthelemy:
The modules, and in particular the monitorData method, must specify severity and level for their logs.
It is important to allow the proper filtering, at the source or not, of the logs. 

@mfasDa 
@afurs 
@aferrero2707 
@JianLIUhep 
@wiechula 
Could you check my changes ?


ktf:
Most likely v12.

PR #901, opened 2021-10-13
zero energy clusters removed

peressounko:


Barthelemy:
Please, fix the format. Thank you

PR #902, opened 2021-10-13
[QC-671] Set the activity in the MonitorObject earlier.

Barthelemy:
Set the activity in the MonitorObject in the TaskRunner already instead of waiting to store in the CheckRunner.


Barthelemy:
errors unrelated , merging

PR #904, opened 2021-10-13
TOF: extend digit QC

njacazio:
- Extend configuration of checks/tasks
- Use standard types
- Improve performance
- Use std::make_shared
- Using default destructor in TaskDigits
- Add running mode for TaskDigits
- Simplified RawMultiplicity checker
- Improve messages for shifter
- Update TOF jsons

### DIFF
--- a/Modules/TOF/include/Base/MessagePad.h
+++ b/Modules/TOF/include/Base/MessagePad.h
@@ -73,11 +73,11 @@ struct MessagePad {
   }
 
   template <typename T>
-  TPaveText* MakeMessagePad(T* histogram, const Quality& quality)
+  std::shared_ptr<TPaveText> MakeMessagePad(T* histogram, const Quality& quality)
   {
-    TPaveText* msg = new TPaveText(mPadLowX, mPadLowY, mPadHighX, mPadHighY, "blNDC");
+    std::shared_ptr<TPaveText> msg = std::make_shared<TPaveText>(mPadLowX, mPadLowY, mPadHighX, mPadHighY, "blNDC");
     if (mEnabledFlag) {
-      histogram->GetListOfFunctions()->Add(msg);
+      histogram->GetListOfFunctions()->Add(msg.get());
     }
     msg->SetBorderSize(1);
     msg->SetTextColor(kBlack);
--- a/Modules/TOF/src/CheckCompressedData.cxx
+++ b/Modules/TOF/src/CheckCompressedData.cxx
@@ -67,7 +67,7 @@ void CheckCompressedData::beautify(std::shared_ptr<MonitorObject> mo, Quality ch
 {
   if (mo->getName() == "hDiagnostic") {
     auto* h = dynamic_cast<TH2F*>(mo->getObject());
-    TPaveText* msg = mShifterMessages.MakeMessagePad(h, checkResult);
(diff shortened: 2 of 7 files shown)


Modules/TOF/include/Base/MessagePad.h
```diff
@@ -0,0 +1,94 @@
...
+// ROOT includes
+#include "TPaveText.h"
+#include "TList.h"
+#include "TH1F.h"
+#include "TH2F.h"
+
+namespace o2::quality_control_modules::tof
+{
+
+struct MessagePad {
```
ercolessi, line 31: Great idea

njacazio:
@Barthelemy @knopers8 we are ready to have this reviewed and merged! Suggestions are as always more than welcome, note that I added a class to have a message pad in the QCG, maybe this could be a framework feature instead of a TOF one.
Thanks!

Modules/TOF/include/Base/MessagePad.h
```diff
@@ -0,0 +1,107 @@
...
+  float mPadLowX = 0.6;                 /// Position of the message PAD in low x
+  float mPadLowY = 0.5;                 /// Position of the message PAD in low y
+  float mPadHighX = 0.9;                /// Position of the message PAD in high x
+  float mPadHighY = 0.75;               /// Position of the message PAD in high y
+  std::vector<std::string> mMessages{}; /// Message to print
+  int mEnabledFlag = 1;                 /// Flag to enable or disable the pad
+
+  template <typename T>
+  void Configure(const T& CustomParameters)
```
Barthelemy, line 43: We use lower-case initial letter in O2.

Modules/TOF/src/CheckCompressedData.cxx
```diff
@@ -70,34 +67,15 @@ void CheckCompressedData::beautify(std::shared_ptr<MonitorObject> mo, Quality ch
...
     auto* h = dynamic_cast<TH2F*>(mo->getObject());
-    TPaveText* msg = new TPaveText(0.9, 0.1, 1.0, 0.5, "blNDC");
-    h->GetListOfFunctions()->Add(msg);
-    msg->SetBorderSize(1);
-    msg->SetTextColor(kWhite);
-    msg->SetFillColor(kBlack);
-    msg->AddText("Default message for hDiagnostic");
-    msg->SetName(Form("%s_msg", mo->GetName()));
-
+    TPaveText* msg = mShifterMessages.MakeMessagePad(h, checkResult);
```
Barthelemy, line 70: I think that there is a memory leak here. `msg` is not deleted.

Modules/TOF/include/Base/MessagePad.h
```diff
@@ -0,0 +1,107 @@
...
+
+  void AddMessage(const std::string& message)
+  {
+    mMessages.push_back(message);
+  }
+
+  template <typename T>
+  TPaveText* MakeMessagePad(T* histogram, const Quality& quality)
+  {
+    TPaveText* msg = new TPaveText(mPadLowX, mPadLowY, mPadHighX, mPadHighY, "blNDC");
```
Barthelemy, line 78: You could use a smart pointer here to avoid problems with memory leaks.

PR #907, opened 2021-10-14
[MCH] Fixing segmentation fault in digits QCTask 

seperrin:
Removed bad handling of temporary histogram pointer in mergeable object class


Barthelemy (approved):

Barthelemy:
no they are not

PR #911, opened 2021-10-15
[QC-679] Set the run number and the partition in the infologger

Barthelemy:


Framework/src/CheckRunner.cxx
```diff
@@ -423,6 +423,8 @@ void CheckRunner::start(const ServiceRegistry& services)
   mActivity.mPeriodName = computePeriodName(services, mConfig.fallbackPeriodName);
   mActivity.mPassName = computePassName(mConfig.fallbackPassName);
   mActivity.mProvenance = computeProvenance(mConfig.fallbackProvenance);
+  string partitionName = o2::quality_control::core::computePartitionName(services);
```
knopers8, line 426: no need for the namespace

Framework/src/QcInfoLogger.cxx
```diff
@@ -50,11 +50,31 @@ void QcInfoLogger::setDetector(const std::string& detector)
   ILOG(Debug, Support) << "Detector set to " << detector << ENDM;
 }
 
+void QcInfoLogger::setRunPartition(int run, std::string& partitionName)
```
knopers8, line 53: Is there any reason to always set those two in one method, but others separately?

Framework/src/TaskRunner.cxx
```diff
@@ -208,6 +208,8 @@ void TaskRunner::endOfStream(framework::EndOfStreamContext& eosContext)
 void TaskRunner::start(const ServiceRegistry& services)
 {
   mRunNumber = o2::quality_control::core::computeRunNumber(services, mTaskConfig.fallbackRunNumber);
+  string partitionName = o2::quality_control::core::computePartitionName(services);
```
knopers8, line 211: no need for the namespace

PR #913, opened 2021-10-18
Using TProfile for hBusyTime and hEventSize plots

gvolpe79:


Modules/HMPID/src/HmpidTask.cxx
```diff
@@ -125,16 +158,16 @@ void HmpidTask::monitorData(o2::framework::ProcessingContext& ctx)
       if (!mDecoder->decodeBufferFast()) {
         ILOG(Error) << "Error decoding the Superpage !" << ENDM;
```
Barthelemy, line 159: If the task starts receiving bad data it will spam the infologger. Please add at least a level (see suggestion) and if possible use the auto-mute described ([here](https://github.com/AliceO2Group/InfoLogger/blob/master/doc/README.md)).
```suggestion
        ILOG(Error, Devel) << "Error decoding the Superpage !" << ENDM;
```

gvolpe79:
Hi Barth, thanks!  Let me know if something needs to be done on my side.

PR #920, opened 2021-10-21
TOF: Fix out of scope shared ptr

njacazio:


njacazio:
Hi @Barthelemy  I tested it and ran successfully, indeed it should not matter as the object is recreated if it was deleted in the meantime

PR #922, opened 2021-10-21
[QC-682] override stop() but do not register a callback

Barthelemy:


Barthelemy:
actually it does not work if I don't keep the callback assignment in the QC

PR #926, opened 2021-10-22
Trend laser calibration from QCG

cimordas:
This draft depends on this [PR](https://github.com/AliceO2Group/AliceO2/pull/7394) to be finalized and tested with clang


Modules/Common/include/Common/TH1Reductor.h
```diff
@@ -24,7 +24,7 @@ namespace o2::quality_control_modules::common
 /// \brief A Reductor which obtains the most popular characteristics of TH1.
 ///
 /// A Reductor which obtains the most popular characteristics of TH1.
-/// It produces a branch in the format: "mean/D:stddev:entries"
+/// It produces a branchin the format: "mean/D:stddev:entries"
```
wiechula, line 27: ?

Modules/TPC/run/tpcQCTrending_laserCalib.json
```diff
@@ -0,0 +1,118 @@
...
+      },
+      "Activity": {
+        "number": "42",
+        "type": "2"
+      },
+      "monitoring": {
+        "url": "infologger:///debug?qc"
+      },
+      "consul": {
+        "url": "http://consul-test.cern.ch:8500"
```
Barthelemy, line 19: We used to have this url everywhere but it spams the DNS for no useful reason. Could you just use an empty url ? 
Thank you

cimordas: I changed the url to an empty string `""`. Is this right?

wiechula (approved):
A bit ugly, but ok for the moment I think. Actually, it would be nice if the framework would allow for using any type in the trending tree which has a ROOT dictionary, not only basic types.

Modules/Common/include/Common/TH1Reductor.h
cimordas: I think this fix got reverted when I updated my QC yesterday, as I remember fixing it already once. Now it should be fine

Modules/TPC/run/tpcQCTrending_laserCalib.json
Barthelemy: yes that's good

Barthelemy:
@wiechula Do not hesitate to create a ticket in JIRA, we consider all requests !

PR #932, opened 2021-10-27
[QC-681] Use dpl infologger

Barthelemy:


Modules/CMakeLists.txt
```diff
@@ -5,17 +5,17 @@ add_subdirectory(Daq)
...
+#add_subdirectory(ITS)
+#add_subdirectory(MFT)
+#add_subdirectory(PHOS)
+#add_subdirectory(FT0)
+#add_subdirectory(TRD)
+#add_subdirectory(HMPID)
+#add_subdirectory(CPV)
+#add_subdirectory(GLO)
+#add_subdirectory(FV0)
+#add_subdirectory(FDD)
```
knopers8, line 21: Are you sure? :)

knopers8 (changes requested):
Thanks, all seems good, but please do not comment out the detector modules.

Modules/CMakeLists.txt
Barthelemy: I cannot speak freely :D
Thank you for spotting this obvious mistake.

PR #936, opened 2021-11-02
[QC-692] Set the log level earlier

Barthelemy:


Framework/src/runQC.cxx
```diff
@@ -37,6 +37,7 @@ using namespace o2;
 using namespace o2::framework;
 using namespace o2::utilities;
 using namespace o2::configuration;
+using namespace std;
```
knopers8, line 40: you did not use it in the end.

knopers8 (approved):

Framework/src/runQC.cxx
Barthelemy: yes

PR #937, opened 2021-11-02
Cycle must be 10 seconds minimum

Barthelemy:


knopers8:
@Barthelemy Genuine problem with formatting.

Framework/src/InfrastructureSpecReader.cxx
```diff
@@ -87,6 +87,11 @@ TaskSpec InfrastructureSpecReader::readSpecEntry<TaskSpec>(std::string taskName,
   ts.moduleName = taskTree.get<std::string>("moduleName");
   ts.detectorName = taskTree.get<std::string>("detectorName");
   ts.cycleDurationSeconds = taskTree.get<int>("cycleDurationSeconds");
+  if(ts.cycleDurationSeconds < 10) {
+    ILOG(Error, Support) << "Cycle duration is too short (" << ts.cycleDurationSeconds << "), replaced by a duration of 10 seconds." << ENDM;
+    ts.cycleDurationSeconds = 10;
+  }
+
```
knopers8, line 94: I would move validation to `TaskRunnerFactory::extractConfig`, if we ever move to other sources of defining QC workflows (e.g. by specifying `InfrastructureSpec` in C++) then the configuration will be still properly validated.

Barthelemy: good point

knopers8 (approved):

PR #944, opened 2021-11-05
add TOF acceptance cut in matching QC

noferini:
This is to avoid to use tracks outside the TOF acceptance to compute efficiency


noferini:
maxOS error not related to this PR (glew error)

PR #945, opened 2021-11-05
For some logs that are crucial we log to a file.

Barthelemy:


Barthelemy:
Needed because we don't keep the output of the script when ran with the cron. Thus we don't know if it has run and with what config.

Framework/script/RepoCleaner/o2-qc-repo-cleaner
```diff
@@ -223,6 +229,20 @@ def storeMonitoringMetrics(success, duration):
...
 
+def storeCrucialLog(message):
+    """
+    Store few but very important messages to the file ~/repocleaner_logs.txt
+    :param message:
+    :return:
+    """
+    logs_filename = str(Path.home()) + "/repocleaner_logs.txt"   # very limited but important logs
+    try:
+        f = open(logs_filename, "a")
```
knopers8, line 240: Will it behave if we have to repocleaners keeping a file open at the same time? (e.g. one took long enough to be still active once a new instance starts)

knopers8 (approved):

Framework/script/RepoCleaner/o2-qc-repo-cleaner
Barthelemy: The repocleaner checks (in the current version not installed in the CC7 QCDB) that there are no other repocleaner running.

PR #946, opened 2021-11-08
fix for selection and denominator for efficiency

chiarazampolli:
@noferini , FYI.

### DIFF
--- a/Modules/TOF/include/TOF/TOFMatchedTracks.h
+++ b/Modules/TOF/include/TOF/TOFMatchedTracks.h
@@ -94,7 +94,7 @@ class TOFMatchedTracks final : public TaskInterface
 
   // for track selection
   float mPtCut = 0.1f;
-  float mEtaCut = 1.4f;
+  float mEtaCut = 0.8f;
   int32_t mNTPCClustersCut = 40;
   float mDCACut = 100.f;
   float mDCACutY = 10.f;
--- a/Modules/TOF/src/TOFMatchedTracks.cxx
+++ b/Modules/TOF/src/TOFMatchedTracks.cxx
@@ -42,10 +42,14 @@ namespace o2::quality_control_modules::tof
 TOFMatchedTracks::~TOFMatchedTracks()
 {
   for (int i = 0; i < trkType::SIZE; ++i) {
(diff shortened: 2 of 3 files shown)


noferini:
I would change the default of eta cut to 0.8

diff --git a/Modules/TOF/include/TOF/TOFMatchedTracks.h b/Modules/TOF/include/TOF/TOFMatchedTracks.h
index 3cba08e81e..def7a0192e 100644
--- a/Modules/TOF/include/TOF/TOFMatchedTracks.h
+++ b/Modules/TOF/include/TOF/TOFMatchedTracks.h
@@ -94,7 +94,7 @@ class TOFMatchedTracks final : public TaskInterface
 
   // for track selection
   float mPtCut = 0.1f;
-  float mEtaCut = 1.4f;
+  float mEtaCut = 0.8f;
   int32_t mNTPCClustersCut = 40;
   float mDCACut = 100.f;
   float mDCACutY = 10.f;

chiarazampolli:
Ciao @noferini , 

Done!

Chiara

noferini:
if (auto param = mCustomParameters.find("EtaCut"); param != mCustomParameters.end()) {

  if (auto param = mCustomParameters.find("etaCut"); param != mCustomParameters.end()) {

"EtaCut" -> "etaCut" otherwise change it in the json

chiarazampolli:
Good catch, @noferini !

noferini:
Last comment (I hope)
I can run with "ITS-TPC,TPC" json but not with "ITS-TPC" one.
This because now TPC tracks are used also in the case of global track, then also the TPC only track input is required.
ITS-TPC json should be fixed.

chiarazampolli:
Very useful again! Done (I tested it and it worked, but please check as well).

chiarazampolli:
Failure on macos seems not related to the PR, @Barthelemy , can you check?

Barthelemy:
@TimoWilken The macos build is still failing

@chiarazampolli I am force merging

TimoWilken:
@Barthelemy sorry, I missed that the `o2` defaults also override the XRootD version. Fixed in https://github.com/alisw/alidist/pull/3537.

PR #948, opened 2021-11-08
[QC-640] Service discovery dynamic port

Barthelemy:


Barthelemy:
@TimoWilken The macos build fails without errors now.

knopers8 (approved):

TimoWilken:
> The macos build fails without errors now.

That was an error cloning a git repo. Restarted. (And this time it should work, as it's now using the updated XRootD.)

PR #952, opened 2021-11-14
ITS noisypix task without decoding digits

nicolovalle:
No more digits query


iravasen:
Code already tested by @nicolovalle privately. The PR can be merged.

PR #955, opened 2021-11-16
updated MC root file readers: fixed error when reading more than one TF

Katka19:


Katka19:
Hi, to be honest, I am not sure why this small commit is having problems with the checks. If it is a problem on my side, then I will investigate and close the PR.

Barthelemy:
it is unrelated, merging.

PR #956, opened 2021-11-17
TRD noise removal from plots

bazinski:
Remove the noisy MCMs from the tracklets plots and pulseheight plots.
Keep both the noisy and new plots.
The noisy plots might get removed later.
2 Tasks, pulseheight and tracklettask make connection to ccdbapi, this will of course change later.


Barthelemy:
the test failures seem legit.

bazinski:
oui, I am at home sick will get back to this once fully functional.

bazinski:
Hi @Barthelemy 
That should be better, forgot to remove noise.cxx|h from cmake and linkdef.

Barthelemy:
@TimoWilken The two failing tests don't have logs.

Barthelemy (approved):

PR #957, opened 2021-11-18
Fix sporadic

ktf:


knopers8 (approved):
We might have some test checking if Lifetime is Timeframe, we will see.

ktf:
Indeed, I fixed those as well.

davidrohr (approved):
This passes the full system test on the EPN and at least fixes part of the problem with the rate-throttling. I think we can go ahead and merge it, and then start looking for the next issue that breaks the throttling.

ktf:
Merging this  to get it in the nightly.

PR #962, opened 2021-11-22
Fix the help message of o2-qc-metadata-updater

Barthelemy:


Framework/src/runMetadataUpdater.cxx
```diff
@@ -93,4 +93,4 @@ int main(int argc, const char* argv[])
   return 0;
 }
 
-// -0.378347x + 2782.30
\ No newline at end of file
+// -0.378347x + 2782.30
```
knopers8, line 96: What is this secret linear function?

knopers8 (approved):
clang format complains

Framework/src/runMetadataUpdater.cxx
Barthelemy: hmmm, good question. I would blame my IDE but I am not sure. Checking

knopers8: I think it came from here, so I should ask myself that question :D https://github.com/AliceO2Group/QualityControl/blob/7c2048d7a03e8a7c5a640d967e7cd6c5d700cff9/Framework/src/runLocationCalculator.cxx#L125

Barthelemy: it is coming from runLocationCalculator.cxx which I copied. 
Thus, tell me, what is this secret linear function ? ;)

knopers8: Then it might have been a merger performance function wrt the number of inputs. Definitely not needed anywhere anymore :)

PR #966, opened 2021-11-24
TPC: add mergeable output for Clusters task

tklemenz:
The Clusters task now has the option to use mergeable output to be able to run the task locally and do only the merging on the remote node. For the visualization of the merged output the post-processing task `ClusterVisualizer` is needed.

The Clusters task can still be run in the exact same way before if needed.

An exemplary json file for mergeable output in a local Clusters task is `Modules/TPC/run/tpcQCClusters_direct.json`. The `taskParameter` `mergeableOutput` has to be set to true (false is default even if the option is completely omitted).

### DIFF
--- a/Modules/TPC/include/TPC/Clusters.h
+++ b/Modules/TPC/include/TPC/Clusters.h
@@ -19,10 +19,10 @@
 #define QC_MODULE_TPC_CLUSTERS_H
 
 // O2 includes
-#include "TPCQC/Clusters.h"
 #include "TPCQC/CalPadWrapper.h"
 
 // QC includes
+#include "TPC/ClustersData.h"
 #include "QualityControl/TaskInterface.h"
 
 class TCanvas;
@@ -54,7 +54,7 @@ class Clusters /*final*/ : public TaskInterface // todo add back the "final" whe
   void reset() override;
 
  private:
-  o2::tpc::qc::Clusters mQCClusters{};                         ///< O2 Cluster task to perform actions on cluster objects
+  ClustersData mQCClusters{};                                  ///< O2 Cluster task to perform actions on cluster objects
   std::vector<o2::tpc::qc::CalPadWrapper> mWrapperVector{};    ///< vector holding CalPad objects wrapped as TObjects; published on QCG; will be non-wrapped CalPad objects in the future
   std::vector<std::unique_ptr<TCanvas>> mNClustersCanvasVec{}; ///< summary canvases of the NClusters object
   std::vector<std::unique_ptr<TCanvas>> mQMaxCanvasVec{};      ///< summary canvases of the QMax object
--- a/Modules/TPC/include/TPC/Clusters.h
+++ b/Modules/TPC/include/TPC/Clusters.h
@@ -22,8 +22,8 @@
 #include "TPCQC/CalPadWrapper.h"
 
 // QC includes
-#include "TPC/ClustersData.h"
 #include "QualityControl/TaskInterface.h"
+#include "TPC/ClustersData.h"
 
 class TCanvas;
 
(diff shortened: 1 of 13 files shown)


knopers8 (commented):
Just to be sure you are aware, there is a way to let QCG visualize a custom class which inherits from a ROOT class:
https://github.com/AliceO2Group/QualityControl/blob/master/doc/Advanced.md#display-a-non-standard-root-object-in-qcg
Perhaps this way you could avoid using a postprocessing task to display the results, though i did not follow your code in detail

tklemenz:
Hello Piotr, thanks for the hint but I think in this case it is not possible to do it like this. In our case we have a data object `ClustersData` (mergeable) containing multiple `CalDet` objects where in the visualizer for each of them three canvases (containing multiple histograms) are created and displayed in the QCG. For example like `qc/TPC/MO/Clusters/c_Sides_Q_Max` in the test QCG. I think there is no other way to do it.

Modules/TPC/include/TPC/ClustersData.h
```diff
@@ -0,0 +1,68 @@
...
+namespace o2::quality_control_modules::tpc
+{
+
+class ClustersData final : public TObject, public MergeInterface
+{
+ public:
+  ClustersData()
+    : o2::mergers::MergeInterface(), TObject()
+  {
+  }
```
wiechula, line 41: Perhaps the constructor should take the name as argument, which defaults to `ClusterData`. If we want to use it for the Digits as well it should have a different name.

Modules/TPC/src/ClusterVisualizer.cxx
```diff
@@ -0,0 +1,181 @@
...
+      }
+    }
+  }
+
+  mPath = config.get<std::string>("qc.postprocessing." + name + ".path");
+}
+
+void ClusterVisualizer::initialize(Trigger, framework::ServiceRegistry&)
+{
+  mCdbApi.init("ccdb-test.cern.ch:8080");
```
wiechula, line 115: Shouldn't this be configurable from the json?

wiechula (changes requested):
Just small things.

Modules/TPC/include/TPC/ClustersData.h
tklemenz: I was actually thinking to do a separate class for the RawDigits task but I can try to do it within this one.

Modules/TPC/src/ClusterVisualizer.cxx
tklemenz: absolutely

Modules/TPC/include/TPC/ClustersData.h
tklemenz: I implemented a setName function that is called in the initialize function of the Clusters/RawDigits task to set the proper name of the data object.

tklemenz:
This PR needs https://github.com/AliceO2Group/AliceO2/pull/7792 from O2 now.

Modules/TPC/include/TPC/Clusters.h
```diff
@@ -54,7 +54,8 @@ class Clusters /*final*/ : public TaskInterface // todo add back the "final" whe
   void reset() override;
 
  private:
-  o2::tpc::qc::Clusters mQCClusters{};                         ///< O2 Cluster task to perform actions on cluster objects
+  bool mIsMergeable = true;
+  ClustersData mQCClusters{ "N_Clusters" };                    ///< O2 Cluster task to perform actions on cluster objects
```
wiechula, line 58:
```suggestion
  ClustersData mQCClusters{};                    ///< O2 Cluster task to perform actions on cluster objects
```
The default should be "N_Clusters" so this is not really needed.

wiechula (approved):
Just one very minor thing, which could also be fixed later. Please remove the [WIP]

Modules/TPC/include/TPC/Clusters.h
wiechula: But it also doesn't hurt to leave it like this...

tklemenz: done

wiechula (approved):

wiechula:
@knopers8 , I think the failure in the CI is unrelated. Could you have a look?

PR #967, opened 2021-11-25
Increase timeout for ccdb database test

Barthelemy:
Temporarily, until we understand why it is slow sporadically when ran in docker for the PRs check.


Barthelemy:
@ktf

PR #968, opened 2021-11-25
[QC-707] Open ROOT file sink only for the time it is needed

knopers8:
I did not test it, but if something is wrong then `batch_test` should see it.


knopers8:
@TimoWilken Clang format seems to complain about files which are not touched by this PR, could you please have a look?

TimoWilken:
@knopers8 the workflow used the wrong way to get a diff, so it complained about files that changed on master in the meantime as well. #969 should fix this.

knopers8:
Thanks!

Barthelemy (approved):

PR #970, opened 2021-11-25
restore decoding error histo if requested (default=no)

noferini:
@ercolessi @njacazio 
Add the possibility to require one additional histo (TRM errors) in TOF digit QC.
        "taskParameters": {
          "Diagnostic" : "false"
        },

if enabled it requires also this input to be added
patterns:TOF/PATTERNS

This is needed to check MC anchored to DATA (since we don't have this info from Raw QC there)

### DIFF
--- a/Modules/TOF/tofdigits.json
+++ b/Modules/TOF/tofdigits.json
@@ -1,111 +1,119 @@
 {
-  "qc": {
-    "config": {
-      "database": {
-        "implementation": "CCDB",
-        "host": "ccdb-test.cern.ch:8080",
-        "username": "not_applicable",
-        "password": "not_applicable",
-        "name": "not_applicable"
+  "qc" : {
+    "config" : {
+      "database" : {
+        "implementation" : "CCDB",
+        "host" : "ccdb-test.cern.ch:8080",
+        "username" : "not_applicable",
+        "password" : "not_applicable",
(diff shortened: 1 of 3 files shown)


njacazio:
Ciao @noferini we miss the clang format then I approve!

noferini:
> Ciao @noferini we miss the clang format then I approve!

I forgot it.
Done

Modules/TOF/tofdigits.json
```diff
@@ -1,111 +1,105 @@
...
-        "moduleName": "QcTOF",
-        "detectorName": "TOF",
-        "cycleDurationSeconds": "10",
-        "maxNumberCycles": "-1",
-        "dataSource_comment": "The other type of dataSource is \"direct\", see basic-no-sampling.json.",
-        "dataSource": {
-          "type": "dataSamplingPolicy",
-          "name": "tof-digits"
+    "tasks" : {
+      "TaskDigits" : {
```
njacazio, line 26: Maybe I would add the `:TOF/PATTERNS` to a task which is disabled so that we have trace of what is needed, what do you think?

noferini: I added also the policy tof-digits-dia
which can be used when enabling diagnostic

njacazio (approved):

PR #971, opened 2021-11-25
Fix for cases in which sourceID is -1

chiarazampolli:
@knopers8 , @sawenzel

### DIFF
--- a/Modules/GLO/src/VertexingQcTask.cxx
+++ b/Modules/GLO/src/VertexingQcTask.cxx
@@ -144,10 +144,10 @@ void VertexingQcTask::monitorData(o2::framework::ProcessingContext& ctx)
       }
     }
     for (const auto& lbl : mcLbl) {
-      auto header = mMCReader.getMCEventHeader(lbl.getSourceID(), lbl.getEventID());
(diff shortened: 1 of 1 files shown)


chiarazampolli:
This fixes the issue, but I am checking if it is normal that sourceID and eventID of the labels of the vertices can be -1.
Chiara

Modules/GLO/src/VertexingQcTask.cxx
```diff
@@ -144,10 +144,10 @@ void VertexingQcTask::monitorData(o2::framework::ProcessingContext& ctx)
       }
     }
     for (const auto& lbl : mcLbl) {
-      auto header = mMCReader.getMCEventHeader(lbl.getSourceID(), lbl.getEventID());
       if (lbl.getSourceID() != 0) { // using only underlying event,  which is source 0
```
knopers8, line 147: Maybe I am missing something obvious, but won't `-1` still pass this check?

knopers8: ah, sorry, i get it now :)

knopers8 (approved):

PR #979, opened 2021-12-03
Fix MFT readout QC config

knopers8:
@tomas-herman FYI, we discovered it while generating aliecs workflows for the new FLP suite release.

### DIFF
--- a/Modules/MFT/qc-mft-readout.json
+++ b/Modules/MFT/qc-mft-readout.json
@@ -50,7 +50,7 @@
         "policy" : "OnEachSeparately",
         "dataSource" : [ {
           "type" : "Task",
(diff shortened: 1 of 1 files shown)


Barthelemy:
@tomas-herman you should have received an invitation to gain write access to the repo. This will enable the possibility for you to review PRs in your detector directory and to get those assigned as well (CODEOWNER is not enough apparently).

tomas-herman (approved):

tomas-herman:
@knopers8 @Barthelemy thank you for catching this error and sending me the invitation. It was a change to get rid of the warning in infologger about the QC Task name longer than 13 characters and I missed this one place, sorry about that.

PR #980, opened 2021-12-03
Allow for dropped messages in batch_test

knopers8:
Dispatcher will be non-blocking soon, so we might see dropped messages in the tests. Let's just make sure there are not too many.


knopers8:
Needed for https://github.com/AliceO2Group/AliceO2/pull/7770

Barthelemy (approved):

PR #982, opened 2021-12-03
[EMCAL-530] Adding new class for cluster QC

vkrsingh:
Basic functionality monitoring cluster energy, position
and shower shape parameters, separately for EMCAL
and DCAL

### DIFF
--- a/Modules/EMCAL/include/EMCAL/ClusterTask.h
+++ b/Modules/EMCAL/include/EMCAL/ClusterTask.h
@@ -1,0 +1,73 @@
+// Copyright 2019-2020 CERN and copyright holders of ALICE O2.
+// See https://alice-o2.web.cern.ch/copyright for details of the copyright holders.
+// All rights not expressly granted are reserved.
+//
+// This software is distributed under the terms of the GNU General Public
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+///
+/// \file   ClusterTask.h
+/// \author Vivek Kumar Singh
+///
+
+#ifndef QC_MODULE_EMC_EMCCLUSTERTASK_H
+#define QC_MODULE_EMC_EMCCLUSTERTASK_H
+
+#include "QualityControl/TaskInterface.h"
+#include <DataFormatsEMCAL/EventHandler.h>
+#include <EMCALBase/ClusterFactory.h>
+#include "EMCALBase/Geometry.h"
+
+class TH1;
+class TH2;
+
+using namespace o2::quality_control::core;
+
+namespace o2::quality_control_modules::emcal::ClusterTask
+{
+
+/// \brief Example Quality Control DPL Task
+/// \author Vivek Kumar Singh
+class ClusterTask final : public TaskInterface
+{
+ public:
+  /// \brief Constructor
+  ClusterTask() = default;
+  /// Destructor
+  ~ClusterTask() override;
+
+  // Definition of the methods for the template method pattern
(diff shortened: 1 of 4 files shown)


Modules/EMCAL/src/ClusterTask.cxx
```diff
@@ -0,0 +1,217 @@
...
+  mEventHandler = std::make_unique<o2::emcal::EventHandler<o2::emcal::Cell>>();
+  mClusterFactory = std::make_unique<o2::emcal::ClusterFactory<o2::emcal::Cell>>();
+
+  mEventHandler->setClusterData(Cluster, CellIndex, ClusterTriggerRecord, CellIndexTriggerRecord);
+  mEventHandler->setCellData(Cell, CellTriggerRecord);
+
+  for (int iev = 0; iev < mEventHandler->getNumberOfEvents(); iev++) {
+    auto inputEvent = mEventHandler->buildEvent(iev);
+
+    resetHistograms();
```
mfasDa, line 143: Histograms must not be reset in the event loop. The idea is to fill histograms over a longer time in order to accumulate some statistics. Usually histograms a accumulative and are reset only at the start of run (in startOfActity), with an option to have this per cycle duration.

Furthermore: If you reset histograms within the event loop the data will never get published - there is no publishing of histograms while the run function is still active (no asynchronous processing). If you reset the histograms in the loop the data is lost forever.

Never call resetHistograms actively (except from startOfActivity) - let this be done by the framework.

Modules/EMCAL/src/ClusterTask.cxx
```diff
@@ -0,0 +1,217 @@
...
+      Bool_t clsTypeEMC;
+
+      math_utils::Point3D<float> emcx = analysisCluster.getGlobalPosition();
+
+      TVector3 clustpos(emcx.X(), emcx.Y(), emcx.Z());
+      Double_t emcphi = clustpos.Phi();
+      Double_t emceta = clustpos.Eta();
+
+      if (emcphi < 0)
+        emcphi = TVector2::Phi_0_2pi(emcphi); //TLorentz vector is defined between -pi to pi, so negative phi has to be flipped
```
mfasDa, line 166: if statement not needed, Phi_0_2pi will check internally whether the input phi is negative. You can do directly

``` c++
 Double_t emcphi = TVector2::Phi_0_2pi(clustpos.Phi());
 Double_t emceta = clustpos.Eta();
 ```

**Review by mfasDa — CHANGES_REQUESTED** [review]
For the first iteration mostly OK, except from the reset, the function must be provided but not called explicitly, see explanation in the code. Furthermore one small correction.

**Review by mfasDa — APPROVED** [review]
Now it is fine to be merged.

File: Modules/EMCAL/include/EMCAL/ClusterTask.h [hunk 3]
```diff
@@ -0,0 +1,70 @@
...
+  /// Destructor
+  ~ClusterTask() override;
+
+  // Definition of the methods for the template method pattern
+  void initialize(o2::framework::InitContext& ctx) override;
+  void startOfActivity(Activity& activity) override;
+  void startOfCycle() override;
+  void monitorData(o2::framework::ProcessingContext& ctx) override;
+  void endOfCycle() override;
+  void endOfActivity(Activity& activity) override;
```
mfasDa, line 49: missing
``` c++
 void reset() override;
```

Modules/EMCAL/include/EMCAL/ClusterTask.h
```diff
@@ -0,0 +1,70 @@
...
+  void resetHistograms();
```
mfasDa, line 50: Should be protected, access from outside should be steered via reset.

mfasDa (changes requested):
Needs virtual reset function to be implemented.

mfasDa (approved):
Should be OK now, ready to merge.

PR #1003, opened 2022-01-10
TRD reduce spectrum count

bazinski:
Collapse the 540 cluster spectra, 1 for each chamber, into a singular 2d histogram.
This should lighten the load, we still end up with the same information.

Additonally add the currently used json file from my home directory to this github repo. Its sampling rate must be reduced still.

### DIFF
--- a/Modules/TRD/TRDQC.json
+++ b/Modules/TRD/TRDQC.json
@@ -1,0 +1,144 @@
+{
+  "qc": {
+    "config": {
+      "database": {
+        "implementation": "CCDB",
+        "host": "ccdb-test.cern.ch:8080",
+        "username": "not_applicable",
+        "password": "not_applicable",
+        "name": "not_applicable"
+      },
+      "Activity": {
+        "number": "505618",
+        "type": "2",
+        "periodName": "",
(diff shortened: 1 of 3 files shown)


Modules/TRD/TRDQC.json
```diff
@@ -0,0 +1,144 @@
...
+        "type": "2",
+        "periodName": "",
+        "passName": "",
+        "provenance": "qc"
+      },
+      "monitoring": {
+        "url": "infologger:///debug?qc"
+      },
+      "consul": {
+        "url": "http://consul-test.cern.ch:8500"
```
Barthelemy, line 22: The url for consul should be left empty. 
```suggestion
        "url": ""
```

Barthelemy (commented):
Thank you Sean. Just one little change required.

Modules/TRD/TRDQC.json
bazinski: Ah shot, was never what was supposed to go in that stuff in anycase ;-)

Barthelemy:
The failing test is unrelated, merging.

PR #1004, opened 2022-01-11
[QC-684] Refactor task and check names

Barthelemy:


Barthelemy:
@TimoWilken @ktf There seems to be something worng with the builders.

knopers8 (approved):

PR #1010, opened 2022-01-14
[QC-310] Refresh config in Runners::init

Barthelemy:
We use the ptree provided by ECS via DPL/FairMQ to rebuild a config in `init()` and use it. If it is not available we just keep the one we got at the beginning. 

I had to refactor a bit to have something understandable in the runners and be able to reconfigure. 

One thing is missing in the PR: the CheckRunner does not actually apply the config. This will be done in a later PR as it will require some refactoring and this PR is already too big.


Framework/include/QualityControl/AggregatorRunnerFactory.h
```diff
@@ -36,7 +37,7 @@ class AggregatorRunnerFactory
   AggregatorRunnerFactory() = default;
   virtual ~AggregatorRunnerFactory() = default;
 
-  static framework::DataProcessorSpec create(AggregatorRunnerConfig arc, std::vector<AggregatorConfig> acs);
+  static framework::DataProcessorSpec create(const o2::quality_control::core::InfrastructureSpec& infrastructureSpec);
```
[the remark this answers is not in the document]
knopers8: I don't like that it requres the whole `InfrastructureSpec` to go forward. How about reducing it to what it actually needs?
Also, requiring `InfrastructureSpec` instead of `Configs` makes an exception with regards to other QC devices, but perhaps it cannot be done in a different way.

Framework/include/QualityControl/CheckRunner.h
```diff
@@ -114,7 +114,7 @@ class CheckRunner : public framework::Task
   void setTaskStoreSet(std::unordered_set<std::string> storeSet) { mInputStoreSet = storeSet; }
   std::string getDeviceName() { return mDeviceName; };
 
-  static framework::DataProcessorLabel getLabel() { return { "qc-check-runner" }; }
+  static framework::DataProcessorLabel getCheckLabel() { return { "qc-check-runner" }; }
```
knopers8, line 117: Hmm, it is not a Check label, but CheckRunner's, isn' it?

Framework/include/QualityControl/TaskRunner.h
```diff
@@ -90,7 +90,7 @@ class TaskRunner : public framework::Task
   const framework::Options& getOptions() const { return mTaskConfig.options; };
 
   /// \brief Data Processor Label to identify all Task Runners
-  static framework::DataProcessorLabel getLabel() { return { "qc-task-runner" }; }
+  static framework::DataProcessorLabel getTaskLabel() { return { "qc-task-runner" }; }
```
knopers8, line 93: The same comment as for `getCheckLabel`

Framework/include/QualityControl/runnerUtils.h
```diff
@@ -119,6 +119,43 @@ inline std::string computeProvenance(const std::string& fallbackProvenance = "")
   return provenance;
 }
 
+inline std::string indentTree(int level)
+{
+  std::string s;
+  for (int i = 0; i < level; i++)
+    s += "  ";
+  return s;
+}
```
knopers8, line 128:
```suggestion
inline std::string indentTree(int level)
{
  return std::string(level * 2, ' ');
}
```
One-liner version

Framework/include/QualityControl/runnerUtils.h
```diff
@@ -119,6 +119,43 @@ inline std::string computeProvenance(const std::string& fallbackProvenance = "")
...
+
+inline void printTree(boost::property_tree::ptree& pt, int level = 0)
```
knopers8, line 130: I can't quite follow the whole logic, but does it work for arrays? I don't see any `[` and `]` characters.

Framework/src/AggregatorRunner.cxx
```diff
@@ -68,10 +120,13 @@ AggregatorRunner::AggregatorRunner(AggregatorRunnerConfig arc, const std::vector
   }
 }
 
-AggregatorRunner::~AggregatorRunner()
+void AggregatorRunner::extractConfigs(const o2::quality_control::core::InfrastructureSpec& infrastructureSpec)
```
knopers8, line 123:
```suggestion
void AggregatorRunner::extractConfigs(const core::InfrastructureSpec& infrastructureSpec)
```

Framework/src/TaskRunner.cxx
```diff
@@ -59,7 +63,40 @@ TaskRunner::TaskRunner(const TaskRunnerConfig& config)
...
+      }
+
+      // prepare the information we need
+      auto infrastructureSpec = InfrastructureSpecReader::readInfrastructureSpec(updatedTree);
+      // find the correct taskSpec
+      auto it = find_if(infrastructureSpec.tasks.begin(),
+                        infrastructureSpec.tasks.end(),
+                        [this](const TaskSpec& ts) { return ts.taskName == mTaskConfig.taskName; });
+      if (it != infrastructureSpec.tasks.end()) {
+        mTaskConfig = TaskRunnerFactory::extractConfig(infrastructureSpec.common, *it, mTaskConfig.parallelTaskID, it->resetAfterCycles);
```
knopers8, line 87: `it->resetAfterCycles`
That will break if we have a multinode setup with local tasks and delta mergers. See:
https://github.com/AliceO2Group/QualityControl/blob/95eb565ef04146919c2034099e26289ae766230e/Framework/src/InfrastructureGenerator.cxx#L136

It seems to me that maybe we should somehow extract such logic and use the same function in `InfrastructureGenerator` and here?

Framework/src/TaskRunnerFactory.cxx
```diff
@@ -22,14 +22,22 @@
...
 #include <Headers/DataHeader.h>
+#include <Framework/ConfigParamSpec.h>
+#include <Framework/CallbackService.h>
+#include <Framework/CompletionPolicyHelpers.h>
+#include <Framework/TimesliceIndex.h>
+#include <Framework/DataSpecUtils.h>
+#include <Framework/InputRecordWalker.h>
+#include <Framework/InputSpan.h>
+#include <Framework/O2ControlLabels.h>
+#include <Framework/DataProcessorLabel.h>
```
knopers8, line 33: Coudl you please make sure we really need all of this? :D

knopers8 (changes requested):
Thanks, indeed let's keep Check Runners for another PR, that is already significant. Please see my comments.

Framework/include/QualityControl/AggregatorRunner.h
```diff
@@ -135,6 +136,14 @@ class AggregatorRunner : public framework::Task
    */
   void store(core::QualityObjectsType& qualityObjects);
 
+  void refreshConfig(framework::InitContext& iCtx);
+  inline void extractConfigs(const o2::quality_control::core::InfrastructureSpec& infrastructureSpec);
```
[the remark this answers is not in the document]
Barthelemy: I had read that modern compilers are able to do inline accross files. I tried to find it back but only found https://stackoverflow.com/questions/8450030/can-modern-c-compilers-inline-functions-that-are-defined-in-a-cpp-file
It seems that some compilers can do that but they do it on their own and the inline keyword is more or less useless.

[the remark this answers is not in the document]
Barthelemy: removed

Framework/include/QualityControl/AggregatorRunnerFactory.h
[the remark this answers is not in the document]
Barthelemy: Yes, I agree, I did not realize that. I changed it.

Framework/include/QualityControl/CheckRunner.h
Barthelemy: yep

Framework/include/QualityControl/runnerUtils.h
Barthelemy: well, I learned something here :)

Framework/include/QualityControl/runnerUtils.h
Barthelemy: Yes, I am not sure why I went this route. I have replaced it with the standard `json_parser`

Framework/src/TaskRunnerFactory.cxx
Barthelemy: let me add a few more ;)

Framework/src/TaskRunner.cxx
Barthelemy: Good that you spotted that, I completely overlooked the issue.

knopers8 (approved):

knopers8:
Please resolve the conflicts.

PR #1019, opened 2022-01-24
QC-310: reconfigure checkrunner

Barthelemy:
We ignore the topology changes at this stage. We update (delete and rebuild) the checks that belong to the check runner.

### DIFF
--- a/Framework/src/CheckRunner.cxx
+++ b/Framework/src/CheckRunner.cxx
@@ -212,6 +212,7 @@ void CheckRunner::init(framework::InitContext& iCtx)
     iCtx.services().get<CallbackService>().set(CallbackService::Id::Reset, [this]() { reset(); });
     iCtx.services().get<CallbackService>().set(CallbackService::Id::Stop, [this]() { stop(); });
 
+    updatePolicyManager.reset();
     for (auto& [checkName, check] : mChecks) {
       check.init();
       updatePolicyManager.addPolicy(check.getName(), check.getUpdatePolicyType(), check.getObjectsNames(), check.getAllObjectsOption(), false);
--- a/Framework/src/CheckRunner.cxx
+++ b/Framework/src/CheckRunner.cxx
@@ -187,6 +187,7 @@ void CheckRunner::refreshConfig(InitContext& iCtx)
         // search if we have this check in this runner and replace it
         if (mChecks.find(checkSpec.checkName) != mChecks.end()) {
           auto checkConfig = Check::extractConfig(infrastructureSpec.common, checkSpec);
+          mChecks.erase(checkConfig.name);
           mChecks.emplace(checkConfig.name, checkConfig);
           ILOG(Debug, Devel) << "Check " << checkSpec.checkName << " has been updated" << ENDM;
         }
--- a/Framework/src/CheckRunner.cxx
+++ b/Framework/src/CheckRunner.cxx
@@ -185,10 +185,10 @@ void CheckRunner::refreshConfig(InitContext& iCtx)
       // Topology changes are ignored: New checks are ignored. Removed checks are ignored.
       for (const auto& checkSpec : infrastructureSpec.checks) {
         // search if we have this check in this runner and replace it
-        if ( mChecks.find(checkSpec.checkName) != mChecks.end() ) {
+        if (mChecks.find(checkSpec.checkName) != mChecks.end()) {
           auto checkConfig = Check::extractConfig(infrastructureSpec.common, checkSpec);
           mChecks.emplace(checkConfig.name, checkConfig);
-          ILOG(Debug, Devel) << "Check "<< checkSpec.checkName << " has been updated" << ENDM;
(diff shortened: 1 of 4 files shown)


Framework/src/CheckRunner.cxx
```diff
@@ -314,18 +319,18 @@ QualityObjectsType CheckRunner::check()
                       << ENDM;
 
   QualityObjectsType allQOs;
-  for (auto& check : mChecks) {
-    if (updatePolicyManager.isReady(check.getName())) {
-      auto newQOs = check.check(mMonitorObjects);
+  for (auto& tuple : mChecks) {
```
knopers8, line 322:
```suggestion
  for (auto& [checkName, check] : mChecks) {
```
something like this would be more clear, same comment above

Framework/src/CheckRunner.cxx
```diff
@@ -181,11 +180,17 @@ void CheckRunner::refreshConfig(InitContext& iCtx)
...
-      // TODO: we should probably just preserve the checks list and update their state.
-      // TODO: also see if we should update the detector name
+      // Use the config to reconfigure the check runner.
+      // The configs for the checks we find in the config and in our map are updated.
+      // Topology changes are ignored: New checks are ignored. Removed checks are ignored.
+      for (const auto& checkSpec : infrastructureSpec.checks) {
+        // search if we have this check in this runner and replace it
+        if ( mChecks.find(checkSpec.checkName) != mChecks.end() ) {
+          auto checkConfig = Check::extractConfig(infrastructureSpec.common, checkSpec);
+          mChecks[checkSpec.checkName] = checkConfig;
```
knopers8, line 190: I am surprised that a Check constructor for checkConfig can be used this way, but if that compiles then I have nothing against.

Framework/src/CheckRunner.cxx
```diff
@@ -207,9 +212,9 @@ void CheckRunner::init(framework::InitContext& iCtx)
     iCtx.services().get<CallbackService>().set(CallbackService::Id::Reset, [this]() { reset(); });
     iCtx.services().get<CallbackService>().set(CallbackService::Id::Stop, [this]() { stop(); });
 
-    for (auto& check : mChecks) {
-      check.init();
-      updatePolicyManager.addPolicy(check.getName(), check.getUpdatePolicyType(), check.getObjectsNames(), check.getAllObjectsOption(), false);
+    for (auto& tuple : mChecks) {
+      tuple.second.init();
+      updatePolicyManager.addPolicy(tuple.second.getName(), tuple.second.getUpdatePolicyType(), tuple.second.getObjectsNames(), tuple.second.getAllObjectsOption(), false);
```
knopers8, line 217: What about the policies existing before the refresh? i don't see them being cleaned up.

knopers8 (changes requested):
Thank you, the general idea seems correct, but please consider my comments.

Framework/src/CheckRunner.cxx
Barthelemy: totally agree

Framework/src/CheckRunner.cxx
Barthelemy: To be honest I was surprised myself :) but this was already the way it was done in the constructor.

Framework/src/CheckRunner.cxx
Barthelemy: correct, good catch. fixing

Barthelemy:
I have addressed all comments.

Framework/src/CheckRunner.cxx
```diff
@@ -181,11 +180,17 @@ void CheckRunner::refreshConfig(InitContext& iCtx)
...
-      // TODO: we should probably just preserve the checks list and update their state.
-      // TODO: also see if we should update the detector name
+      // Use the config to reconfigure the check runner.
+      // The configs for the checks we find in the config and in our map are updated.
+      // Topology changes are ignored: New checks are ignored. Removed checks are ignored.
+      for (const auto& checkSpec : infrastructureSpec.checks) {
+        // search if we have this check in this runner and replace it
+        if (mChecks.find(checkSpec.checkName) != mChecks.end()) {
+          auto checkConfig = Check::extractConfig(infrastructureSpec.common, checkSpec);
+          mChecks.emplace(checkConfig.name, checkConfig);
```
knopers8, line 190: From https://www.cplusplus.com/reference/map/map/emplace/ :
> The insertion only takes place if no other element in the container has a key equivalent to the one being emplaced (keys in a map container are unique).

I think that means you are not overwriting the configuration.

Barthelemy: You are totally correct. I guess it is a time to use the new `insert_or_assign`

Barthelemy: well no, it does an insert and does not construct the object. I will just erase the object if it is there.

Barthelemy: @knopers8 I let you make a final check

knopers8 (approved):

PR #1022, opened 2022-01-25
MCH: fixed memory leak in custom mergeable histograms

aferrero2707:
The memory leak was due to the fact that internal histograms of the mergeable objects were not deleted in the object's destructor.

The code also reduces the number of objects published by default by the digits task. A "Diagnostic" custom parameter re-enables the publishing of the additional histograms (for experts only).

### DIFF
--- a/Modules/MUON/MCH/include/MCH/GlobalHistogram.h
+++ b/Modules/MUON/MCH/include/MCH/GlobalHistogram.h
@@ -31,19 +31,36 @@ namespace muonchambers
 
 std::string getHistoPath(int deId);
 
-class DetectorHistogram : public TH2F
+class DetectorHistogram
 {
  public:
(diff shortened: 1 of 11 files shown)


Barthelemy:
Could you fix the format ? if we can review this morning it can be part of the next flp suite

aferrero2707:
I am on that, but I have to do it manually as I am recompiling the O2 stack after an update...

aferrero2707:
@Barthelemy clang fomatting check are now ok. I have still to commit the improvements for the automated checks of the channel-by-channel occupancies and of the reference json files, but I propose to postpone this for the next PR since they require some more test and validation. Therefore this PR fully focuses on the fix of the memory leaks and the solution of the histogram merging problems we had during the pilot run.

Barthelemy (approved):

PR #1025, opened 2022-01-28
ITS tracks: bug fixes + general improvements

iravasen:
Main changes:
- solved bug in the calculation of fraction of clusters into tracks
- improved scales of several plots (acting also on the json)
- improved position of QC check messages + some of the checks


iravasen:
For us (ITS) can be merged as soon as checks are passed

Barthelemy:
ok, I set auto-merge.

Barthelemy (approved):

PR #1026, opened 2022-01-31
[QC-670] set a maximum size for the objects to be sent to the repo

Barthelemy:


Barthelemy:
@knopers8 When you have a moment to review, it is ready

Framework/basic.json
```diff
@@ -6,7 +6,8 @@
         "host": "ccdb-test.cern.ch:8080",
         "username": "not_applicable",
         "password": "not_applicable",
-        "name": "not_applicable"
+        "name": "not_applicable",
+        "maxObjectSize": "2097152",       "": "[Bytes, default=2MB] Maximum size allowed, larger objects are rejected."
```
knopers8, line 10: to be picky, that's actually 2MiB ;)

knopers8 (approved):

Framework/basic.json
Barthelemy: correct

PR #1027, opened 2022-01-31
[QC-736] Set the facility after other flags are set

Barthelemy:
@davidrohr

### DIFF
--- a/Framework/src/runQC.cxx
+++ b/Framework/src/runQC.cxx
@@ -159,8 +159,6 @@ WorkflowSpec defineDataProcessing(const ConfigContext& config)
 {
   WorkflowSpec specs;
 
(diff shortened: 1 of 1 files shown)


davidrohr:
thx!

davidrohr:
Unfortunately, I have no easy way to test it without having an RPM build and installing that on the EPNs.
But the fix looks so trivial, that I am 99.9% sure it is correct.
Could we have a new QC tag with this?

Barthelemy:
Sure we can, I will release a patch version.

Barthelemy:
Here is the patch version: https://github.com/AliceO2Group/QualityControl/releases/tag/v1.45.1
And the pr in alidist: https://github.com/alisw/alidist/pull/3797

PR #1028, opened 2022-02-01
Add multiplicity histo

ValerieRamillien:


Modules/MUON/MID/src/DigitsQcTask.cxx
```diff
@@ -197,6 +237,84 @@ void DigitsQcTask::monitorData(o2::framework::ProcessingContext& ctx)
...
+
+    // loadStripPatterns (ColumnData)
+    for (auto& digit : digits.subspan(rofRecord.firstEntry, rofRecord.nEntries)) { // loop DE //
+      int deIndex = digit.deId;
+      int colId = digit.columnId;
+      int rpcLine = o2::mid::detparams::getRPCLine(deIndex);
+      int ichamber = o2::mid::detparams::getChamber(deIndex);
+      auto isRightSide = o2::mid::detparams::isRightSide(deIndex);
+      auto detId = o2::mid::detparams::getDEId(isRightSide, ichamber, colId);
+      for (int i = 0; i < 4; i++) {
```
dstocco, line 263: The code is correct. Notice however that you could use the mapping to do this part (to avoid code duplication). If you add the data member: 
`Mapping mMapping; ///< Mapping`
in the .h, this part could simply be:
```
for (int board = mMapping.getFirstBoardBP(colId, deIndex), lastBoard = mMapping.getLastBoardBP(colId, deIndex); board <= lastBoard; board++) { // These are the existing boards for this column Id
  if (digit.patterns[board] != 0) {
    double linePos = rpcLine + 0.01 + 0.25*board;
    if (isRightSide)
      mLocalBoardsMap->Fill(colId + 0.5, linepos, 1);
     else
       mLocalBoardsMap->Fill(-colId - 0.5, linepos, 1);
  }
}
```

dstocco:
@Barthelemy : this PR is ok for me. The suggestion I'm proposing is not fundamental. Can you please have a look so that the compilation can start?

Modules/MUON/MID/src/DigitsQcTask.cxx
ValerieRamillien: I will make the modifications in the next commit because in this case it is necessary to add additional instructions for the display  (for DEs with only 1 or 2 local cards)

Barthelemy (approved):

PR #1029, opened 2022-02-04
Recommend to remove unneeded parts of the example code

knopers8:
@Barthelemy We often see some leftovers of examples we put in the Skeleton. How about something like this?

Perhaps it can be "IT IS SAFE TO REMOVE ANYTHING YOU DO NOT NEED" instead.


Barthelemy (approved):
Totally agree

PR #1031, opened 2022-02-04
QC-641 Example of what I meant : remove the param of configure and ca…

Barthelemy:
…ll it when setCustomParameters is called ;.


knopers8 (approved):

Barthelemy:
@TimoWilken Hi, it seems that there is something wrong with the builders.

TimoWilken:
There was a temporary problem connecting to gitlab.cern.ch. This PR is being rechecked at the moment.

PR #1034, opened 2022-02-07
[QC-670] Update doc about max object size.

Barthelemy:


Barthelemy:
@TimoWilken @ktf  Good morning, it seems that there is a problem with the mac builder. Could it be possible to have an automatic recovery of the builders ? it seems to me that it happens quite often.

PR #1037, opened 2022-02-09
adjust CPV physics task

sevdokim:
Update of CPV physics task: now it can process clusters without querying digits and vice versa. Is needed run on async reco where cpv CTFs contain only clusters.

### DIFF
--- a/Modules/CPV/src/PhysicsTask.cxx
+++ b/Modules/CPV/src/PhysicsTask.cxx
@@ -54,7 +54,6 @@ void PhysicsTask::initialize(o2::framework::InitContext& /*ctx*/)
   }
   initHistograms();
   mNEventsTotal = 0;
-  mNEventsFromLastFillHistogramsCall = 0;
 }
 
 void PhysicsTask::startOfActivity(Activity& activity)
@@ -62,7 +61,6 @@ void PhysicsTask::startOfActivity(Activity& activity)
   ILOG(Info, Support) << "startOfActivity() : resetting everything" << activity.mId << ENDM;
   resetHistograms();
   mNEventsTotal = 0;
-  mNEventsFromLastFillHistogramsCall = 0;
 }
 
 void PhysicsTask::startOfCycle()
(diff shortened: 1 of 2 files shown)


Modules/CPV/src/PhysicsTask.cxx
```diff
@@ -90,10 +88,21 @@ void PhysicsTask::monitorData(o2::framework::ProcessingContext& ctx)
...
     if (input.header != nullptr && input.payload != nullptr) {
       const auto* header = header::get<header::DataHeader*>(input.header);
+      //LOG(info) << "monitorData() : obtained input " << header->dataOrigin.str << "/" << header->dataDescription.str;
+      if ((strcmp(header->dataOrigin.str, "CPV") == 0) && (strcmp(header->dataDescription.str, "DIGITS") == 0)) {
+        LOG(info) << "monitorData() : I found digits in inputs";
+        hasDigits = true;
+      }
+      if ((strcmp(header->dataOrigin.str, "CPV") == 0) && (strcmp(header->dataDescription.str, "CLUSTERS") == 0)) {
+        LOG(info) << "monitorData() : I found clusters in inputs";
+        hasDigits = true;
```
knopers8, line 103: hasClusters?

sevdokim: surely clusters! thanks for spotting!

Modules/CPV/src/PhysicsTask.cxx
```diff
@@ -90,10 +88,21 @@ void PhysicsTask::monitorData(o2::framework::ProcessingContext& ctx)
...
   mHist1D[H1DNValidInputs]->Fill(nValidInputs);
 
+  bool hasClusters = false, hasDigits = false;
   for (auto&& input : ctx.inputs()) {
     // get message header
     if (input.header != nullptr && input.payload != nullptr) {
       const auto* header = header::get<header::DataHeader*>(input.header);
+      //LOG(info) << "monitorData() : obtained input " << header->dataOrigin.str << "/" << header->dataDescription.str;
+      if ((strcmp(header->dataOrigin.str, "CPV") == 0) && (strcmp(header->dataDescription.str, "DIGITS") == 0)) {
+        LOG(info) << "monitorData() : I found digits in inputs";
```
Barthelemy, line 98: I would be extremely careful about printing and logging in `monitorData`. This is typically called at high frequency. It will create a flood. I would advise
2. to remove it, if not
2. to make the log DEBUG and limit the number of times it can be printed, and
3. to use the infologger instead of the fairlogger (not mandatory).

Barthelemy (changes requested):

Modules/CPV/src/PhysicsTask.cxx
sevdokim: Hi @Barthelemy 
indeed logging there was unnecessary and was for debug purpose. I forgot to remove it in hurry. Thanks!

Barthelemy (approved):

PR #1040, opened 2022-02-10
[WIP] TrendingITSFHR for OB update

BeHckuU:
1)Added plots for OB layers
2)Changed marker style and colors
3)Changed plot and Y axis titles (# active chips → # active lanes)
4)Changed plot Y ranges for OB
5)Changed plot names for OB("activelanes_trends" → "activelanes_trends")


Modules/ITS/src/TrendingTaskITSFhr.cxx
```diff
@@ -140,23 +140,21 @@ void TrendingTaskITSFhr::storePlots(repository::DatabaseInterface& qcdb)
   int countplots = 0;
   int ilay = 0;
   double ymin[NTRENDSFHR] = { 1e-15, 1e-1, -.5, 1e-9 };
-  double ymax[NTRENDSFHR] = { 1e-3, 1e-5, 9.5, 1 };
+  double ymaxIB[NTRENDSFHR] = { 1e-3, 1e-5, 9.5, 1 };
+  double ymaxOB[NTRENDSFHR] = { 1e-3, 1e-5, 30.5, 1 };
 
-  //Loop on plots
+  // Loop on plots
```
iravasen, line 146: I would remove this loop which save every single histo to the CCDB **provided** track each histo are later drawn and saved into a Canvas including all staves of a layer. Is that true for all histos? 
In this way we drastically reduce the amount of histo going to CCDB.

PR #1044, opened 2022-02-15
Doc: Mailing lists should have less priority when looking for help

knopers8:
@Barthelemy I would put discourse on top, because anyone can find old issues there and those who do not want to see support questions can unwatch. On the other hand, our QC contacts should not unsubscribe from the QC contact mailing list. Do you agree?


Barthelemy:
Agreed as discussed.

Barthelemy (approved):

PR #1053, opened 2022-02-22
Get payload size from DataRefUtils instead of DataHeader

matthiasrichter:
The extended data model allows to have have multiple payloads per header
which requires to extract the actual data size from the message size.
`DataRefUtils::getPayloadSize` handles this transparently.

### DIFF
--- a/Modules/Benchmark/src/TH1FTask.cxx
+++ b/Modules/Benchmark/src/TH1FTask.cxx
@@ -20,6 +20,8 @@
 #include <Headers/DataHeader.h>
 #include "QualityControl/QcInfoLogger.h"
 #include <Framework/InputRecord.h>
+#include <Framework/InputRecordWalker.h>
+#include <Framework/DataRefUtils.h>
 
 namespace o2::quality_control_modules::benchmark
 {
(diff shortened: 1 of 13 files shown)


Modules/CPV/src/PedestalTask.cxx
```diff
@@ -148,31 +150,6 @@ void PedestalTask::monitorData(o2::framework::ProcessingContext& ctx)
...
-  //   auto s = ctx.inputs().get<TObjString*>("string");
-  //   LOG(info) << "String is " << s->GetString().Data();
-
-  // 3. Access CCDB. If it is enough to retrieve it once, do it in initialize().
-  // Remember to delete the object when the pointer goes out of scope or it is no longer needed.
-  //   TObject* condition = TaskInterface::retrieveCondition("QcTask/example"); // put a valid condition path here
-  //   if (condition) {
-  //     LOG(info) << "Retrieved " << condition->ClassName();
-  //     delete condition;
-  //   }
```
matthiasrichter, line 175: Have simply deleted this, its from the template and not used here

Modules/Benchmark/src/TH1FTask.cxx
```diff
@@ -70,10 +72,13 @@ void TH1FTask::monitorData(o2::framework::ProcessingContext& ctx)
 {
   // the minimum that we can do, which includes accessing data and creating non-empty histograms
   size_t dummySum = 0;
-  for (auto&& input : ctx.inputs()) {
+  for (auto&& input : o2::framework::InputRecordWalker(ctx.inputs())) {
```
matthiasrichter, line 75: Using InputRecordWalker ensures looping over all parts of all inputs.

knopers8 (approved):
Thank you! I am fine with merging it as it is, but if you don't mind, I will keep it for a day so any detector modules developers also have their say.

matthiasrichter:
Yes, that's fine. We should leave time for other comments..

PR #1046, opened 2022-02-16
Commited ZDC Module it contains the first version of ZDCRawDataTask. …

cpuggion84:
Created new module ZDC.
Created new task ZDCRawDataTask
Created new checker ZDCRawDataCheck (Created but not yet implemented)
Inside task ZDCRawDataTask produced these Histograms:
26 histograms for pedestals: will be used to provide calibration object for reconstruction (backup average pedestal information,
in case of missing information in current event and current orbit)
1 pedestal summary for central shifter
26 scalers: can be used to produce graphs of luminosity evolution (TBD)
26  signals “oscilloscope-like” waveforms
10 BC maps to align ALICE trigger
10 BC maps to align AUTO trigger
10 BC maps to align AUTO trigger or ALICE trigger
1 channels trasmitted
1 channels fired

PR #1056, opened 2022-02-24
Update CMakeLists.txt

cpuggion84:
ADD subdirectory ZDC


knopers8:
Please include this change in https://github.com/AliceO2Group/QualityControl/pull/1046
By putting it in a separate PR, you are trying to include a non-existent directory, so the tests will not even pass.

PR #1063, opened 2022-03-01
ITS Calibration Threshold Task

IsakovAD:
First version of the ITS Calibration Task. 

Analyses outputs of the "o2-its-threshold-calib-workflow" for VCASN, ITHR and THR runs. 
Provides the following histograms:

- 2D chip vs stave distribution of averaged VCASN, ITHR, THR, NoiseTHR and their RMS's
- 1D average distribution of average VCASN, ITHR, THR, NoiseTHR and their RMS's for each chip in one layer 
- 1D Success Rate histogram

### DIFF
--- a/Modules/ITS/src/ITSThresholdCalibrationTask.cxx
+++ b/Modules/ITS/src/ITSThresholdCalibrationTask.cxx
@@ -1,0 +1,456 @@
+// Copyright CERN and copyright holders of ALICE O2. This software is
+// // distributed under the terms of the GNU General Public License v3 (GPL
+// // Version 3), copied verbatim in the file "COPYING".
+// //
+// // See http://alice-o2.web.cern.ch/license for full licensing information.
+// //
+// // In applying this license CERN does not waive the privileges and immunities
+// // granted to it by virtue of its status as an Intergovernmental Organization
+// // or submit itself to any jurisdiction.
+//
+
+///
+/// \file   ITSThresholdCalibrationTask.cxx
+/// \author Artem Isakov
+///
+
+#include "QualityControl/QcInfoLogger.h"
+#include "ITS/ITSThresholdCalibrationTask.h"
+#include <DataFormatsITS/TrackITS.h>
+#include <DataFormatsITSMFT/ROFRecord.h>
+#include <Framework/InputRecord.h>
+#include "ReconstructionDataFormats/Vertex.h"
+#include "ReconstructionDataFormats/PrimaryVertex.h"
+#include "TLine.h"
+#include "TLatex.h"
+#include <Framework/DataSpecUtils.h>
+
+using namespace o2::itsmft;
+using namespace o2::its;
+
+namespace o2::quality_control_modules::its
+{
+
+ITSThresholdCalibrationTask::ITSThresholdCalibrationTask() : TaskInterface()
+{
+}
+
+ITSThresholdCalibrationTask::~ITSThresholdCalibrationTask()
+{
+  delete hSuccessRate;
+
+  for (int iScan = 0; iScan < 3; iScan++) {
+    for (int iLayer = 0; iLayer < 7; iLayer++) {
+      delete hCalibrationLayer[iLayer][iScan];
+      delete hCalibrationRMSLayer[iLayer][iScan];
+      if (iScan == 2) {
+        delete hCalibrationThrNoiseLayer[iLayer];
+        delete hCalibrationThrNoiseRMSLayer[iLayer];
+      }
+    }
+    for (int iBarrel = 0; iBarrel < 3; iBarrel++) {
+      delete hCalibrationChipSum[iScan][iBarrel];
+      delete hCalibrationChipAverage[iScan][iBarrel];
+      delete hCalibrationChipCounts[iScan][iBarrel];
+
+      delete hCalibrationRMSChipSum[iScan][iBarrel];
+      delete hCalibrationRMSChipAverage[iScan][iBarrel];
+    }
+  }
+
+  for (int iBarrel = 0; iBarrel < 3; iBarrel++) {
+
+    delete hCalibrationThrNoiseChipSum[iBarrel];
+    delete hCalibrationThrNoiseChipAverage[iBarrel];
+
+    delete hCalibrationThrNoiseRMSChipSum[iBarrel];
+    delete hCalibrationThrNoiseRMSChipAverage[iBarrel];
+  }
+}
+
+void ITSThresholdCalibrationTask::initialize(o2::framework::InitContext& /*ctx*/)
+{
+
+  ILOG(Info, Support) << "initialize ITSThresholdCalibrationTask" << ENDM;
+
+  mRunNumberPath = mCustomParameters["runNumberPath"];
+
+  createAllHistos();
+  publishHistos();
+}
+
+void ITSThresholdCalibrationTask::startOfActivity(Activity& /*activity*/)
+{
+  ILOG(Info, Support) << "startOfActivity" << ENDM;
+}
+
+void ITSThresholdCalibrationTask::startOfCycle()
+{
+  ILOG(Info, Support) << "startOfCycle" << ENDM;
+}
+
+void ITSThresholdCalibrationTask::monitorData(o2::framework::ProcessingContext& ctx)
+{
+
+  ILOG(Info, Support) << "START DOING QC General" << ENDM;
+  const auto tunString = ctx.inputs().get<gsl::span<char>>("tunestring");
+  const auto runType = ctx.inputs().get<short int>("runtype");
+  const auto scanType = ctx.inputs().get<char>("scantype");
+
+  string inString(tunString.begin(), tunString.end());
+
+  Int_t iScan;
+  Double_t CalibrationValue;
+  if (scanType == 'V')
+    iScan = 0;
+  else if (scanType == 'I')
+    iScan = 1;
+  else if (scanType == 'T')
+    iScan = 2;
+
+  auto splitRes = split(inString, "Stave:");
+
+  for (auto StaveStr : splitRes) {
+    if (StaveStr.size() > 0) {
+      CalibrationResStruct result = CalibrationParser(StaveStr);
+
+      int currentStave = StaveBoundary[result.Layer] + result.Stave + 1;
+      int currentChip = result.ChipID + 1;
+      int iBarrel = getBarrel(result.Layer);
+
+      if (scanType == 'V') {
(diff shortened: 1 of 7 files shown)


Modules/ITS/src/ITSThresholdCalibrationTask.cxx
```diff
@@ -0,0 +1,456 @@
...
+
+  ILOG(Info, Support) << "START DOING QC General" << ENDM;
+  const auto tunString = ctx.inputs().get<gsl::span<char>>("tunestring");
+  const auto runType = ctx.inputs().get<short int>("runtype");
+  const auto scanType = ctx.inputs().get<char>("scantype");
+
+  string inString(tunString.begin(), tunString.end());
+
+  Int_t iScan;
+  Double_t CalibrationValue;
```
iravasen, line 103: according to coding conventions, can you rename it like ```calibrationValue```?

Modules/ITS/src/ITSThresholdCalibrationTask.cxx
```diff
@@ -0,0 +1,456 @@
...
+      int currentStave = StaveBoundary[result.Layer] + result.Stave + 1;
+      int currentChip = result.ChipID + 1;
+      int iBarrel = getBarrel(result.Layer);
+
+      if (scanType == 'V') {
+        CalibrationValue = result.VCASN;
+      } else if (scanType == 'I') {
+        CalibrationValue = result.ITHR;
+      }
+      if (scanType == 'T') {
```
iravasen, line 126: else if

Modules/ITS/src/ITSThresholdCalibrationTask.cxx
```diff
@@ -0,0 +1,456 @@
...
+    iScan = 2;
+
+  auto splitRes = split(inString, "Stave:");
+
+  for (auto StaveStr : splitRes) {
+    if (StaveStr.size() > 0) {
+      CalibrationResStruct result = CalibrationParser(StaveStr);
+
+      int currentStave = StaveBoundary[result.Layer] + result.Stave + 1;
+      int currentChip = result.ChipID + 1;
```
iravasen, line 118: If I'm not mistaken this will work only for IB when using this number to fill the histo. For IB this number goes from 1 to 9 which corresponds to the binning of the x axis of the corresponding histos, while for MB and OB you have to apply some convertion because you need a number from 1 to 112 (for MB) and from 1 to 196 (for OB). In my offline code I use the following but feel free to change (here I use already your naming for variables):
```
if(layer<5){//ML
        int chipbin = result.ChipID<7 ? (result.ChipID+1)+14*(result.HIC-1)+result.Hs*56 : result.ChipID+14*(result.HIC-1)+result.Hs*56; // this is already from 1 to 112
}
else { //OL
    int chipbin = result.ChipID<7 ? (result.ChipID+1)+14*(result.HIC-1)+result.Hs*98 : result.ChipID+14*(result.HIC-1)+result.Hs*98; //this is already from 1 to 196
} 
```

Modules/ITS/src/ITSThresholdCalibrationTask.cxx
```diff
@@ -0,0 +1,456 @@
...
+      int iBarrel = getBarrel(result.Layer);
+
+      if (scanType == 'V') {
+        CalibrationValue = result.VCASN;
+      } else if (scanType == 'I') {
+        CalibrationValue = result.ITHR;
+      }
+      if (scanType == 'T') {
+        CalibrationValue = result.THR;
+        hCalibrationThrNoiseChipSum[iBarrel]->SetBinContent(currentChip, currentStave, result.Noise + hCalibrationThrNoiseChipSum[iBarrel]->GetBinContent(currentChip, currentStave));
```
iravasen, line 128: You can just fill with result.Noise. When a chip arrives in input, it will not reappear again. In other words its noise value is unique.

Modules/ITS/src/ITSThresholdCalibrationTask.cxx
```diff
@@ -0,0 +1,456 @@
...
+        hCalibrationThrNoiseRMSChipSum[iBarrel]->SetBinContent(currentChip, currentStave, result.NoiseRMS + hCalibrationThrNoiseRMSChipSum[iBarrel]->GetBinContent(currentChip, currentStave));
```
iravasen, line 129: the same here, fill with only result.NoiseRMS

Modules/ITS/src/ITSThresholdCalibrationTask.cxx
```diff
@@ -0,0 +1,456 @@
...
+      }
+
+      hCalibrationChipSum[iScan][iBarrel]->SetBinContent(currentChip, currentStave, CalibrationValue + hCalibrationChipSum[iScan][iBarrel]->GetBinContent(currentChip, currentStave));
```
iravasen, line 132: same here

Modules/ITS/src/ITSThresholdCalibrationTask.cxx
```diff
@@ -0,0 +1,456 @@
...
+      hCalibrationRMSChipSum[iScan][iBarrel]->SetBinContent(currentChip, currentStave, result.RMS + hCalibrationRMSChipSum[iScan][iBarrel]->GetBinContent(currentChip, currentStave));
```
iravasen, line 133: same here

Modules/ITS/src/ITSThresholdCalibrationTask.cxx
```diff
@@ -0,0 +1,456 @@
...
+
+      hCalibrationChipCounts[iScan][iBarrel]->SetBinContent(currentChip, currentStave, hCalibrationChipCounts[iScan][iBarrel]->GetBinContent(currentChip, currentStave) + 1);
```
iravasen, line 135: This histo is to count how many times a chip is received? In principle every chip is received only once but I think it's a good idea to keep the plot so that in case of issue we can check if a chip is received twice (did not happen so far). Here you can simplify with:
```
histo->Fill(currentChip, currentStave)
```

Modules/ITS/src/ITSThresholdCalibrationTask.cxx
```diff
@@ -0,0 +1,456 @@
...
+
+  calculateAverages(iScan);
+
+  for (int iLayer; iLayer < 7; iLayer++) {
+    if (TotalStatus[iLayer] > 0)
+      hSuccessRate->SetBinContent(iLayer + 1, SuccessStatus[iLayer] / TotalStatus[iLayer]);
+  }
+}
+
+void ITSThresholdCalibrationTask::calculateAverages(Int_t iScan)
```
iravasen, line 151: You can remove this function for the same motivation explained above: the chip threshold/ITHR/VCASN is received only once.

Modules/ITS/src/ITSThresholdCalibrationTask.cxx
```diff
@@ -0,0 +1,456 @@
...
+
+      hCalibrationChipCounts[iScan][iBarrel]->SetBinContent(currentChip, currentStave, hCalibrationChipCounts[iScan][iBarrel]->GetBinContent(currentChip, currentStave) + 1);
+
+      if (result.status == 1)
+        SuccessStatus[result.Layer]++;
+      TotalStatus[result.Layer]++;
+    }
+  }
+
+  calculateAverages(iScan);
```
iravasen, line 143: can be removed.

Modules/ITS/src/ITSThresholdCalibrationTask.cxx
```diff
@@ -0,0 +1,456 @@
...
+{
+  /*
+    TString sScanTypes[3]= {"VCASN","ITHR","THR"};
+    TString sBarrelType[3]= {"IB","MB","OB"};
+    Int_t nChips[3]= {9,112, 196};
+    Int_t nStaves[3]={48,54,90};
+    Int_t nXmax[3]= {80,100,400};
+    TString sXtitles[3]={"DAC","DAC","e"};
+  */
+  for (int iScan = 0; iScan < 3; iScan++) {
```
iravasen, line 290: General comment for all plots: give "shifter friendly" titles to plot. For example: "Average chip thresholds for IB" or "Average chip ITHR for MB", ..... The histo names are fine.

Modules/ITS/src/ITSThresholdCalibrationTask.cxx
```diff
@@ -0,0 +1,456 @@
...
+
+void ITSThresholdCalibrationTask::publishHistos()
+{
+  for (unsigned int iObj = 0; iObj < mPublishedObjects.size(); iObj++) {
+    getObjectsManager()->startPublishing(mPublishedObjects.at(iObj));
+    ILOG(Info, Support) << " Object will be published: " << mPublishedObjects.at(iObj)->GetName() << ENDM;
+  }
+}
+
+std::vector<std::string> ITSThresholdCalibrationTask::split(std::string s, std::string delimiter)
```
iravasen, line 440: always according to coding convention I would rename this function as splitString or something similar.

iravasen:
Congratulations @IsakovAD , very good achievement!

Modules/ITS/src/ITSThresholdCalibrationTask.cxx
```diff
@@ -0,0 +1,456 @@
...
+    }
+  }
+  return result;
+}
+
+void ITSThresholdCalibrationTask::endOfCycle()
+{
+
+  std::ifstream runNumberFile(mRunNumberPath.c_str());
+  if (runNumberFile) {
```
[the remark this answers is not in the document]
IsakovAD: I see that in some tasks we do not fill metadata to the TH objects at all. 

For example, it is ignored in FhrTask, FeeTask, while we have metadata in the Cluster and Track tasks. Other Modules do not have an agreement as well.

Do we need to fill metadata or it can be skipped?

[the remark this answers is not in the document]
iravasen: You can remove this because even if in some tasks is still there, we don't use it anymore (i.e. in the json we are not setting anymore the runNumberPath variable). Now the run number is automatically retrieved by ECS logic.

iravasen:
@knopers8 @Barthelemy do you know how can I fix the issue with o2-dataflow-cs8? The problem does not seem to be related to this PR.

Barthelemy:
Must have been a temporary glitch.

Barthelemy (approved):

Barthelemy (approved):

iravasen (approved):

PR #1067, opened 2022-03-03
MCH: moved MergeableTH2Ratio class in common muon namespace

aferrero2707:


aferrero2707:
@Barthelemy @knopers8 this PR is needed in order to properly share the MCH mergeable histogram classes with MID. The changes are relatively simple, just the renaming on one namespace, so it should be simple to review. Could you have a look when you have the possibility?
Thanks!

Barthelemy (approved):

PR #1072, opened 2022-03-08
Allow to iterate on existing objects in QCDB which match an activity

knopers8:
it includes some preparatory work in the framework as well.


iravasen (approved):

Barthelemy (approved):
Very nice, thank you

PR #1073, opened 2022-03-08
Use MergeableTH2Ratio which is common with MCH

dstocco:
This avoids code duplication and makes maintenance easier.


dstocco:
For info @ValerieRamillien @aferrero2707

Barthelemy (approved):

PR #1084, opened 2022-03-14
Adapt ITS postprocessing to use timestamps and activities

knopers8:
This will allow you to use triggers which can iterate on all existing objects in QCDB which match certain criteria, e.g. all runs for apass2 of OCT data.
See the doc for more information:
https://github.com/AliceO2Group/QualityControl/blob/master/doc/PostProcessing.md#i-want-to-run-postprocessing-on-all-already-existing-objects-for-a-run


iravasen (approved):

iravasen:
@knopers8 thanks a lot! This is very useful and we could use it also to improve our post-processing tasks using the data from pilot beam.

PR #1088, opened 2022-03-14
Adapt FT0 postprocessing to use timestamps and activities

knopers8:
This will allow you to use triggers which can iterate on all existing objects in QCDB which match certain criteria, e.g. all runs for apass2 of OCT data.
See the doc for more information:
https://github.com/AliceO2Group/QualityControl/blob/master/doc/PostProcessing.md#i-want-to-run-postprocessing-on-all-already-existing-objects-for-a-run

@afurs @Kavaldrin @sbysiak


Barthelemy:
@TimoWilken Is the mac builder ok ? it seems stuck

Barthelemy (approved):

PR #1093, opened 2022-03-15
Port dictionary generation from O2.

ktf:


Barthelemy:
we actually use the one from O2. I am preparing a PR to remove this one. 

By the way, please notice that the mac builder is stuck.

PR #1104, opened 2022-03-20
MCH: improved pedestals task

aferrero2707:
This PR introduces significant improvements in the visualization of pedestals and noise in detector coordinates.
It also introduces a checker object that flags the pedestal data according to the fraction of channels with good pedestal values.


Barthelemy:
@TimoWilken @ktf the tests are timing out on o2-cs8 without reason. Could you have a look ? Thank you

ktf:
CCDB was down this morning. It should recover at some point.

Barthelemy (approved):

PR #1113, opened 2022-03-23
[FV0] Reenable FV0, adjust laser task to follow dataformat changes in O2

mslupeck:
Also unified the formatting between `DigitQCTask` and it's derivative: `DigitQCTaskLaser`


mslupeck:
Hi @sandor-lokos 
I've changed the formatting of `DigitQCTaksLaser` so that it is easier to compare it with `DigitQCTask` - I hope you don't mind.
But the only difference, I could see (except change of class name), is in line 374:
`// mHistAverageTimeA->Fill(digit.mTriggers.timeA);`
This line is not commented in `DigitQCTask`. Is it intentional?

Barthelemy (approved):

sandor-lokos:
Hi @mslupeck ,
I don't mind at all. In fact, I didn't even touch that file, I created the custom TH1ReductorLaser module and only copy the already existing DigitQcTask module to DigitQcTaskLaser to separate the latter from the former and leave open the way for further improvements.

PR #1127, opened 2022-03-30
MFT: Add trending for the readout task.

tomas-herman:
Adding trending for the readout task.


tomas-herman:
Added a commit removing "qc" from the object path in trending.

Barthelemy (approved):

PR #1138, opened 2022-04-08
[QC-743] Trending on a range of timestamps

knopers8:
This allows to restrain the timestamp range in postprocessing to certain start and end values.
Also, once GUI is ready, we can use the same fields to store time validity of Monitor Objects and translate it to "Valid-From" and "Valid-Until" values in QCDB.

### DIFF
--- a/Framework/include/QualityControl/Activity.h
+++ b/Framework/include/QualityControl/Activity.h
@@ -19,6 +19,8 @@
 
 #include <map>
 #include <string>
+#include <iosfwd>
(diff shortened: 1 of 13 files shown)


Barthelemy:
Have you forgotten to commit a file ?

knopers8:
yes

Barthelemy (commented):
Makes sense, very clear.

Barthelemy (approved):

PR #1144, opened 2022-04-13
MCH: handle missing ORBITS DPL message

aferrero2707:
The commit adds a check for the availability of the DPL message with the list of orbits from each CRU link.
If the message is missing, the orbits counters are increased by the default number of orbits in the time-frame as given by `HBFUtils::getNOrbitsPerTF()`.

### DIFF
--- a/Modules/MUON/MCH/src/PhysicsTaskDigits.cxx
+++ b/Modules/MUON/MCH/src/PhysicsTaskDigits.cxx
@@ -34,6 +34,7 @@
 #include "QualityControl/QcInfoLogger.h"
 #include <Framework/InputRecord.h>
 #include <CommonConstants/LHCConstants.h>
+#include <DetectorsRaw/HBFUtils.h>
 
 using namespace std;
 using namespace o2::mch::raw;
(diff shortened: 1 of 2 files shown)


Modules/MUON/MCH/src/PhysicsTaskDigits.cxx
```diff
@@ -186,18 +187,28 @@ void PhysicsTaskDigits::startOfCycle()
...
+      hasOrbits = true;
+    }
   }
 
-  for (auto& orb : orbits) {
-    storeOrbit(orb);
+  if (hasOrbits) {
+    auto orbits = ctx.inputs().get<gsl::span<uint64_t>>("orbits");
+    if (orbits.empty()) {
+      ILOG(Info, Support) << "WARNING: empty orbits vector" << AliceO2::InfoLogger::InfoLogger::endm;
```
Barthelemy, line 200: In general, it is not recommended to publish log messages in `monitorData` because it can easily flood the infologger in case the condition is true. One way to avoid that is to use a feature of infologger allowing to limit the number of messages per period of time: 
```
    static AliceO2::InfoLogger::InfoLogger::AutoMuteToken msgLimit(LogWarningSupport, 1, 600); // send it once every 10 minutes
    string msg = "object " + path + " is bigger than the maximum allowed size (" + to_string(mMaxObjectSize) + "B) - skipped";
    ILOG_INST.log(msgLimit, "%s", msg.c_str());
```

aferrero2707: @Barthelemy thanks for the code snippet. While the empty orbits message is an abnormal condition that should never occur, I agree that a protection from messages flood is a good idea.
Should the code go in the `configure()` method of the task?

aferrero2707: @Barthelemy thanks! I have modified the code as you suggest.

aferrero2707:
@Barthelemy @knopers8 The errors in the macOS CI build seem to be completely un-related with this PR. Should I force a full CI rebuild, or do you prefer to force-merge if you are happy with the changes? Thanks!

Barthelemy:
I am force merging

Barthelemy (approved):

PR #1153, opened 2022-04-22
Included check of trends comparing to expected physics value

lauraser:


Modules/TPC/src/CheckOfTrendings.cxx
```diff
@@ -29,41 +30,127 @@
 namespace o2::quality_control_modules::tpc
 {
 
-void CheckOfTrendings::configure() {}
+void CheckOfTrendings::configure()
+{
+  if (auto param = mCustomParameters.find("chooseCheckMeanOrExpectedPhysicsValueOrBoth"); param != mCustomParameters.end()) {
+    checkChoice = param->second.c_str();
+    if (checkChoice != "mean" && checkChoice != "ExpextedPhysicsValue" && checkChoice != "both") {
```
bulukutlu, line 37: There are multiple instances of misspelt **ExpextedPhysicsValue** in this file. I think it would lead to problems as the setting in the json file would not be taken into account.

Modules/TPC/include/TPC/CheckOfTrendings.h
```diff
@@ -41,6 +41,13 @@ class CheckOfTrendings : public o2::quality_control::checker::CheckInterface
 
  private:
   ClassDefOverride(CheckOfTrendings, 2);
+  std::string checkChoice;
+  float expectedPhysicsValue;
+  float nSigmaExpectedPhysicsValue;
+  float nSigmaBadExpectedPhysicsValue;
+  float nSigmaMean;
+  float nSigmaBadMean;
+  int pointToTakeForExpectedValueCheck;
```
wiechula, line 50: Please follow naming conventions. Data members should start with `m` and camel case.

Modules/TPC/src/CheckOfTrendings.cxx
wiechula, line 37: Perhaps you should make all these strings (or at least the ones used more than once) `static constexpr std::string_view` here (or in the header if used also in other places) and then use it everywhere to avoid copy'n'paste errors

Modules/TPC/src/CheckOfTrendings.cxx
```diff
@@ -29,41 +30,127 @@
...
+    const std::vector<double> v(yValues + NBins - pointNumber, yValues + NBins);
+    const double sum = std::accumulate(v.begin(), v.end(), 0.0);
+    const double meanFull = sum / v.size();
+
+    std::vector<double> diff(v.size());
+    std::transform(v.begin(), v.end(), diff.begin(), [meanFull](double x) { return x - meanFull; });
+    const double sq_sum = std::inner_product(diff.begin(), diff.end(), diff.begin(), 0.0);
+    const double stdevFull = std::sqrt(sq_sum / v.size());
+
+    if (fabs(meanFull - expectedPhysicsValue) > nSigmaBadExpectedPhysicsValue * stdevFull) {
```
wiechula, line 138: please use `std::abs` instead of `fabs` everywhere.

wiechula (changes requested):

Modules/TPC/src/CheckOfTrendings.cxx
lauraser: I did it only for strings which are used more than twice

Modules/TPC/include/TPC/CheckOfTrendings.h
```diff
@@ -41,6 +41,17 @@ class CheckOfTrendings : public o2::quality_control::checker::CheckInterface
...
+  float mExpectedPhysicsValue;
+  float mNSigmaExpectedPhysicsValue;
+  float mNSigmaBadExpectedPhysicsValue;
+  float mNSigmaMean;
+  float mNSigmaBadMean;
+  int mPointToTakeForExpectedValueCheck;
+  int mPointToTakeForMeanCheck;
+  static constexpr std::string_view mCheckChoiceMean = "Mean";
+  static constexpr std::string_view mCheckChoiceExpectedPhysicsValue = "ExpectedPhysicsValue";
+  static constexpr std::string_view mCheckChoiceBoth = "Both";
```
wiechula, line 54:
```suggestion
  static constexpr std::string_view CheckChoiceMean = "Mean";
  static constexpr std::string_view CheckChoiceExpectedPhysicsValue = "ExpectedPhysicsValue";
  static constexpr std::string_view CheckChoiceBoth = "Both";
```
See naming convention for `constexpr` https://rawgit.com/AliceO2Group/CodingGuidelines/master/naming_formatting.html?showone=Variable_Names#Variable_Names

Modules/TPC/src/CheckOfTrendings.cxx
```diff
@@ -29,41 +30,138 @@
 namespace o2::quality_control_modules::tpc
 {
 
-void CheckOfTrendings::configure() {}
+void CheckOfTrendings::configure()
+{
+  if (auto param = mCustomParameters.find("chooseCheckMeanOrExpectedPhysicsValueOrBoth"); param != mCustomParameters.end()) {
```
wiechula, line 35:
```suggestion
  if (const auto param = mCustomParameters.find("chooseCheckMeanOrExpectedPhysicsValueOrBoth"); param != mCustomParameters.end()) {
```
I think all of the similar lines below can also be declared `const`

wiechula (changes requested):

lauraser:
@wiechula Thanks for the comments! I implemented them in the code now.

wiechula (approved):

wiechula:
@Barthelemy , @knopers8 , the macOS failure seems to be more general and not related to this PR. Can this be merged?

Barthelemy:
@wiechula yes, of course

Barthelemy:
@TimoWilken @ktf could you check the mac builder ? thank you

PR #1155, opened 2022-04-26
[QC-773] Add [task/check/aggregator]Name in configuration files

knopers8:
This adds a corresponding actor name to  the task, check, aggregator and external task config structures.
The root keys are treated as IDs and if names are missing, IDs are used. This way, we can let tasks with
 the same names live together as long as they have unique IDs. This is needed for QC on EPN, where all QC
is ran within the same big workflow and name collisions started to occur.


Barthelemy (approved):

Barthelemy:
I think that the multinode test is legitimately failing

PR #1156, opened 2022-04-26
TRD redo connection to ccdb

bazinski:
This fixes the crashes seen online.

It removes double deletion locally, so should be good for online.

### DIFF
--- a/Modules/TRD/include/TRD/TrackletsTask.h
+++ b/Modules/TRD/include/TRD/TrackletsTask.h
@@ -18,6 +18,7 @@
 #define QC_MODULE_TRD_TRDTRACKLETSTASK_H
 
 #include "QualityControl/TaskInterface.h"
+#include "QualityControl/DatabaseInterface.h"
 #include "DataFormatsTRD/NoiseCalibration.h"
 
 class TH1F;
@@ -47,7 +48,7 @@ class TrackletsTask final : public TaskInterface
   void endOfActivity(Activity& activity) override;
   void reset() override;
   void buildHistograms();
-  void connectToCCDB();
+  void retrieveCCDBSettings();
   void drawLinesMCM(TH2F* histo);
 
  private:
--- a/Modules/TRD/src/PulseHeight.cxx
+++ b/Modules/TRD/src/PulseHeight.cxx
@@ -41,11 +41,9 @@ PulseHeight::~PulseHeight()
 {
 }
 
-void PulseHeight::connectCCDB()
+void PulseHeight::retrieveCCDBSettings()
 {
-  auto& ccdbmgr = o2::ccdb::BasicCCDBManager::instance();
-  // ccdbmgr.setURL("http://localhost:8080");
-  mNoiseMap.reset(ccdbmgr.get<o2::trd::NoiseStatusMCM>("/TRD/Calib/NoiseMapMCM"));
+  mNoiseMap.reset(reinterpret_cast<o2::trd::NoiseStatusMCM*>(retrieveCondition("/TRD/Calib/NoiseMapMCM")));
 }
 
 void PulseHeight::buildHistograms()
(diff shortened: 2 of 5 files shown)


bazinski:
@tdietel @martenole  could you guys approve this soonish?
Apparently qc will get tagged tomorrow.

martenole:
Sorry I am afraid I don't really understand the changes. But I am also no codeowner of the TRD code in QC, so my approval does not really matter.
I thought the double deletion occurred at the EOR only. Don't the changes here only affect the initialization phase?

bazinski:
@martenole I thought so as well, but as soon as I commented the ccdb connection out the problem went away.
one gets a double deletion error and then it takes ages to end. 
Was testing on a single time frame, and the workflow (raw-tf | datareader | qc | o2-dpl-run) would just hang for ages, never waited for it to finish but would come back with an error number after killing. 

This now uses qc's connection to ccdb, not a manual one.
I know you and also @tdietel are not code owners, but figured @Barthelemy and or @knopers8  would query trd so in the interests of getting this through quickly, I figured I would preempt that ;-)

martenole:
Hm ok. When using the CCDB manager I think one should anyhow always use raw pointers. Probably exactly to avoid these problems :)
If the QC's connection works differently then probably this is ok. Otherwise I would just use (const) pointers instead of shared_ptr.

tdietel (approved):
Looks like this is a good shot at fixing the double free bug we have in the TRD QC at the moment. I suggest to merge this before the next tag. Good work, Sean.

Modules/TRD/include/TRD/TrackletsTask.h
```diff
@@ -66,6 +67,7 @@ class TrackletsTask final : public TaskInterface
   std::shared_ptr<TH1F> mTrackletPositionRawn = nullptr;
   std::shared_ptr<TH1F> mTrackletsPerEventn = nullptr;
   std::shared_ptr<o2::trd::NoiseStatusMCM> mNoiseMap = nullptr;
+  o2::quality_control::repository::DatabaseInterface* mDatabase = nullptr;
```
Barthelemy, line 70: why do you need this variable ?

Modules/TRD/src/PulseHeight.cxx
```diff
@@ -39,11 +39,11 @@ PulseHeight::~PulseHeight()
...
 }
 
-void PulseHeight::connectCCDB()
+void PulseHeight::retrieveCCDBSettings()
 {
-  auto& ccdbmgr = o2::ccdb::BasicCCDBManager::instance();
-  //ccdbmgr.setURL("http://localhost:8080");
-  mNoiseMap.reset(ccdbmgr.get<o2::trd::NoiseStatusMCM>("/TRD/Calib/NoiseMapMCM"));
+  //auto& ccdbmgr = o2::ccdb::BasicCCDBManager::instance();
+  //mNoiseMap.reset(ccdbmgr.get<o2::trd::NoiseStatusMCM>("/TRD/Calib/NoiseMapMCM"));
```
Barthelemy, line 45: The coding guidelines recommend removing the code instead of commenting it out.

Modules/TRD/src/PulseHeight.cxx
```diff
@@ -39,11 +39,11 @@ PulseHeight::~PulseHeight()
...
+  mNoiseMap.reset((o2::trd::NoiseStatusMCM*)retrieveCondition("/TRD/Calib/NoiseMapMCM"));
```
Barthelemy, line 46: We forbid c-style casts in O2. Please use `reinterpret_cast`, `static_cast` or `dynamic_cast` depending on the case.

Modules/TRD/src/TrackletsTask.cxx
```diff
@@ -61,12 +61,13 @@ void TrackletsTask::drawLinesMCM(TH2F* histo)
   }
 }
 
-void TrackletsTask::connectToCCDB()
+void TrackletsTask::retrieveCCDBSettings()
```
Barthelemy, line 64: This is a duplicate. Please move it to a header accessible by both classes.

Barthelemy (changes requested):

bazinski:
The macOS errors are missing libraries and some stuff not related to this code.

Barthelemy:
yes, this is a problem with the builder.

Barthelemy (approved):

PR #1163, opened 2022-05-03
ITS: Reintroduced local geometry path

iravasen:
Important for P2 operations

### DIFF
--- a/Modules/ITS/include/ITS/ITSClusterTask.h
+++ b/Modules/ITS/include/ITS/ITSClusterTask.h
@@ -101,6 +101,7 @@ class ITSClusterTask : public TaskInterface
   int mNRofsMonitor = 0;
   int nBCbins;
   long int mTimestamp;
+  std::string mGeomPath;
 
   const int mNStaves[7] = { 12, 16, 20, 24, 30, 42, 48 };
   const int mNHicPerStave[NLayer] = { 1, 1, 1, 8, 8, 14, 14 };
--- a/Modules/ITS/include/ITS/ITSFhrTask.h
+++ b/Modules/ITS/include/ITS/ITSFhrTask.h
@@ -172,11 +172,9 @@ class ITSFhrTask final : public TaskInterface
   TH2I* mChipStaveEventHitCheck[7];
   TH1D* mOccupancyPlot[7];
 
-  std::string mRunNumberPath;
-  std::string mRunNumber = "000000";
-
   // Geometry decoder
   o2::its::GeometryTGeo* mGeom;
+  std::string mGeomPath;
 };
 } // namespace o2::quality_control_modules::its
 
--- a/Modules/ITS/include/ITS/ITSNoisyPixelTask.h
+++ b/Modules/ITS/include/ITS/ITSNoisyPixelTask.h
@@ -99,6 +99,7 @@ class ITSNoisyPixelTask : public TaskInterface
   int mEnableLayers[7];
   o2::itsmft::TopologyDictionary* mDict;
   o2::its::GeometryTGeo* mGeom;
+  std::string mGeomPath;
   long int mTimestamp;
 };
 } //  namespace o2::quality_control_modules::its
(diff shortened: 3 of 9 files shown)


iravasen:
@knopers8 If I understand correctly, in `o2-qc` workflow there is not possibility to add something like `--configKeyValues "NameConf.mDirGeom=<path>;NameConf.mDirGRP=<path>;"` so to set the paths of the geometry and grp file. Right? If true, this motivates the need of this PR.

davidrohr:
@knopers8 @Barthelemy : I don't understand why we cannot parse the configKeyValues in o2-qc? I think in general we should add support for this.

@iravasen : This PR is fine with me as a workaround. But just for my understanding, if I put
```
"geomPath" : ""
```
with an empty string in the json file, the old behavior will be maintained?

iravasen:
@davidrohr yes, If I put `geomPath: ""`, then I see the following when running the workflow:
```
[24727:qc-task-ITS-ITSClusterTask]: [11:36:18][INFO] Loading geometry from o2sim_geometry-aligned.root
[24727:qc-task-ITS-ITSClusterTask]: Error in <TFile::TFile>: file /home/its/iravasen/workdir/o2sim_geometry-aligned.root does not exist
[24727:qc-task-ITS-ITSClusterTask]: [11:36:18][FATAL] Failed to open file o2sim_geometry-aligned.root
```
it tries to find the geom file in the path where I actually run the workflow and it fails because I put the geom file in another path. This is the same behaviour I see with the older code (i.e. the code which is now at P2).

knopers8:
> @knopers8 @Barthelemy : I don't understand why we cannot parse the configKeyValues in o2-qc? I think in general we should add support for this.
> 

I am not very happy of adding yet another way of configuration, but we could consider it. Would you know an example how it is done?

davidrohr:
> > @knopers8 @Barthelemy : I don't understand why we cannot parse the configKeyValues in o2-qc? I think in general we should add support for this.
> 
> I am not very happy of adding yet another way of configuration, but we could consider it. Would you know an example how it is done?

Well, configKeyValues are used in many places in O2. If you cannot configure them via the command line, all QC tasks using them are basically stuck to their default values.

The usage in O2 is quite trivial. They must be added to the possible command line options, e.g.
https://github.com/AliceO2Group/AliceO2/blob/dev/GPU/Workflow/src/gpu-reco-workflow.cxx#L53
and then configured from there:
https://github.com/AliceO2Group/AliceO2/blob/dev/GPU/Workflow/src/gpu-reco-workflow.cxx#L116

There are also options to store to file what was used and reload that, but perhaps not needed for QC.

One relevant thing though: The libraries with the objects that are configured must be loaded when the parsing runs, otherwise it will not find their ROOT dictionaries and fail.

knopers8:
Thank you for the example, I also understand the need.

> One relevant thing though: The libraries with the objects that are configured must be loaded when the parsing runs, otherwise it will not find their ROOT dictionaries and fail.

This is actually good to know. We load libraries for corresponding tasks quite late, during the DPL's `init()`. Still, I guess we can store the argument somewhere until we load the user modules and then execute this:
```
  o2::conf::ConfigurableParam::updateFromString(cfgc.options().get<std::string>("configKeyValues"));
```

iravasen:
@knopers8 thanks for adding this to qc. About this PR, I see two tests failing but it seems again a glitch. I would still merge this PR by tomorrow at noon so that next week we can test at P2 (at least for FLP).

davidrohr (approved):
With https://github.com/AliceO2Group/O2DPG/pull/363 the Full System Test with QC passes.

Barthelemy:
the mac builder failure is incorrect I think

Barthelemy:
As I did not follow up this item, I let you @knopers8 force merge it.

davidrohr:
Can we force-merge this and get it into the next tag? I tested manually on the EPNs and it works.

JianLIUhep:
@knopers8 @Barthelemy, would it be possible to patch QC with this PR on FLPs at P2? It would be helpful if we can revalidate the QC on FLP this week asap since we will resume the global commissioning with detector next week. Thank you.

Barthelemy:
@JianLIUhep Hi Jian, I really don't know if that's feasible. I can produce a patched version (1.56.1) but I don't know if it is possible to install it on your flps. @vascobarroso , is there a procedure for that ?

PR #1164, opened 2022-05-03
TRD fix qc crashes.

bazinski:
- revert the connection to ccdb to be via ccdbmgr
- make the noisymcm a raw ptr.
- down the line the timestamp should be customisable from the json file, commented code left in for that.

This now works on my desktop and the epn, so hopefully good online.


martenole (approved):
Looks good to me

Barthelemy:
the mac builder is broken, if you remove the automatic merge I will force merge.

davidrohr (approved):
Tried with Full System Test with QC and it solves the problem of the crash

@Barthelemy @knopers8 : Could we get this into the next QC tag?

bazinski:
@Barthelemy I assume the request to remove the automatic merge is not to me? I understand its in the settings of the repo.

Barthelemy:
yes it will be in the next release

PR #1167, opened 2022-05-06
new ClustQcTask and TracksQcTask; add BCCounts histo for DigitQcTask

ValerieRamillien:


Barthelemy (approved):

ValerieRamillien:
I don't see errors in logs from MID part, perhaps the problem is more general and not related to this PR.

Barthelemy:
I agree. @TimoWilken @ktf There is something breaking the builds.

PR #1173, opened 2022-05-09
Adaption of ITS FHR checks to the framework guidelines 

zz951:
New itsFhrCheck with addMetaData and so on.


Modules/ITS/include/ITS/ITSFhrTask.h
```diff
@@ -172,12 +172,8 @@ class ITSFhrTask final : public TaskInterface
   TH2I* mChipStaveEventHitCheck[7];
   TH1D* mOccupancyPlot[7];
 
-  std::string mRunNumberPath;
-  std::string mRunNumber = "000000";
-
   // Geometry decoder
   o2::its::GeometryTGeo* mGeom;
-  std::string mGeomPath;
```
iravasen, line 180: @zz951 can you please re-introduce the geomPath both here and in the json? Please take a look at what is in the QC master:   https://github.com/AliceO2Group/QualityControl/blob/1a77eacf17f13d1e403391b9f3a4a23282c0c192/Modules/ITS/src/ITSFhrTask.cxx#L96

Modules/ITS/itsFhr.json
```diff
@@ -40,7 +40,6 @@
 		    "HitNumberCut": "512",
                     "GetTFFromBinding": "0",
                     "decoderThreads": "8",
-                    "geomPath": "/tmp/qc_tmp/o2sim",
```
iravasen, line 43: please reintroduce the  `geomPath` in the json.

Modules/ITS/src/ITSFhrCheck.cxx
```diff
@@ -33,77 +33,68 @@ void ITSFhrCheck::configure() {}
...
+          } else if ((h->GetBinContent(ibin) / (double)mNPixelPerStave[1]) > 0.00005 && strcmp(result.getMetadata("Noi_Pix").c_str(), "good") == 0) {
+            result.updateMetadata("Noi_Pix", "medium");
           }
         } else {
           if ((h->GetBinContent(ibin) / (double)mNPixelPerStave[2]) > 0.0001) {
-            result = (result.getLevel() % 100) + 100;
-          } else if ((h->GetBinContent(ibin) / (double)mNPixelPerStave[2]) > 0.00005 && (result.getLevel() / 100 < 1)) {
-            result = (result.getLevel() % 100) + 200;
+            result.updateMetadata("Noi_Pix", "bad");
+          } else if ((h->GetBinContent(ibin) / (double)mNPixelPerStave[2]) > 0.00005 && strcmp(result.getMetadata("Noi_Pix").c_str(), "good") == 0) {
```
iravasen, line 81: do you really need the `strcmp` in this `else if` condition? You enter this `else if` only if the fraction of noisy pixels is < 0.0001. I might miss something, please let me know.

Modules/ITS/src/ITSFhrTask.cxx
```diff
@@ -38,62 +38,62 @@ ITSFhrTask::ITSFhrTask()
 
 ITSFhrTask::~ITSFhrTask()
 {
```
iravasen, line 40: VERY IMPORTANT: please remove any local change from the FHR task. I would not modify the version at P2.

iravasen: of course keep the latest modifications about RunNumberPath and geometry path. Keep maybe a copy of the modified version. Then do `git checkout <filename>` to remove all changes. Then merge the changes from the QC master.

iravasen:
In the json, in the part related to FhrCheck can you please put `"policy": "OnEachSeparately",` (i.e. replace `OnAny` with `OnEachSeparately`).

Modules/ITS/src/ITSFhrTask.cxx
zz951: ok, I got it.

iravasen:
Other trivial comment that came to my mind: do not forget `histo->GetXaxis()->SetLabelSize(0.02)` for L3 and L4 TH2. This improves visibility.

iravasen:
Other trivial comment that came to my mind: do not forget `histo->GetXaxis()->SetLabelSize(0.02)` for L3 and L4 TH2. This improves visibility.

iravasen:
@zz951 all looks fine apart the ITSFhrTask.cxx. Too many changes which I cannot explain and which might be very dangerous. The only change that should appear in this file is the one about the bin label size, isn't it? If this is the case, please take the master version of the task, add the label size modification and push it to your branch.

iravasen (approved):

PR #1177, opened 2022-05-13
Correct efficiency calculation

IsakovAD:
This PR should solve issues with the tracking efficiency observed with the O2DPG workflow:

1) Ratio of histograms replaced by TEfficiency object to have correct merging
2) Track duplicated are removed from efficiency calculation, but stored in separate histograms

### DIFF
--- a/Modules/ITS/include/ITS/ITSTrackSimTask.h
+++ b/Modules/ITS/include/ITS/ITSTrackSimTask.h
@@ -28,7 +28,7 @@
 #include "Field/MagneticField.h"
 #include "TGeoGlobalMagField.h"
 #include "DataFormatsITS/TrackITS.h"
-
+#include <TEfficiency.h>
 class TH1D;
 class TH2D;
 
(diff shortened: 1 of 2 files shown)


iravasen:
@IsakovAD thanks a lot for this PR! All looks good to me. I have 1 question: will you add also the histograms that will contain the efficiency calculated after the merging of num and den? Does this produce comparable results as for the TEfficiency? Thanks.

IsakovAD:
@iravasen, I have tested this feature offline, and the TEfficiency object correctly calculates ratios even after merging. The trick is that the TEfficiency object keeps two histograms (den and num) and shows their ratio. Merging of TEff1 and TEff2 is performed by a sum of den1+den2, num1+num2. Essentially this is the same approach suggested by  Petr with the external merger class. So I suggest to have the TEfficiency object and not introduce additional code for merging.

iravasen:
@IsakovAD thanks a lot for the explanation, now it's very clear. Approving PR and enabling auto-merge.

iravasen:
@Barthelemy @knopers8 it seems that there is again an issue with the macOS test. If this is confirmed, can you please force-merge?

PR #1178, opened 2022-05-13
[QC-783] Fix service discovery disconnection

Barthelemy:


Barthelemy:
@knopers8 could you have a look at this PR please ?

knopers8 (approved):

Framework/src/ServiceDiscovery.cxx
```diff
@@ -120,14 +120,19 @@ void ServiceDiscovery::runHealthServer(unsigned int port)
...
       });
       tcp::socket socket(io_service);
       acceptor.async_accept(socket, [&](boost::system::error_code ec) {
-        if (!ec)
+        if (ec.failed()) {
+          threadInfoLogger << "async_accept error: " << ec.message() << ENDM;
           timer.cancel();
+        }
       });
+      io_service.run();
```
awegrzyn, line 135: This need to be run once per process. Maybe it was removed from QC/DPL code at some point? Could you verify?

awegrzyn: It's seems reasonable to have it, do you have any stack traces? Or the only behaviour was that process stopped responding on Consul health checks?

Barthelemy: The only behaviour that I can see is disappearance of the service in Consul.

PR #1185, opened 2022-05-17
MCH: remove stat boxes and improve detector contour drawing

aferrero2707:
This PR introduces a call to `object->SetStats(0)` for all the histograms for which we want the stats box to be hidden.

It also moves the drawing of the detector contours into separate functions, which are now called in the `beautify()` method of the checkers. This way the contours do not need to be stored in the histograms upon creation, and there is no more need to copy copy around the histograms' list of functions when merging.

This solves the memory leak issue that was introduced in the `beautify()` method of the `MergeableTH2Ratio` class. The method is now removed since it is no longer needed.

### DIFF
--- a/Modules/MUON/MCH/include/MCH/PedestalsTask.h
+++ b/Modules/MUON/MCH/include/MCH/PedestalsTask.h
@@ -52,13 +52,24 @@ class PedestalsTask final : public TaskInterface
   void reset() override;
 
  private:
+  template <typename T>
+  void publishObject(T* histo, std::string drawOption, bool statBox)
+  {
+    histo->SetOption(drawOption.c_str());
+    if (!statBox) {
+      histo->SetStats(0);
+    }
+    mAllHistograms.push_back(histo);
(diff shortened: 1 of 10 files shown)


aferrero2707:
@aphecetche @Barthelemy @knopers8 this PR contains some additional clean-up of the mergeable histograms code, as well as the systematic removal of the stat boxes for all 2D histograms and part of the 1D ones.

aferrero2707:
@aphecetche @Barthelemy @knopers8 this PR fixes the setting of default drawing options for the plots, and removes the stat boxes for all 2D histograms and part of the 1D ones. Could you please have a look, in particular to the histogram configuration part (see for example https://github.com/AliceMCH/QualityControl/blob/mch-remove-stat-boxes/Modules/MUON/MCH/include/MCH/PhysicsTaskDigits.h#L74)? Thanks!

Barthelemy (approved):

Barthelemy:
@aphecetche ok to merge ?

aphecetche:
Ok for me

PR #1188, opened 2022-05-18
Add a switch to turn on the fix for the online mode

Barthelemy:
Based on the env var QC_TEST_FIX_ONLINE


knopers8 (approved):
Isn't the convention to set a variable to `1` or even to any non-empty string? I am fine with how it is now though.

PR #1201, opened 2022-05-20
FT0: use double for the pi constant

knopers8:


knopers8:
@sandor-lokos

knopers8:
force-merging to unblock the release

PR #1205, opened 2022-05-24
MCH: do not use the deprecated method

knopers8:


knopers8:
@aferrero2707 @aphecetche This is an attempt to fix a deprecation warning I noticed, please let me know if it is fine with you

aferrero2707:
Thanks @knopers8! I completely overlooked this one, sorry... I was implementing the same fix in my next PR, but yours came before.

PR #1206, opened 2022-05-25
[FDD] Added the histograms of amplitude distribution for the coincidence of front and back layer of the FDD

arvindkhuntia:


arvindkhuntia:
Hi @Barthelemy, 
Could you please let me know, what is the error? I don't see anything related to FDD.

Thanks in advance.
With best regards,
Arvind

PR #1214, opened 2022-05-27
Add single variable for the histoname

njacazio:


njacazio:
@ercolessi could you also have a look, especially to the noise part

Barthelemy (approved):

PR #1219, opened 2022-05-30
fix for TOF ME QC to load TPC tracks even if TPC-only ME not required

noferini:


njacazio (approved):
Thanks Francesco!

noferini:
macOS fails with
ld: library not found for -lcrypto.44
clang: error: linker command failed with exit code 1 (use -v to see invocation)
I saw the same error in another PR. It seems a library problem, in any case not related to this PR.

njacazio:
I agree with you, I would ask @knopers8 or @Barthelemy to merge if they also agree

noferini:
Can the PR be merged? It is blocking another PR in O2DPG.

PR #1221, opened 2022-05-31
[QC-798] Add a check to ensure entries increase every cycle

Barthelemy:


Modules/Common/src/IncreasingEntries.cxx
```diff
@@ -0,0 +1,75 @@
...
+    if (histo == nullptr) {
+      continue;
+    }
+
+    double previousNumberEntries = mLastEntries.count(moName) > 0 ? mLastEntries.at(moName) : 0;
+    double currentNumberEntries = histo->GetEntries();
+
+    if (previousNumberEntries == currentNumberEntries) {
+      result = Quality::Bad;
+      result.addReason(FlagReasonFactory::Unknown(),
```
knopers8, line 58: `NoDetectorData` could have been used perhaps

Modules/Common/src/IncreasingEntries.cxx
```diff
@@ -0,0 +1,75 @@
...
+// This software is distributed under the terms of the GNU General Public
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+///
+/// \file   IncreasingEntries.cxx
+/// \author My Name
```
knopers8, line 14: to be filled on the next occasion

PR #1226, opened 2022-06-01
Add histogram for noisy flagged channels

njacazio:


njacazio:
Hi, I propose that we add a function to parse boolean parmeters, if you agree I will use it also in other QC tasks e.g. asynch QC task. So that we have a single point of failure and we don't duplicate code. Is this ok? Alternatively I can add a header in the TOF path, let me know!!

PR #1229, opened 2022-06-02
Fix ITS Fhr task bugs and issue in THnSparse filling with large events

iravasen:
Fixes:
- Fill noisy pix plot (TH2Poly) only if reasonable statistics is reached (1M events) and do it for the next 10000 events. The cut on the occupancy still allow to cut very high in case of pp or Pb-Pb collisions. 
- Dead chip maps filled with 0s: now filled with 1s. 
- THnSparse filled only up to a selectable (from json) number of events. Important for Pb-Pb, pp collisions.


Barthelemy (approved):

Barthelemy:
Given the successes in some tests, and the need to release, I merge it.

PR #1231, opened 2022-06-02
TOF: Correct loading of geometry

sawenzel:
To load a geometry, we need to give the simulation prefix
and not the complete filename.
The geometry manager will construct the filename itself.
(Unfortunately the interface definition/documentation was
not precise in this regard.)

### DIFF
--- a/Modules/PID/include/PID/TaskFT0TOF.h
+++ b/Modules/PID/include/PID/TaskFT0TOF.h
@@ -112,7 +112,7 @@ class TaskFT0TOF final : public TaskInterface
   float mMinDCAtoBeamPipeCut = 100.f;
   float mMinDCAtoBeamPipeCutY = 10.f;
   std::string mGRPFileName = "o2sim_grp.root";
-  std::string mGeomFileName = "o2sim_geometry.root";
+  std::string mSimPrefix = "o2sim";
   float mBz = 0; ///< nominal Bz
   int mTF = -1;  // to count the number of processed TFs
   const float cinv = 33.35641;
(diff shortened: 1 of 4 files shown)


sawenzel:
@noferini : This fixes a problem with the async_reco script for me. After this it should run fine in the nightlies.

noferini:
Thanks, I'm checking it in our case

chiarazampolli:
Hi @sawenzel , 

Could you please clarify how one should set the geometry? Maybe there are other places in which it is wrong. E.g., do we need the same here? https://github.com/AliceO2Group/AliceO2/blob/943ff7c1b3d73aeec2e91292c1468909965de605/Detectors/GlobalTracking/src/MatchITSTPCQC.cxx#L134
Note that in that task the filename is passed from the json (it should have been done like this also for TOF, probably I just forgot to add parsing the option). 

Chiara

sawenzel:
@chiarazampolli : So yes, I tried to run async_reco on alibi with data and it failed during TOF QC in geometry loading.
The present way proposed here will always work (and definitely fixes a problem for the moment).

See also discussion in https://github.com/AliceO2Group/AliceO2/pull/9063

chiarazampolli:
Ciao @sawenzel , 

Ok, my suspicion is that you were not taking the correct geometry file from the input files of the reco job, which was updated not very long ago. 

Chiara

sawenzel:
For the record, with latest tag of QC and dev of O2, I see:
```
[29013:qc-task-TOF-TaskFT0TOF]: Error in <TFile::TFile>: file /tmp/aliperf/anchorTest2/o2sim_geometry.root_geometry-aligned.root does not
 exist
[29013:qc-task-TOF-TaskFT0TOF]: [15:23:59][FATAL] Failed to open file o2sim_geometry.root_geometry-aligned.root
[29013:qc-task-TOF-TaskFT0TOF]: For later analysis we write a core dump to core_dump_29013
[29
```

This needs fixing before we can integrate into alibi.

chiarazampolli:
This means that the file is not there.

PR #1234, opened 2022-06-03
TPCQC: Adds update in PID task

sbhawani:


Modules/TPC/src/PID.cxx
```diff
@@ -43,18 +43,13 @@ void PID::initialize(o2::framework::InitContext& /*ctx*/)
...
-    getObjectsManager()->startPublishing(&hist);
-    getObjectsManager()->addMetadata(hist.GetName(), "custom", "43");
-  }
+  //pass map of vectors of histograms to be beutified!
+  o2::tpc::qc::helpers::setStyleHistogramsInMap(mQCPID.getMapOfHisto());
+  for (auto const& [key, vecOfHist] : mQCPID.getMapOfHisto()) {
+    for (auto& hist : vecOfHist) {
+      getObjectsManager()->startPublishing(hist.get());
+    }
+ }
```
wiechula, line 52: Avoid warnig of unused value:
```suggestion
  for (auto const& pair : mQCPID.getMapOfHisto()) {
    for (auto& hist : pair.second) {
      getObjectsManager()->startPublishing(hist.get());
    }
 }
```

wiechula (changes requested):

Modules/TPC/src/PID.cxx
sbhawani: I have implemented the changes

wiechula (approved):

PR #1244, opened 2022-06-06
CPV: make sure configure() was called in PedestalCheck

sevdokim:
Hello! I noticed that configure() was not called in pedestal run due to some reason. So decided to make it sure that it was called at least once before check() anything.

### DIFF
--- a/Modules/CPV/include/CPV/PedestalCheck.h
+++ b/Modules/CPV/include/CPV/PedestalCheck.h
@@ -42,16 +42,17 @@ class PedestalCheck : public o2::quality_control::checker::CheckInterface
  private:
   int getRunNumberFromMO(std::shared_ptr<MonitorObject> mo);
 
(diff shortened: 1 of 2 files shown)


knopers8:
@Barthelemy this does not seem correct that we do not execute `CheckInterface::configure`

sevdokim:
hi @knopers8 
it seems to be strange for me. Locally everything is executed normally but in the production any output regarding configure() is missing at least for cpv pedestal check.

Barthelemy:
Could you please create a ticket in JIRA explaining the issue ? we will have a look.

sevdokim:
hi @Barthelemy 
https://alice.its.cern.ch/jira/projects/QC/issues/QC-804

Barthelemy (approved):
I approve because you need it but I will also investigate what is happening.

PR #1246, opened 2022-06-07
Update offline track task

KotliarovAr:
Update according to PR#1215


iravasen:
@KotliarovAr thanks a lot! All looks good to me.

iravasen (approved):

PR #1257, opened 2022-06-09
Add protection for oob access

ktf:


njacazio:
@ktf indeed you spotted a bug! Thanks a lot! but the oob issue is because we are incrementing the wrong index (bound to 91) instead of the good one (bound to 72)

ktf:
@njacazio happy to help. Maybe we should get a mac for everyone since linux seems to be a tad too forgiving for OOB accesses.

njacazio:
Gladly ;)

njacazio (approved):

Barthelemy (approved):

PR #1258, opened 2022-06-09
Use utility to correctly get the MID digits

dstocco:
The decoding sends three subspecs:
- data
- calibration events
- fet events

The QC should match all of them via concrete matcher (without specifying the subspecs) and then get the desired digits internally. This ensures that:
- the workflow matches all outputs
- the workflow can be built even in the case where some of the subspecs are missing (e.g. in simulation we do not have calibration or fet events).

This is a common issue with all workflows reading MID digits. To avoid code duplication, a dedicated class is used to handle MID digits.
This PR uses such a class in the MID QC.

### DIFF
--- a/Modules/MUON/MID/CMakeLists.txt
+++ b/Modules/MUON/MID/CMakeLists.txt
@@ -2,17 +2,24 @@
 
 add_library(O2QcMID)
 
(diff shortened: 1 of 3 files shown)


dstocco:
@ValerieRamillien : citing you so that you're informed of the status of this PR

Barthelemy (approved):

PR #1263, opened 2022-06-09
Provisionary fix for configKeyValues option

shahor02:
@knopers8 @Barthelemy as a follow-up of my email: given that the QC initializations are quite convoluted and we need to fix the ConfigurableParam functionality urgently in QC, here is the least invasive patch to pass the globally defined option to devices init. I would consider this as a temporary fix until the global/local options conflict will be fixed.


Barthelemy:
Thank you for this.
Indeed, it is a quick and dirty trick. 
I'll let Piotr comment as he was the one who made most of the work on the config.

knopers8 (approved):
Thank you and apologies that I could not reply to your email yesterday. This should be indeed fine for now.

PR #1269, opened 2022-06-11
Fix in MID digits ROFs binding

shahor02:
@dstocco The fullCI is failing with https://ali-ci.cern.ch/alice-build-logs/AliceO2Group/AliceO2/9036/40689a74e6fe534f9a51a27002226eaa90dfc6c0/build_O2_fullCI/pretty.html, I assume this is a reason.

### DIFF
--- a/Modules/MUON/MID/src/DigitsQcTask.cxx
+++ b/Modules/MUON/MID/src/DigitsQcTask.cxx
@@ -273,7 +273,7 @@ static std::pair<uint32_t, uint32_t> getROFSize(const o2::mid::ROFRecord& rof, g
 void DigitsQcTask::monitorData(o2::framework::ProcessingContext& ctx)
 {
   auto digits = o2::mid::specs::getData(ctx, "digits", o2::mid::EventType::Standard);
-  auto rofs = o2::mid::specs::getRofs(ctx, "digits", o2::mid::EventType::Standard);
+  auto rofs = o2::mid::specs::getRofs(ctx, "digitrofs", o2::mid::EventType::Standard);
 
   int multHitMT11B = 0;
   int multHitMT12B = 0;


dstocco:
Hi @shahor02 , no this was fine. GetRofs("digits"), will search for "digits_rof". If you change it as you did, it will search for "digitrofs_rof". I will need to check what .json is used and fix the .json instead. Is it qc-async/mid.json?

dstocco:
Ah I see. The problem is actually that the CI is not using the QC master....so it will fail.
You should revert this commit. When the master QC is used in the CI, the problem will be gone.

shahor02:
@dstocco the crash comes from sync phase, I see the following:
```
~/alice/O2DPG/DATA/production$ git grep 'MID/DATAROF'
qc-async/mid.json:      "query": "digits:MID/DATA;digitrofs:MID/DATAROF",
qc-sync/mid-digits.json:      "query": "digits:MID/DATA;digits_rof:MID/DATAROF",
qc-sync/mid.json:      "query": "digits:MID/DATA;digits_rof:MID/DATAROF",
```

If only json needs to be fixed, let me know if this PR needs to be closed

shahor02:
> Ah I see. The problem is actually that the CI is not using the QC master....so it will fail. You should revert this commit. When the master QC is used in the CI, the problem will be gone.

which commit you refer to?

dstocco:
This one: https://github.com/AliceO2Group/QualityControl/commit/f154edfc4ad17919a15473addff8a2eef981c9d5
I modified the query in the `.json` to cope with that commit.
But, if this commit is not yet used by the CI, the modification of the `.json` file will result in the error you observed.

dstocco:
Notice that I waited to modify `qc-async/mid.json` (using `digits_rof` instead of `digits`) as I did for the other in order to avoid breaking the CI. I did not realise that the qc-sync is used.
Once the QC commit I mentioned before will be used in the CI, I plan to modify the `qc-async/mid.json` as well.

shahor02:
OK, I cannot revert anything in the QC, please ask @Barthelemy or @knopers8 (or, perhaps they can bump the QC).

dstocco:
It would be better too bump the QC, actually...

dstocco:
I can make a quick fix on O2DPG in the meanwhile...

dstocco:
See https://github.com/AliceO2Group/O2DPG/pull/461/commits/fd28532c3d520bc04dc19c13036dbc5a3a56d77e

dstocco:
I see that the fix is in O2DPG is approved. Thanks. So you can close this PR.

PR #1272, opened 2022-06-13
Always use FLP/DISTSUBTIMEFRAME/0 with '/0'

davidrohr:
Whenever one subscribes to FLP/DISTSUBTIMEFRAME, one needs to request subspec 0, otherwise there is a problem with the matching.

### DIFF
--- a/Modules/ITS/itsFLP.json
+++ b/Modules/ITS/itsFLP.json
@@ -96,7 +96,7 @@
             "id": "RAWDATA",
             "active": "true",
             "machines": [],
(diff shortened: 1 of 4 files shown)


vascobarroso:
@knopers8 @Barthelemy

davidrohr:
Note, this is not urgent, since these JSONs are not used. I just fixed it also here for consistency.

iravasen:
thanks @davidrohr, clear to me. Approving. But we need @Barthelemy or @knopers8 to force merge.

PR #1290, opened 2022-06-21
TPC: added QC task for junk detection

tklemenz:


Modules/TPC/src/JunkDetection.cxx
```diff
@@ -0,0 +1,132 @@
...
+  junkDetectionMsg->AddText("Removal Strategy A");
+  junkDetectionMsg->AddText(fmt::format("Number of Clusters before Removal: {}", statsA[2]).data());
+  junkDetectionMsg->AddText(fmt::format("Removed Fraction: {:.2f}%", statsA[4]).data());
+  junkDetectionMsg->AddLine(.0, .5, 1., .5);
+  junkDetectionMsg->AddText("Removal Strategy B");
+  junkDetectionMsg->AddText(fmt::format("Number of Clusters before Removal: {}", statsB[2]).data());
+  junkDetectionMsg->AddText(fmt::format("Removed Fraction: {:.2f}%", statsB[4]).data());
+
+  c->cd();
+  junkDetectionMsg->Draw();
```
wiechula, line 127:
```suggestion
  junkDetectionMsg->SetBit(TObject::kCanDelete);
  junkDetectionMsg->Draw();
```

Modules/TPC/include/TPC/JunkDetection.h
```diff
@@ -0,0 +1,61 @@
...
+
+/// \brief TPC JunkDetection QC Task
+/// \author Thomas Klemenz
+class JunkDetection final : public TaskInterface
+{
+ public:
+  /// \brief Constructor
+  JunkDetection() = default;
+  /// Destructor
+  ~JunkDetection() = default;
```
wiechula, line 41:
```suggestion
  ~JunkDetection()
  {
    for (auto hist : mJDHistos) {
    delete hist;
    }
  }
```

wiechula (changes requested):

tklemenz:
I updated the PR with the suggested changes and added a few more.

- Added a switch in the json file to enable/disable the canvas output in case we want to run locally (TCanvas not mergeable).


Unrelated to junkDetection:

- Removed unnecessary log output from `Common/Utils.h` and added a useful line
- Removed unnecessary log output from `GLO/DataCompressionQcTask.cxx`
- Removed `using o2::tpc::qc::Clusters` from `TPC/ClustersData.h` and specified the namespace explicitly in each instance of a `Clusters` object

The last point was needed to make `ClustersData` usable in macros. Otherwise, there was a complaint about a forward declaration of `o2::quality_control_modules::tpc::Clusters`.
```
error: field has incomplete type 'o2::quality_control_modules::tpc::Clusters'
  Clusters mClusters;
```
Apparently the same object name in different namespaces (o2::tpc::qc::Clusters vs o2::quality_control_modules::tpc::Clusters) lead to confusion during compilation with `using o2::tpc::qc::Clusters`.

Let me know if you think this doesn't make sense.

wiechula (approved):

Barthelemy (approved):

PR #1294, opened 2022-06-22
TRD added tracklets per timeframe, remove ccdb server setting from unused code

bazinski:
- add tracklet per time frame spectrum.
- overlays for digits is unmanagable so reverted to what is in master. Will have to find a quicker way to handle the canvases, because what I have is unworkable.
- remove alice-ccdb from pulseheight task which is no longer used.

### DIFF
--- a/Modules/TRD/TRDQC.json
+++ b/Modules/TRD/TRDQC.json
@@ -112,32 +112,6 @@
 	"saveObjectsToFile":"qcpulseheight.root"
       }
     },
-    "checks": {
-      "QcCheck": {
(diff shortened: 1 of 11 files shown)


bazinski:
@Barthelemy @knopers8 ok testing concluded, not sure what was going wrong but the spectra now getting updated as expected. I think qcg-test is just taking a serious amount strain, might also explain why my overlaid spectra were so slow. Will fix the overlay spectra seperately.

Barthelemy (approved):

PR #1296, opened 2022-06-22
GLO: Added task for data compression monitoring for all detectors

tklemenz:
This PR adds a qc task for monitoring the entropy compression and compression factors of the detectors. It can be used for any number of detectors at a time.

In addition, I added Utils to Common with a function for getting parameters from taskParameters in the config easily. The function was recently introduced in the TPC code but I think it should rather be in a common place. I left it also in TPC/Utility for now for compatibility reasons. Later this will be removed and the corresponding tasks using the function will be adjusted.


Modules/GLO/src/DataCompressionQcTask.cxx
```diff
@@ -0,0 +1,142 @@
...
+#include "Framework/ProcessingContext.h"
+#include <Framework/InputRecord.h>
+
+// QC includes
+#include "QualityControl/QcInfoLogger.h"
+#include "GLO/DataCompressionQcTask.h"
+#include "Common/Utils.h"
+
+// root includes
+#include "TH1.h"
```
wiechula, line 27: Doesn't `unique_ptr` need the header? So perhaps move this to the header.

Modules/GLO/src/DataCompressionQcTask.cxx
```diff
@@ -0,0 +1,142 @@
...
+  TH1::AddDirectory(false);
+
+  const auto nBins = o2::quality_control_modules::common::getFromConfig<int>(mCustomParameters, "nBins");
+  const auto xMin = o2::quality_control_modules::common::getFromConfig<int>(mCustomParameters, "xMin");
+  const auto xMax = o2::quality_control_modules::common::getFromConfig<int>(mCustomParameters, "xMax");
+
+  // initialize histograms for active detectors
+  bool useAll = o2::quality_control_modules::common::getFromConfig<bool>(mCustomParameters, "useAll");
+  for (const auto& det : detList) {
+    bool useDet = o2::quality_control_modules::common::getFromConfig<bool>(mCustomParameters, det);
```
wiechula, line 48:
```suggestion
for (int iDet = 0; iDet < o2::detectors::DetID::nDetectors; ++iDet) {
    const bool useDet = o2::quality_control_modules::common::getFromConfig<bool>(mCustomParameters, o2::detectors::DetID::getName(iDet));
```

Modules/GLO/src/DataCompressionQcTask.cxx
```diff
@@ -0,0 +1,142 @@
...
+    for (const auto& hist : pair.second) {
+      hist->Reset();
+    }
+  }
+}
+
+void DataCompressionQcTask::processMessage(const o2::ctf::CTFIOSize& ctfEncRep, const std::string detector)
+{
+  auto entropyCompression = (ctfEncRep.ctfIn - ctfEncRep.ctfOut) / float(ctfEncRep.ctfIn);
+  auto compression = (ctfEncRep.rawIn - ctfEncRep.ctfOut) / float(ctfEncRep.rawIn);
```
wiechula, line 136:
```suggestion
  const auto entropyCompression = float(ctfEncRep.ctfIn - ctfEncRep.ctfOut) / float(ctfEncRep.ctfIn);
  const auto compression = float(ctfEncRep.rawIn - ctfEncRep.ctfOut) / float(ctfEncRep.rawIn);
```

wiechula (changes requested):

Modules/GLO/src/DataCompressionQcTask.cxx
tklemenz: It actually worked with the forward declaration in the case of `TH1`. For `TCanvas` the include was needed in the header. I moved the TH1 include to the header now anyway.

tklemenz:
@knopers8 @wiechula @Barthelemy The error looks unrelated to me.

PR #1302, opened 2022-06-24
TPC Generic Histogram Check

HorstMa:
This Check compares the X and Y mean of any 1D and 2D Histogram against an expected value. This can be done with a range (1 Range = medium, 2 Range = Bad) or with the standard deviation of the mean (3 StdDev = Medium, 6 StdDev = Bad).


wiechula (changes requested):

Modules/TPC/src/GenericHistogramCheck.cxx
```diff
@@ -0,0 +1,354 @@
...
+          } else if (fabs(mMeanY - mExpectedValueY) > 6 * mStdevY) {
+            resultStdDevY = Quality::Bad;
+          } else {
+            resultStdDevY = Quality::Medium;
+          }
+        }
+      }
+    }
+    // find the worst quality that is NOT 10 ("Null")
+    std::vector<Quality> Quals = { resultRangeX, resultRangeY, resultStdDevX, resultStdDevY };
```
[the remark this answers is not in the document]
HorstMa: I can't make this constant, otherwise, the sort does not work.

Modules/TPC/src/GenericHistogramCheck.cxx
```diff
@@ -0,0 +1,353 @@
...
+  Quality result = Quality::Null;
+  for (auto const& moObj : *moMap) {
+    auto mo = moObj.second;
+    if (!mo) {
+      continue;
+      ILOG(Error, Support) << "No MO found" << ENDM;
+    }
+    const auto moName = mo->getName();
+
+    THn* hN = THn::CreateHn(moName.c_str(), moName.c_str(), (TH2F*)mo->getObject());
```
wiechula, line 114: Why is `hN` needed? It is not used except for the line below. I guess the dimension of a histogram can also be obtained directly from the histogram.
Also, the object does not seem to be deleted, so it's a memory leak...

wiechula (changes requested):
One more thing

Modules/TPC/src/GenericHistogramCheck.cxx
HorstMa: There is no way to even initiallize the object since we don't know if its an TH1 or TH2. The (TH2F*) also works for TH1s. when I try to initallize with auto it initiallizes it as a TObject not as a TH1/TH2.

Modules/TPC/src/GenericHistogramCheck.cxx
```diff
@@ -0,0 +1,347 @@
...
+    mHistDimension = hN->GetNdimensions();
+    delete hN;
+    hN = nullptr;
```
wiechula, line 116:
```suggestion
    mHistDimension = ((TH1*)mo->getObject())->GetDimension();
```

wiechula (changes requested):

Modules/TPC/src/GenericHistogramCheck.cxx
```diff
@@ -0,0 +1,343 @@
...
+std::string GenericHistogramCheck::getAcceptedType() { return "TCanvas"; }
+
+void GenericHistogramCheck::beautify(std::shared_ptr<MonitorObject> mo, Quality checkResult)
+{
+  auto moName = mo->getName();
+  TPaveText* msg = new TPaveText(0.11, 0.85, 0.9, 0.95, "NDC");
+  msg->SetBorderSize(1);
+  TText* txt = new TText(0, 0, "Quality::Null");
+  TText* txt2 = new TText(0, 0, "Quality::Null");
+  double xText = 0;
```
wiechula, line 220: auto h = (TH1*)mo->getObject();

wiechula (changes requested):

wiechula (approved):

PR #1305, opened 2022-06-25
FT0: optimization for DigitQcTask, and new features for PostProc

afurs:


Modules/FT0/src/DigitQcTask.cxx
```diff
@@ -310,78 +305,57 @@ void DigitQcTask::startOfCycle()
...
-
-    if (curTfTimeMin < 0)
-      curTfTimeMin = mTimeCurNS;
-    if (mTimeCurNS < curTfTimeMin)
-      curTfTimeMin = mTimeCurNS;
-    if (mTimeCurNS > curTfTimeMax)
-      curTfTimeMax = mTimeCurNS;
-
-    if (digit.mTriggers.amplA == -5000 && digit.mTriggers.amplC == -5000 && digit.mTriggers.timeA == -5000 && digit.mTriggers.timeC == -5000)
+    if (digit.mTriggers.amplA == -5000 /*&& digit.mTriggers.amplC == -5000 && digit.mTriggers.timeA == -5000 && digit.mTriggers.timeC == -5000*/) {
```
mslupeck, line 324: Hi Artur,
Do you know where the value of -5000 was taken from?
The default amplitude (according to https://github.com/AliceO2Group/AliceO2/blob/dev/DataFormats/Detectors/FIT/common/include/DataFormatsFIT/Triggers.h#L45 ) is 0, not -5000. It used to be -5000 though, so it may be a leftover. Have you tested that this condition works as expected?

afurs: Hi @mslupeck , ah, yes, you are right, I took condition from previous QC version. So in previous version we had the same situation. Ok, lets keep this in current PR(due to CICD is already passed). I'll prepare hotfix.

afurs:
Hi @Barthelemy , could you please merge this PR? I'll prepare some trivial hotfix after, as we discussed with @mslupeck .

PR #1310, opened 2022-06-27
FT0: hotfix for DigitQCtask

afurs:
Trivial hotfix, TCM presence condition is fixed and reduced. No need in waiting CICD to be finished.

### DIFF
--- a/Modules/FT0/src/DigitQcTask.cxx
+++ b/Modules/FT0/src/DigitQcTask.cxx
@@ -321,7 +321,7 @@ void DigitQcTask::monitorData(o2::framework::ProcessingContext& ctx)
   for (auto& digit : digits) {
     const auto& vecChData = digit.getBunchChannelData(channels);
     bool isTCM = true;
-    if (digit.mTriggers.amplA == -5000 /*&& digit.mTriggers.amplC == -5000 && digit.mTriggers.timeA == -5000 && digit.mTriggers.timeC == -5000*/) {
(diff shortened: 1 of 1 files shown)


Modules/FT0/src/DigitQcTask.cxx
```diff
@@ -321,7 +321,7 @@ void DigitQcTask::monitorData(o2::framework::ProcessingContext& ctx)
   for (auto& digit : digits) {
     const auto& vecChData = digit.getBunchChannelData(channels);
     bool isTCM = true;
-    if (digit.mTriggers.amplA == -5000 /*&& digit.mTriggers.amplC == -5000 && digit.mTriggers.timeA == -5000 && digit.mTriggers.timeC == -5000*/) {
+    if (digit.mTriggers.timeA == -5000 && digit.mTriggers.timeC == -5000) {
```
mslupeck, line 324: Is there any particular reason why you used hardcoded value `-5000` instead of the `o2::fit::Triggers::DEFAULT_TIME`

afurs: > Is there any particular reason why you used hardcoded value `-5000` instead of the `o2::fit::Triggers::DEFAULT_TIME`

Hi @maciej, yes , hotfix w/o compiling :)

afurs:
Hello @Barthelemy , there are some unrelated errors during CICD. This is trivial hotfix, I think there are no need in waiting CICD to be finished.

Barthelemy:
merging then

PR #1311, opened 2022-06-27
[EMCAL-524] Low,High Gain Error per SM added

cterrevo:
histogram to monitor Low, High Gain Errors per SM added


mfasDa (approved):
Looks good

PR #1313, opened 2022-06-28
Remove the old retrieveCondition and use retrieveConditionAny everywhere

Barthelemy:
Note that retrieveConditionAny is not const.


Barthelemy:
@aphecetche

knopers8 (approved):

PR #1315, opened 2022-06-28
ITS: normalized cos(lambda) plot by n_clusters 

iravasen:
- Normalization to n_clusters of cos(lambda) plots
- removed the same version of the plot per layer (kept the one for the full detector only)


iravasen:
@IsakovAD this is completed. I summarize here the changes:
- removal of cluster size * cos(lambda) plot layer by layer keeping only the one for the full layer.
- Rebin of x axis so to have 1 chip granularity for IB and 1 HIC granularity for OB. This improves visibility and reduces by a factor 10 the size of the plot. 
- Rebin of y axis for better visibility
- Normalizion to number of clusters. 

Let me know in case of any comments. In the meantime I remove WIP so that the tests can start.

Barthelemy (approved):

PR #1327, opened 2022-07-06
[MRRTF-186] QC for mch mid matched tracks

aphecetche:
Introduce a new task that can be used to monitor MCH standalone tracks and/or MCH-MID matched tracks.

In order to ensure the current QC (MCH only) tracks won't break at Pt2, this is a new task and not a replacement for the current one. Only once validated will this new one replace the current one.

This PR has also a couple of fix for the current tracks task, please do not squash.

### DIFF
--- a/Modules/MUON/Common/etc/qc-tracks.json
+++ b/Modules/MUON/Common/etc/qc-tracks.json
@@ -1,0 +1,78 @@
+{
+  "qc": {
+    "config": {
+      "database": {
+        "implementation": "CCDB",
+        "host": "localhost:6464",
+        "username": "not_applicable",
+        "password": "not_applicable",
+        "name": "not_applicable"
+      },
+      "conditionDB": {
+        "url": "http://alice-ccdb.cern.ch"
+      },
+      "Activity": {
+        "number": "42",
+        "type": "2",
+        "start":"1655682579514"
+      }
+    },
+    "tasks": {
+      "FwdTracks": {
+        "active": "true",
+        "className": "o2::quality_control_modules::muon::TracksTask",
+        "moduleName": "QcMuonCommon",
+        "detectorName": "GLO",
+        "cycleDurationSeconds": "120",
+        "maxNumberCycles": "-1",
+        "dataSource": {
+          "type": "direct",
+          "query": "mchtracks:MCH/TRACKS;mchtrackrofs:MCH/TRACKROFS;mchtrackclusters:MCH/TRACKCLUSTERS;mchtrackdigits:MCH/CLUSTERDIGITS;muontracks:GLO/MTC_MCHMID"
+        },
+        "taskParameters": {
+          "maxTracksPerTF": "100",
+          "standaloneMCHTracks": true,
+          "matchedMCHMIDTracks": true
+        },
+        "location": "remote"
+      },
+      "MCHStdTracks": {
+        "active": "true",
+        "className": "o2::quality_control_modules::muon::TracksTask",
+        "moduleName": "QcMuonCommon",
+        "detectorName": "GLO",
+        "cycleDurationSeconds": "120",
+        "maxNumberCycles": "-1",
+        "dataSource": {
+          "type": "direct",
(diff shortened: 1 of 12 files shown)


aphecetche:
@aferrero2707 if you want to have a look

Modules/MUON/Common/src/TrackPlotter.cxx
```diff
@@ -0,0 +1,339 @@
...
+{
+  TAxis* axis = h->GetXaxis();
+  for (int i = 1; i <= 10; i++) {
+    auto label = fmt::format("CH{}", i);
+    axis->SetBinLabel(i, label.c_str());
+  }
+}
+
+uint16_t computeDsBinX(int feeId, int linkId, int elinkId)
+{
```
aferrero2707, line 49: It might be worth moving the function in the helpers, since we use the same kind of binning in other tasks.

Modules/MUON/Common/src/TrackPlotter.cxx
```diff
@@ -0,0 +1,339 @@
...
+  if (tracks.size() > 1) {
+    for (auto i = 0; i < tracks.size(); i++) {
+      auto ti = getMomentum4D(tracks[i]);
+      for (auto j = i + 1; j < tracks.size(); j++) {
+        if (tracks[i].getSign() == tracks[j].getSign()) {
+          continue;
+        }
+        auto tj = getMomentum4D(tracks[j]);
+        auto p = ti + tj;
```
aferrero2707, line 235: Would it make sense to add a cut on the track time difference in order to reduce the combinatorial background?

Barthelemy:
we enforce squashing but I can make an exception :)

Modules/MUON/Common/include/MUONCommon/Helpers.h
```diff
@@ -0,0 +1,32 @@
...
+
+#include <gsl/span>
+
+class TH1;
+class TLine;
+
+namespace o2::quality_control_modules::muon
+{
+TLine* addHorizontalLine(TH1& histo, double y);
+TLine* addVerticalLine(TH1& histo, double x, int lineColor = 1, int lineStyle = 10,
```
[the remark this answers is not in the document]
aphecetche: done

Modules/MUON/Common/include/MUONCommon/Helpers.h
```diff
@@ -0,0 +1,32 @@
...
+                       int lineWidth = 1);
+void cleanup(TH1& histo, const char* classname);
+void markBunchCrossing(TH1& histo,
+                       const char* regexp,
+                       gsl::span<int> bunchCrossings);
+
```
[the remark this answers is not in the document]
aphecetche: indeed. removed.

Modules/MUON/Common/src/TrackPlotter.cxx
aphecetche: the issue with the combinatorial background was elsewhere : I was making track pairs across the full TF, not within each ROF. Now fixed.

aphecetche:
> we enforce squashing but I can make an exception :)

 Ah, did not know that. It's not super highly critical to keep the commits separated, so I can also squash them on my end (as I anyway need to fix the compilation issue on Linux...)

Modules/MUON/Common/src/TrackPlotter.cxx
aphecetche: yes, but for another PR though...

Modules/MUON/Common/etc/qc-tracks.json
```diff
@@ -0,0 +1,78 @@
...
+          "maxTracksPerTF": "100",
+          "standaloneMCHTracks": true,
+          "matchedMCHMIDTracks": true
+        },
+        "location": "remote"
+      },
+      "MCHStdTracks": {
+        "active": "true",
+        "className": "o2::quality_control_modules::muon::TracksTask",
+        "moduleName": "QcMuonCommon",
```
aferrero2707, line 42: Should be `"moduleName": "QcMUONCommon"`

Barthelemy:
as most tests passed, I will merge

Barthelemy (approved):

PR #1333, opened 2022-07-09
ITS fix cluster size dist and average cluster size for empty bins

iravasen:


Barthelemy:
@TimoWilken We seem to have an issue with build/QualityControl/O2fst/o2. Could you have a look ? thank you.

merging, error unrelated.

Barthelemy (approved):

PR #1334, opened 2022-07-09
TRD fix errorneous massive 2d histogram causing root errors, and other sundry fixes

bazinski:
- fix an errorneous large spectrum with 25000 x 25000 spectra, finger trouble in the zeros producing very odd root error messages.
- clean up digits per pad per layer and tracklets per mcm per layer
- fix tracklets per timeframe.


Barthelemy:
Error unrelated, merging.

PR #1346, opened 2022-07-14
Do not allow to publish in two PP callbacks in one iteration

knopers8:
When objects were published during update() and finalize() in one Data Processor's run(),
we would create two sets of MonitorObjects. CheckRunner cannot cope with this, it may accept
just one set at a time, thus we avoid this from happening.

A proper solution would be to let CheckRunner process more than one MOC for a source at the same time,
but this might require some major changes.


knopers8:
it also includes some clang-tidy advices.

Barthelemy (approved):

PR #1357, opened 2022-07-21
MCH: fix histogram reset at each cycle

aferrero2707:
- added method to properly reset the numerator and denominator in the mergeable histogram ratios
- added option "ICES" to the histogram reset to avoid removing the associated functions
- added reset of orbit counters in digits task

### DIFF
--- a/Modules/MUON/Common/include/MUONCommon/MergeableTH1Ratio.h
+++ b/Modules/MUON/Common/include/MUONCommon/MergeableTH1Ratio.h
@@ -60,6 +60,8 @@ class MergeableTH1Ratio : public TH1F, public o2::mergers::MergeInterface
 
   void update();
 
+  void Reset(Option_t*  option = "");
+
  private:
   TH1D* mHistoNum{ nullptr };
   TH1D* mHistoDen{ nullptr };
--- a/Modules/MUON/Common/include/MUONCommon/MergeableTH1Ratio.h
+++ b/Modules/MUON/Common/include/MUONCommon/MergeableTH1Ratio.h
@@ -60,7 +60,7 @@ class MergeableTH1Ratio : public TH1F, public o2::mergers::MergeInterface
 
   void update();
 
-  void Reset(Option_t* option = "");
+  void Reset(Option_t* option = "") override;
 
  private:
   TH1D* mHistoNum{ nullptr };
@@ -68,7 +68,7 @@ class MergeableTH1Ratio : public TH1F, public o2::mergers::MergeInterface
   std::string mTreatMeAs = "TH1F";
   double mScalingFactor = 1.;
 
-  ClassDefOverride(MergeableTH1Ratio, 1);
+  ClassDefOverride(MergeableTH1Ratio, 2);
 };
 
 } // namespace o2::quality_control_modules::muon
--- a/Modules/MUON/Common/include/MUONCommon/MergeableTH1Ratio.h
+++ b/Modules/MUON/Common/include/MUONCommon/MergeableTH1Ratio.h
@@ -60,7 +60,7 @@ class MergeableTH1Ratio : public TH1F, public o2::mergers::MergeInterface
 
   void update();
 
-  void Reset(Option_t*  option = "");
+  void Reset(Option_t* option = "");
 
  private:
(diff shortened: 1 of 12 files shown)


aferrero2707:
@aphecetche this PR contains fixes to the resetting of histograms in the tasks where the reset was already implemented. I have not touched the tracks tasks.

knopers8:
@aphecetche , @aferrero2707 FYI, we plan to release QC tomorrow morning.

Modules/MUON/Common/include/MUONCommon/MergeableTH1Ratio.h
```diff
@@ -60,13 +60,15 @@ class MergeableTH1Ratio : public TH1F, public o2::mergers::MergeInterface
...
 
+  void Reset(Option_t* option = "") override;
+
  private:
   TH1D* mHistoNum{ nullptr };
   TH1D* mHistoDen{ nullptr };
   std::string mTreatMeAs = "TH1F";
   double mScalingFactor = 1.;
 
-  ClassDefOverride(MergeableTH1Ratio, 1);
```
aphecetche, line 69: thanks but actually in this instance it's not really needed as you're only adding methods, not data members.

Modules/MUON/Common/src/MergeableTH1Ratio.cxx
```diff
@@ -115,4 +115,11 @@ void MergeableTH1Ratio::update()
   }
 }
 
+void MergeableTH1Ratio::Reset(Option_t* option)
+{
+  getNum()->Reset(option);
```
aphecetche, line 120: do we have a guarantee that getNum() won't return a nullptr ?

aferrero2707: The num and den histograms are initialized in the constructors. They could be null if there is not enough memory, but then things would break much earlier, for example during the merge or update, which are also not protected. I can however add protections everywhere just to be on the safe side...

Modules/MUON/Common/include/MUONCommon/MergeableTH1Ratio.h
aferrero2707: So should I revert the version?

aphecetche: Actually, yes, please. That'll save me from changing the ccdb-proxy of mchview to deal with a new version ;-)

Modules/MUON/Common/src/MergeableTH1Ratio.cxx
aphecetche: well, the default constructor does not create histograms, does it ? So basically if you do : 

```
MergeableTH1Ratio h;
h.Reset();
```

you'd get a crash, right ?

aferrero2707: You have a good point here...

Modules/MUON/Common/include/MUONCommon/MergeableTH1Ratio.h
aferrero2707: I should actually remove the scaling factor parameter as I did for the MergeableTH2Ratio, in which case the bump in version would be justified.
However I can also keep the class as-is and revert the version bump if that's easier for you to handle.

On the other hand, is the version bump for  MergeableTH2Ratio ok? It was forgotten in the previous PR...

aphecetche: If no data member are added/removed, I would keep the versions as they are. 
For TH2Ratio I had to "patch" the ccdb-proxy https://gitlab.cern.ch/alice-muon/ccdb-proxy/-/blob/master/ccdb/types.go#L154 and while it seems to work, I would like to avoid more patches like that if possible.

aferrero2707: Ok, so I revert both version bumps, right?

aphecetche: yes, thanks.

PR #1359, opened 2022-07-21
MFT: plots for cluster size in pixels

diana0x0f:
Adding histograms for cluster size in pixels, separately for normal (frequent) cluster shapes and for grouped clusters.

### DIFF
--- a/Modules/MFT/include/MFT/QcMFTClusterTask.h
+++ b/Modules/MFT/include/MFT/QcMFTClusterTask.h
@@ -58,6 +58,7 @@ class QcMFTClusterTask /*final*/ : public TaskInterface // todo add back the "fi
   std::unique_ptr<TH1F> mClusterOccupancy = nullptr;
   std::unique_ptr<TH1F> mClusterPatternIndex = nullptr;
   std::unique_ptr<TH1F> mClusterSizeSummary = nullptr;
+  std::unique_ptr<TH1F> mGroupedClusterSizeSummary = nullptr;
   std::unique_ptr<TH2F> mClusterOccupancySummary = nullptr;
 
   std::unique_ptr<TH2F> mClusterPatternSensorIndices = nullptr;
@@ -87,7 +88,7 @@ class QcMFTClusterTask /*final*/ : public TaskInterface // todo add back the "fi
   // internal functions
   void getChipMapData();
 
-  // cluster size in pixels 
+  // cluster size in pixels
   int mClusterSize = { 0 };
 
   // dictionary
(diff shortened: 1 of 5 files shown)


knopers8:
The failure of FST seems legit:
```
reco_NOGPU.log:[2661:qc-task-MFT-QcMFTClusterTask]: [05:59:26][ERROR] Exception caught: Unknown argument requested patterns - no matching route found for patterns 
reco_NOGPU.log-[2661:qc-task-MFT-QcMFTClusterTask]:     /sw/slc7_x86-64/QualityControl/1359-slc7_x86-64-local2/lib/libO2QcMFT.so: decltype(auto) o2::framework::InputRecord::get<gsl::span<unsigned char, 18446744073709551615ul>, char const*>(char const*, int) const
reco_NOGPU.log-[2661:qc-task-MFT-QcMFTClusterTask]:     /sw/slc7_x86-64/QualityControl/1359-slc7_x86-64-local2/lib/libO2QcMFT.so: o2::quality_control_modules::mft::QcMFTClusterTask::monitorData(o2::framework::ProcessingContext&)
```

diana0x0f:
Hi @knopers8 ,
the query is defined in `qc-mft-cluster.json`, the patterns branch is filled in the `runMFTClustersRootFileReader.cxx` file, it follows the same structure as the rest. It runs with no errors on my machine. What else could be there that prevents it from successfully running? thanks

tomas-herman:
Hi @knopers8, perhaps some more info that could help understand the issue. The use of the patterns that we are implementing in this PR is already in production by ITS without issues, e.g.: https://github.com/AliceO2Group/QualityControl/blob/41f4b2bc1776a2bca6346547719f95826575f589/Modules/ITS/src/ITSClusterTask.cxx#L150
The pattern object is defined here: https://github.com/AliceO2Group/AliceO2/blob/dev/DataFormats/Detectors/ITSMFT/common/include/DataFormatsITSMFT/ClusterTopology.h respectively here: https://github.com/AliceO2Group/AliceO2/blob/dev/DataFormats/Detectors/ITSMFT/common/include/DataFormatsITSMFT/ClusterPattern.h

knopers8:
hm, perhaps the config file for the FST does not request "patterns"? I am not 100% sure, but I think this is the file it is using:
https://github.com/AliceO2Group/O2DPG/blob/master/DATA/production/qc-sync/mft.json

tomas-herman:
Hi @knopers8, this json file was not updated for quite some time so if it is being used for the check I understand why it fails. Is this a new check? I am surprised we did not have this issue sooner.

I have created a PR with the latest update: https://github.com/AliceO2Group/O2DPG/pull/551

Once it is merged the test should go ok?

knopers8:
If this is indeed the json file which is used there, it should stop failing, but as I said, I am not sure myself if this is the one, as I don't take care of those full system tests.

tomas-herman (approved):

tomas-herman:
With this fix it should finally work: https://github.com/AliceO2Group/O2DPG/pull/555

PR #1362, opened 2022-07-22
[EMCAL-792] Switch for warnings of the checker in the IL as checkerParam

mfasDa:


Modules/EMCAL/src/RawErrorCheck.cxx
```diff
@@ -41,6 +37,16 @@ namespace o2::quality_control_modules::emcal
 void RawErrorCheck::configure()
 {
   mGeometry = o2::emcal::Geometry::GetInstanceFromRunNumber(300000);
+
+  // switch on/off messages on the infoLogger
```
knopers8, line 41: Were you aware that you can use custom severity values and select the severity cut off threshold in the QC config file?

knopers8: I am fine with it as it is though.

mfasDa: No, I was not, because that affects also another checker which we want to mute on the infoLogger while we still tune the checkers. But for the moment can we go ahead with this and I might change it in a second iteration?

knopers8: sure

PR #1363, opened 2022-07-24
FIT: all calibration tasks are turned off for global refactoring in FITCalibration libs

afurs:


afurs:
Unrelated issue

knopers8 (approved):

afurs:
@knopers8 @Barthelemy shall I restart this PR by adding new commit?

knopers8:
o2-dataflow is spurrious and FST is already being redone, so no need to push new commits. It would be good to check if FST survives if you are disabling a few tasks. That is unless you are sure FST does not run them, then I can force-merge.

afurs:
> o2-dataflow is spurrious and FST is already being redone, so no need to push new commits. It would be good to check if FST survives if you are disabling a few tasks. That is unless you are sure FST does not run them, then I can force-merge.

Yes, calibration tasks which are excluded for a while in this PR, are not included int FST( ft0.json config)

knopers8:
the errors are unrelated, force-merging

PR #1364, opened 2022-07-25
TOF: Add protection agains 0 division

njacazio:
With this I add a protection against any division by 0 in the trending task.

### DIFF
--- a/Modules/TOF/src/TrendingRate.cxx
+++ b/Modules/TOF/src/TrendingRate.cxx
@@ -97,7 +97,7 @@ void TrendingRate::computeTOFRates(TH2F* h, std::vector<int>& bcInt, std::vector
     return;
   }
 
-  if (mActiveChannels > 0.f) {
+  if (mActiveChannels > 0) {
     mNoiseRatePerChannel = (hback->GetMean() - 0.5) / orbit_lenght * h->GetNbinsX() / mActiveChannels;
   }
 
--- a/Modules/TOF/src/TrendingRate.cxx
+++ b/Modules/TOF/src/TrendingRate.cxx
@@ -98,12 +98,7 @@ void TrendingRate::computeTOFRates(TH2F* h, std::vector<int>& bcInt, std::vector
   }
 
   if (mActiveChannels > 0.f) {
-    if (orbit_lenght > 0.f) {
-      mNoiseRatePerChannel = (hback->GetMean() - 0.5) / orbit_lenght * h->GetNbinsX() / mActiveChannels;
-    } else {
-      ILOG(Warning, Support) << "Unphysical orbit_lenght " << orbit_lenght << ", setting noise rate per channel to 0" << ENDM;
-      mNoiseRatePerChannel = 0.f;
-    }
+    mNoiseRatePerChannel = (hback->GetMean() - 0.5) / orbit_lenght * h->GetNbinsX() / mActiveChannels;
   }
 
   std::vector<int> signals;
@@ -149,10 +144,6 @@ void TrendingRate::computeTOFRates(TH2F* h, std::vector<int>& bcInt, std::vector
         continue;
       }
       const float mu = TMath::Log(1.f / (1.f - prob));
-      if (orbit_lenght <= 0.f) {
-        ILOG(Warning, Support) << "Unphysical orbit_lenght, cannot compute rate" << orbit_lenght << ENDM;
-        continue;
-      }
       const float rate = mu / orbit_lenght;
       bcInt.push_back(ibc);
       bcRate.push_back(rate);
--- a/Modules/TOF/src/TrendingRate.cxx
+++ b/Modules/TOF/src/TrendingRate.cxx
@@ -97,8 +97,13 @@ void TrendingRate::computeTOFRates(TH2F* h, std::vector<int>& bcInt, std::vector
     return;
   }
 
-  if (mActiveChannels) {
-    mNoiseRatePerChannel = (hback->GetMean() - 0.5) / orbit_lenght * h->GetNbinsX() / mActiveChannels;
+  if (mActiveChannels > 0.f) {
+    if (orbit_lenght > 0.f) {
+      mNoiseRatePerChannel = (hback->GetMean() - 0.5) / orbit_lenght * h->GetNbinsX() / mActiveChannels;
+    } else {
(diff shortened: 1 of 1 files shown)


njacazio:
@ercolessi please have a look as well

ercolessi:
> @ercolessi please have a look as well
This looks good thanks a lot Nicolò!! It will hopefully solve the issue.

Modules/TOF/src/TrendingRate.cxx
```diff
@@ -97,8 +97,13 @@ void TrendingRate::computeTOFRates(TH2F* h, std::vector<int>& bcInt, std::vector
...
 
-  if (mActiveChannels) {
-    mNoiseRatePerChannel = (hback->GetMean() - 0.5) / orbit_lenght * h->GetNbinsX() / mActiveChannels;
+  if (mActiveChannels > 0.f) {
+    if (orbit_lenght > 0.f) {
+      mNoiseRatePerChannel = (hback->GetMean() - 0.5) / orbit_lenght * h->GetNbinsX() / mActiveChannels;
+    } else {
+      ILOG(Warning, Support) << "Unphysical orbit_lenght " << orbit_lenght << ", setting noise rate per channel to 0" << ENDM;
+      mNoiseRatePerChannel = 0.f;
+    }
```
noferini, line 106: Line 100-106 not needed.
Check on mActiveChannels was already done in previous version but as integer (it is an integer)
Check on orbit_length not needed since it is a constant
I would remove all this part from the commit

Modules/TOF/src/TrendingRate.cxx
```diff
@@ -123,29 +128,43 @@ void TrendingRate::computeTOFRates(TH2F* h, std::vector<int>& bcInt, std::vector
       const int bcmax = ibc * 18;
       TH1D* hs = hDiffGlobal.ProjectionY(Form("sign_%d_%d", bcmin, bcmax), ibc, ibc);
       hs->SetTitle(Form("%d < BC < %d", bcmin, bcmax));
+      if (hb->GetBinContent(1) <= 0.) {
+        ILOG(Warning, Support) << "Background bin content is zero, cannot normalize histogram" << ENDM;
+        continue;
+      }
```
noferini, line 134: It is better to move this condition to hback and external to the for loop.
There is no need to check every time inside the loop (note that if you continue inside the loop you have to destroy hs and hb before to continue, outside the loop you don't need to care about that)

Modules/TOF/src/TrendingRate.cxx
```diff
@@ -97,8 +97,13 @@ void TrendingRate::computeTOFRates(TH2F* h, std::vector<int>& bcInt, std::vector
     return;
   }
```
noferini, line 98: This is the point where the current version crashes.
If nb==0 -> hback is not created and then you cannot ask for integral (crash)

Please add this protection here
  if (nb == 0) { // threshold too high? return since hback was not created
    delete hpdiff;
    return;
  }

noferini:
One minor thing
I would also replace
  mTrend->Fill();
with
  if(mNoiseRatePerChannel >= 0) mTrend->Fill();
to remove empty points from the trending plot (up to you)

njacazio:
> if(mNoiseRatePerChannel >= 0) mTrend->Fill();

I see but then we risk not realizing that something is off, no?

noferini:
> > if(mNoiseRatePerChannel >= 0) mTrend->Fill();
> 
> I see but then we risk not realizing that something is off, no?

I don't have a strong opinion on this.Ok if you prefer to keep also this cases

Modules/TOF/src/TrendingRate.cxx
```diff
@@ -93,11 +93,16 @@ void TrendingRate::computeTOFRates(TH2F* h, std::vector<int>& bcInt, std::vector
   mPreviousPlot->Reset();
   mPreviousPlot->Add(h);
 
+  if (nb == 0) { // threshold too high? return since hback was not created
```
njacazio, line 96: Ciao @noferini indeed, now I get the case. This is how you suggested

njacazio:
Thanks @noferini, indeed I was a bit quick with my changes, now this should be mergiable

Modules/TOF/src/TrendingRate.cxx
noferini: fine

noferini (approved):

knopers8:
I saw this in the FST log:
```
./itstpcMatch.logTASK-EXIT-CODE: 0
./tofMatchTracks.logRunning: TIME="#walltime %e" /usr/bin/time --output=tofMatchTracks.log_time ./tofMatchTracks.log_tmp.sh
./tofMatchTracks.log[INFO] Diagnostic switched on since required for calibInfo time
./tofMatchTracks.log[ERROR] invalid workflow in o2-tof-matcher-workflow: No matching output found for TOF/DIAFREQ/0 as requested by data processor "TOFCalibWriter". Candidates:
./tofMatchTracks.log-DPL/ENUM/1635885222
./tofMatchTracks.log-DPL/ENUM/1708084303
./tofMatchTracks.log-DPL/ENUM/3804761563
--
./sim-challenge.log/sw/BUILD/f106642532d4cc663cb8d372851158d3416b86c6/O2-full-system-test/sim-challenge
./sim-challenge.logRunning: TIME="#walltime %e" /usr/bin/time --output=tofMatchTracks.log_time ./tofMatchTracks.log_tmp.sh
./sim-challenge.log[INFO] Diagnostic switched on since required for calibInfo time
./sim-challenge.log[ERROR] invalid workflow in o2-tof-matcher-workflow: No matching output found for TOF/DIAFREQ/0 as requested by data processor "TOFCalibWriter". Candidates:
./sim-challenge.log-DPL/ENUM/1635885222
./sim-challenge.log-DPL/ENUM/1708084303
./sim-challenge.log-DPL/ENUM/3804761563
```
could it be related?

noferini:
No, it is another problem, TOF calibInfo should be no longer required in MC, I will have a look and fix it in FST

noferini:
I open a PR to fix it
https://github.com/AliceO2Group/AliceO2/pull/9522

noferini:
The fix for TOF matching is merged.

knopers8 (approved):

PR #1366, opened 2022-07-26
Add C to allowed languages

davidrohr:
There is a bug in LLVM CMake (for LLVM > 14), which is pulled in since you load the O2 dependencies, which breaks the cmake phase if C is not in the list of allowerd languages.


knopers8 (approved):
thank you!

PR #1368, opened 2022-07-27
Condition to fill hitmaps changed from nTriggers to nTF

IsakovAD:
Condition for filling THnSparse hitmaps were changed from the first 1000 triggers to the first 1 TF. 
Previous criteria required to adjust the threshold based on current IR, while the new one should be more generalized


iravasen:
@IsakovAD thanks a lot! Just one question, the TFid always start from 1? I'm wondering because in FHR task we sample only 1% of the data.

IsakovAD:
@iravasen no, TFid always starts from 0

iravasen:
ok yes you are right. And is a continuous number within the QC task right? I.e. 0,1,2,... (it follows the TF numbering reaching QC only).

IsakovAD:
Yes, each new cycle we have +1 TF. 
I tested the code for 20 TF and the size of ThnSparse object didn't increase in time

iravasen:
GOOD, can I remove WIP and proceed?

iravasen:
@IsakovAD thanks! Comment:
- `mTFCount` is not defined and not incremented. 
- clang-format

IsakovAD:
@iravasen variable mTFCount already was in use in this code. It was defined in the header and is incremented on the line 717

iravasen:
thank you, sorry I missed it. All looks fine to me! Approving

iravasen (approved):

PR #1375, opened 2022-07-30
[EMCAL-769] Add error code for BUNCH_LENGTH_EXCEED errror

mfasDa:
Add handling of error code for bunch length exceeding
channel payload size introduced in
https://github.com/AliceO2Group/AliceO2/pull/9568
In addition fixing range of the minor ALTRO errors based
on the conversion error type to int instead of ordering
in the enum class.


mfasDa:
Needs https://github.com/AliceO2Group/AliceO2/pull/9568

Barthelemy (approved):

PR #1376, opened 2022-07-30
TRD remove spectra causing problems, and configure checks from json

bazinski:
- TRD QC is crashing in 2 spectra pulseheight per chamber (both places) on the fill and ClsDetTime also on fill. These are presumed to follow the "Invalid signature" message from ROOT.
- Allow the checks to be configured from json.

DigitsTask:
driftregionstart : 0 red line on pulseheight plot
driftregionend: 5  red line on pulseheight plot.
pulseheightthreshold : 400 min sum of all 3 adjacent adc to enter the pulseheight, displayed in title
chamberstoignore": "16_3_0 06_1_0" applies to pulseheight generation

PulseHeightCheck:
driftregionstart": "7.0",
driftregionend": "20.0",
pulseheightpeaklower": "0.0",
pulseheightpeakupper": "5.0",

pulseheightminsum": "1500", min counts before starting the check.
pulseheightratio": "1.1", min ratio of peak to average of drift region as defined by the paramters above.

TrackletsCheck: applies to tracklets per timeframe
integralthreshold 10000.0
ratiothreshold": "0.9 - 90% of counts must be above the integralthreshold
zerobinratiothrehold 0.9  90% of counts that must be outside bin 0

### DIFF
--- a/Modules/TRD/src/DigitsTask.cxx
+++ b/Modules/TRD/src/DigitsTask.cxx
@@ -48,8 +48,35 @@ void DigitsTask::retrieveCCDBSettings()
   if (mChamberStatus == nullptr) {
     ILOG(Info, Support) << "mChamberStatus is null, no chamber status to display" << ENDM;
   }
-  if (mChamberStatus == nullptr) {
-    ILOG(Info, Support) << "mChamberStatus is null, no chamber status plotting" << ENDM;
+}
+
+void DigitsTask::buildChamberIgnoreBP()
+{
+  mChambersToIgnoreBP.reset();
+  // Vector of string to save tokens
+  std::vector<std::string> tokens;
+  // stringstream class check1
+  std::stringstream check1(mChambersToIgnore);
+  std::string intermediate;
+  // Tokenizing w.r.t. space ','
+  while (getline(check1, intermediate, ',')) {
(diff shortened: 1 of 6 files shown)


bazinski:
@tdietel @martenole you guys might want to take a look through thiis

Modules/TRD/src/DigitsTask.cxx
```diff
@@ -340,7 +362,7 @@ void DigitsTask::buildHistograms()
     getObjectsManager()->startPublishing(h);
   }
 
-  int cn = 0;
+  /*  int cn = 0;
```
Barthelemy, line 365: we usually advise not to comment out code but to remove it. Git will remember any ways.

Barthelemy (approved):

Modules/TRD/src/DigitsTask.cxx
bazinski: @Barthelemy it was more due to the fact that we have never replicated this problem. Invalid signature is using a 1 parameter update to a 2d spectra, which does not exist in the code. The segfault is clearly on these these std::array of spectra, but its done the same way as other places. So the underlying cause is a mystery. Its left in to be able to trivially put them back for testing.

PR #1378, opened 2022-08-01
[EMCAL-767] add post processing task for Calib

cterrevo:
first post processing task for Calibr monitoring (BadChannel, TimeParam)


mfasDa (approved):
Looks good!

PR #1383, opened 2022-08-02
[EMCAL-524] Make Amplitude cut for Time histo configurable

cterrevo:


mfasDa (approved):
Looks fine as well! Was also tested offline

PR #1386, opened 2022-08-03
GLO: add division by zero check in data compression task

tklemenz:
This adds a check for zero division to the data compression task. If the input size is zero, the compression factor will be shown as zero.


martenole (approved):

tklemenz:
The error is unrelated.

knopers8 (approved):

PR #1388, opened 2022-08-03
TRD remove sharedptr for tracklets and add charge spectra

bazinski:
- removed the shared pointers from trackletstask, a possible reason for the crashes in digits task.
- add charge plots for the 3 charge windows, and also on a per chamber basis.


bazinski:
@martenole this removes the shared ptrs possibly causing problems with digitstask, will do digits after this is merged.

martenole (approved):

knopers8 (approved):

PR #1395, opened 2022-08-08
MFT: reduced number of plots, geometry loaded, cluster position plots added

diana0x0f:
- Removed patternID per ladder (280 histos less)
- Replace digit double column occupancy per ladder for a single histogram (per chip) (280 histos less, 1 new histo)
- Loaded MFT geometry 
- Added histograms for clusters z-position and clusters xy-position per layer

### DIFF
--- a/Modules/MFT/src/QcMFTClusterTask.cxx
+++ b/Modules/MFT/src/QcMFTClusterTask.cxx
@@ -32,6 +32,8 @@
 #include "CCDB/BasicCCDBManager.h"
 #include "CCDB/CCDBTimeStampUtils.h"
 #include "MFTTracking/IOUtils.h"
+#include "MFTBase/GeometryTGeo.h"
+#include <DetectorsBase/GeometryManager.h>
 
 // Quality Control
 #include "QualityControl/QcInfoLogger.h"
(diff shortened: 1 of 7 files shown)


Modules/MFT/src/QcMFTClusterTask.cxx
```diff
@@ -217,6 +232,11 @@ void QcMFTClusterTask::monitorData(o2::framework::ProcessingContext& ctx)
   auto clustersPattern = ctx.inputs().get<gsl::span<unsigned char>>("patterns");
   auto patternIt = clustersPattern.begin();
 
+  // get clusters with global xy position
+  mClustersGlobal.clear();
+  mClustersGlobal.reserve(clusters.size());
+  o2::mft::ioutils::convertCompactClusters(clusters, patternIt, mClustersGlobal, mDict);
+
   if (clusters.size() < 1)
     return;
```
tomas-herman, line 241: I would probably move these 2 lines just after line 229 `const auto clusters = ctx.inputs().get<gsl::span<o2::itsmft::CompClusterExt>>("randomcluster");` .

tomas-herman (approved):

tomas-herman:
Hi @Barthelemy and @knopers8,

I see` build/QualityControl/O2fst/o2` and  `build/QualityControl/o2-cs8` reported as failed but the log file has no errors. I assume they are false errors? 

Please let us know if you spot something real.

Thanks,
Tomas

PR #1401, opened 2022-08-11
[QC-840] Allow to exclude certain directories in QC file merger

knopers8:


knopers8:
marking as WIP because it does not work for multiple arguments

Barthelemy (approved):

PR #1402, opened 2022-08-11
MFT: new plots for the digit task

diana0x0f:
Added new plots: number of digits per ROF (automatic binning according to ROF length); number of digits per time bin; digits per BC


Barthelemy:
I let Tomas review.

tomas-herman (approved):

PR #1291, opened 2022-06-22
TPC: Generic Check of Pads

HorstMa:
This PR is a generic check of the 72 TPC pads for an expected mean value, a comparison of the mean of one pad to the global mean and a check for empty bins inside the pads

PR #1404, opened 2022-08-11
TPC: Generic Check of Pads

HorstMa:
This PR is a generic check of the 72 TPC pads for an expected mean value, a comparison of the mean of one pad to the global mean and a check for empty bins inside the pads
This is the same as #1291, but with the changes implemented.

### DIFF
--- a/Modules/TPC/include/TPC/CheckOfPads.h
+++ b/Modules/TPC/include/TPC/CheckOfPads.h
@@ -1,0 +1,75 @@
+// Copyright CERN and copyright holders of ALICE O2. This software is
+// distributed under the terms of the GNU General Public License v3 (GPL
+// Version 3), copied verbatim in the file "COPYING".
+//
+// See http://alice-o2.web.cern.ch/license for full licensing information.
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+///
+/// \file   CheckOfPads.h
+/// \author Maximilian Horst
+///
+
+#ifndef QC_MODULE_TPC_CheckOfPads_H
+#define QC_MODULE_TPC_CheckOfPads_H
+
+#include "QualityControl/CheckInterface.h"
+
+namespace o2::quality_control_modules::tpc
+{
+
+/// \brief
+///
+/// \author Maximilian Horst
+class CheckOfPads : public o2::quality_control::checker::CheckInterface
+{
+
+ public:
+  /// Default constructor
+  CheckOfPads() = default;
+  /// Destructor
+  ~CheckOfPads() override = default;
+
+  // Override interface
+  void configure() override;
+  Quality check(std::map<std::string, std::shared_ptr<MonitorObject>>* moMap) override;
+  void beautify(std::shared_ptr<MonitorObject> mo, Quality checkResult = Quality::Null) override;
+  std::string getAcceptedType() override;
+
+ private:
+  ClassDefOverride(CheckOfPads, 1);
+  static constexpr std::string_view CheckChoiceMean = "Mean";
+  static constexpr std::string_view CheckChoiceExpectedValue = "ExpectedValue";
+  static constexpr std::string_view CheckChoiceBoth = "Both";
+  std::vector<std::string> mSectorsNameEV;
+  std::vector<Quality> mSectorsQualityEV;
+  std::vector<std::string> mSectorsNameMean;
+  std::vector<Quality> mSectorsQualityMean;
+  std::vector<Quality> mSectorsQualityEmpty;
+  std::vector<std::string> mSectorsName;
+  std::vector<Quality> mSectorsQuality;
+  std::vector<std::string> mMOsToCheck2D;
+  std::string mCheckChoice = "NULL";
+  std::vector<float> mPadMeans;
+  std::vector<float> mPadStdev;
+  std::vector<float> mEmptyPadPercent;
+  float mMediumQualityLimit;
+  float mBadQualityLimit;
+  float mExpectedValue;
+  float mExpectedValueMediumSigmas;
+  float mExpectedValueBadSigmas;
+  float mMeanMediumSigmas;
+  float mMeanBadSigmas;
+  float mTotalMean;
+  float mTotalStdev;
+  bool mEmptyCheck = false;
+  bool mExpectedValueCheck = false;
+  bool mMeanCheck = false;
+};
+
+} // namespace o2::quality_control_modules::tpc
+
+#endif // QC_MODULE_TPC_CheckOfPads_H
--- a/Modules/TPC/src/CheckOfPads.cxx
+++ b/Modules/TPC/src/CheckOfPads.cxx
@@ -1,0 +1,328 @@
+// Copyright CERN and copyright holders of ALICE O2. This software is
+// distributed under the terms of the GNU General Public License v3 (GPL
+// Version 3), copied verbatim in the file "COPYING".
+//
+// See http://alice-o2.web.cern.ch/license for full licensing information.
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+///
+/// \file   CheckOfPads.cxx
+/// \author Laura Serksnyte, Maximilian Horst
+///
+
+#include "TPC/CheckOfPads.h"
+#include "QualityControl/MonitorObject.h"
+#include "QualityControl/Quality.h"
+#include "QualityControl/QcInfoLogger.h"
+#include "Common/Utils.h"
+
+// ROOT
+#include <TCanvas.h>
+#include <TH1.h>
+#include <TH2.h>
+#include <TPad.h>
+#include <TList.h>
+#include <TPaveText.h>
+
+#include <iostream>
+
+namespace o2::quality_control_modules::tpc
+{
+void CheckOfPads::configure()
+{
+  mMediumQualityLimit = common::getFromConfig<float>(mCustomParameters, "mediumQualityPercentageOfWorkingPads", 0.7);
+  mBadQualityLimit = common::getFromConfig<float>(mCustomParameters, "badQualityPercentageOfWorkingPads", 0.3);
+  const std::string checkChoiceString = common::getFromConfig<std::string>(mCustomParameters, "CheckChoice", "Mean");
+
+  if (size_t finder = checkChoiceString.find("ExpectedValue"); finder != std::string::npos) {
+    mExpectedValueCheck = true;
+  }
+  if (size_t finder = checkChoiceString.find("Mean"); finder != std::string::npos) {
+    mMeanCheck = true;
+  }
+  if (size_t finder = checkChoiceString.find("Empty"); finder != std::string::npos) {
+    mEmptyCheck = true;
+  }
+
+  if (mExpectedValueCheck) {
+    // load expectedValue and allowed standard deviations for medium/bad quality
+    mExpectedValue = common::getFromConfig<float>(mCustomParameters, "ExpectedValue", 1.0);
+    mExpectedValueMediumSigmas = common::getFromConfig<float>(mCustomParameters, "ExpectedValueSigmaMedium", 3.0);
+    mExpectedValueBadSigmas = common::getFromConfig<float>(mCustomParameters, "ExpectedValueSigmaBad", 6.0);
+  }
+  // Check if Mean comparison is wished for:
+  if (mMeanCheck) {
+    // load Mean Sigma Medium
+    mMeanMediumSigmas = common::getFromConfig<float>(mCustomParameters, "MeanSigmaMedium", 3.0);
+    mMeanBadSigmas = common::getFromConfig<float>(mCustomParameters, "MeanSigmaBad", 6.0);
+  }
+
+  if (auto param = mCustomParameters.find("MOsNames2D"); param != mCustomParameters.end()) {
+    auto temp = param->second.c_str();
+    std::istringstream ss(temp);
+    std::string token;
+    while (std::getline(ss, token, ',')) {
+      mMOsToCheck2D.emplace_back(token);
+    }
+  }
+}
+
+//______________________________________________________________________________
+Quality CheckOfPads::check(std::map<std::string, std::shared_ptr<MonitorObject>>* moMap)
+{
+  Quality resultEV = Quality::Null;
+  Quality resultMean = Quality::Null;
+  Quality resultGlobal = Quality::Null;
+  Quality resultEmpty = Quality::Null;
+  for (auto const& moObj : *moMap) {
+    auto mo = moObj.second;
+    if (!mo) {
+      continue;
+    }
+    const auto moName = mo->getName();
+    if (auto it = std::find(mMOsToCheck2D.begin(), mMOsToCheck2D.end(), moName); it != mMOsToCheck2D.end()) {
+      const size_t end = moName.find("_2D");
+      const auto histSubName = moName.substr(7, end - 7);
+      resultEV = Quality::Good;
+      resultMean = Quality::Good;
+
+      auto* canv = (TCanvas*)mo->getObject();
+      if (!canv) {
+        continue;
+      }
+      // Check all histograms in the canvas
+
+      mTotalMean = 0.;
+
+      for (int tpads = 1; tpads <= 72; tpads++) {
+        const auto padName = fmt::format("{:s}_{:d}", moName, tpads);
+        const auto histName = fmt::format("h_{:s}_ROC_{:02d}", histSubName, tpads - 1);
+        TPad* pad = (TPad*)canv->GetListOfPrimitives()->FindObject(padName.data());
+        if (!pad) {
+          mSectorsName.push_back("notitle");
+          mSectorsQualityEV.push_back(Quality::Null);
+          continue;
+        }
+        TH2F* h = (TH2F*)pad->GetListOfPrimitives()->FindObject(histName.data());
+        if (!h) {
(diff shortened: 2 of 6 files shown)


Barthelemy:
I'll let @wiechula review.

Modules/TPC/include/TPC/CheckOfPads.h
```diff
@@ -0,0 +1,76 @@
...
+ private:
+  ClassDefOverride(CheckOfPads, 1);
+  static constexpr std::string_view CheckChoiceMean = "Mean";
+  static constexpr std::string_view CheckChoiceExpectedValue = "ExpectedValue";
+  static constexpr std::string_view CheckChoiceBoth = "Both";
+  std::vector<std::string> mSectorsName_EV;
+  std::vector<Quality> mSectorsQuality_EV;
+  std::vector<std::string> mSectorsName_Mean;
+  std::vector<Quality> mSectorsQuality_Mean;
+  std::vector<Quality> mSectorsQuality_Empty;
```
wiechula, line 51: Naming convention
```suggestion
  std::vector<std::string> mSectorsNameEV;
  std::vector<Quality> mSectorsQualityEV;
  std::vector<std::string> mSectorsNameMean;
  std::vector<Quality> mSectorsQualityMean;
  std::vector<Quality> mSectorsQualityEmpty;
```

Modules/TPC/src/CheckOfPads.cxx
```diff
@@ -0,0 +1,377 @@
...
+void CheckOfPads::beautify(std::shared_ptr<MonitorObject> mo, Quality)
+{
+  auto moName = mo->getName();
+  if (auto it = std::find(mMOsToCheck2D.begin(), mMOsToCheck2D.end(), moName); it != mMOsToCheck2D.end()) {
+
+    int padsTotal = 0, padsstart = 1000;
+    auto* tcanv = (TCanvas*)mo->getObject();
+    std::string histNameS, histName;
+    padsstart = 1;
+    padsTotal = 72;
```
wiechula, line 315:
```suggestion
    auto* tcanv = (TCanvas*)mo->getObject();
    const int padsstart = 1;
    const int padsTotal = 72;
```

Modules/TPC/src/CheckOfPads.cxx
```diff
@@ -0,0 +1,377 @@
...
+    size_t end = moName.find("_2D");
+    auto histSubName = moName.substr(7, end - 7);
+    histNameS = fmt::format("h_{}_ROC", histSubName);
```
wiechula, line 318:
```suggestion
    const size_t end = moName.find("_2D");
    const auto histSubName = moName.substr(7, end - 7);
    const auto histNameS = fmt::format("h_{}_ROC", histSubName);
```

Modules/TPC/src/CheckOfPads.cxx
```diff
@@ -0,0 +1,377 @@
...
+
+      const std::string padName = fmt::format("{:s}_{:d}", moName, tpads);
+      TPad* pad = (TPad*)tcanv->GetListOfPrimitives()->FindObject(padName.data());
+      if (!pad) {
+        continue;
+      }
+      pad->cd();
+      TH1F* h = nullptr;
+      histName = fmt::format("{}_{:02d}", histNameS, tpads - 1);
+      h = (TH1F*)pad->GetListOfPrimitives()->FindObject(histName.data());
```
wiechula, line 329:
```suggestion
      const auto histName = fmt::format("{}_{:02d}", histNameS, tpads - 1);
      const auto h = (TH1F*)pad->GetListOfPrimitives()->FindObject(histName.data());
```

Modules/TPC/src/CheckOfPads.cxx
```diff
@@ -0,0 +1,377 @@
...
+#include <TList.h>
+#include <TPaveText.h>
+
+#include <iostream>
+
+namespace o2::quality_control_modules::tpc
+{
+void CheckOfPads::configure()
+{
+  if (auto param = mCustomParameters.find("mediumQualityPercentageOfWorkingPads"); param != mCustomParameters.end()) {
```
wiechula, line 35:
```suggestion
  if (const auto param = mCustomParameters.find("mediumQualityPercentageOfWorkingPads"); param != mCustomParameters.end()) {
```

Modules/TPC/src/CheckOfPads.cxx
```diff
@@ -0,0 +1,377 @@
...
+    mMediumQualityLimit = std::atof(param->second.c_str());
+  } else {
+    mMediumQualityLimit = 0.7;
+    ILOG(Info, Support) << "Chosen check requires mediumQualityPercentageOfWorkingPads which is not given. Setting to default 0.7 ." << ENDM;
+  }
+  if (auto param = mCustomParameters.find("badQualityPercentageOfWorkingPads"); param != mCustomParameters.end()) {
```
wiechula, line 41:
```suggestion
  if (const 
  auto param = mCustomParameters.find("badQualityPercentageOfWorkingPads"); param != mCustomParameters.end()) {
```

Modules/TPC/src/CheckOfPads.cxx
wiechula, line 35: Here is a lot of copy'n'paste. And there is already a function for this: https://github.com/AliceO2Group/QualityControl/blob/352b520fbc9187ed7e857e263946d3c6b0580ce6/Modules/Common/include/Common/Utils.h
Used e.g. here: https://github.com/AliceO2Group/QualityControl/blob/master/Modules/TPC/src/PID.cxx#L46
Perhaps you can patch `Common/Utils.h` with
```diff
diff --git a/Modules/Common/include/Common/Utils.h b/Modules/Common/include/Common/Utils.h
index ad0127ca..15b14e8b 100644
--- a/Modules/Common/include/Common/Utils.h
+++ b/Modules/Common/include/Common/Utils.h
@@ -30,15 +30,14 @@ namespace o2::quality_control_modules::common
 /// \param name Name of the taskParameter
 /// \return taskParameter converted to bool, int, float, double or std::string depending on the template type
 template <typename T>
-T getFromConfig(const std::unordered_map<std::string, std::string>& params, const std::string_view name)
+T getFromConfig(const std::unordered_map<std::string, std::string>& params, const std::string_view name, T retVal = T{})
 {
   const auto last = params.end();
   const auto itParam = params.find(name.data());
-  T retVal{};
 
   if (itParam == last) {
     LOGP(warning, "missing parameter {}", name.data());
-    LOGP(warning, "Please add '{}': '<value>' to the 'taskParameters'.", name.data());
+    LOGP(warning, "Please add '{}': '<value>' to the 'taskParameters'. Using default value {}.", name.data(), retVal);
   } else {
     const auto& param = itParam->second;
     if constexpr (std::is_same<int, T>::value) {
```
and update all occurrances below with something like:
```c++
mMediumQualityLimit = common::getFromConfig<float>(mCustomParameters, "mediumQualityPercentageOfWorkingPads", 0.7);
```

Modules/TPC/src/CheckOfPads.cxx
```diff
@@ -0,0 +1,377 @@
...
+  Quality resultMean = Quality::Null;
+  Quality resultGlobal = Quality::Null;
+  Quality resultEmpty = Quality::Null;
+  for (auto const& moObj : *moMap) {
+    auto mo = moObj.second;
+    if (!mo) {
+      continue;
+    }
+    auto moName = mo->getName();
+    std::string histName, histNameS;
```
wiechula, line 135:
```suggestion
    const auto moName = mo->getName();
```

Modules/TPC/src/CheckOfPads.cxx
```diff
@@ -0,0 +1,377 @@
...
+    if (auto it = std::find(mMOsToCheck2D.begin(), mMOsToCheck2D.end(), moName); it != mMOsToCheck2D.end()) {
+      size_t end = moName.find("_2D");
+      auto histSubName = moName.substr(7, end - 7);
```
wiechula, line 138:
```suggestion
      const size_t end = moName.find("_2D");
      const auto histSubName = moName.substr(7, end - 7);
```

Modules/TPC/src/CheckOfPads.cxx
```diff
@@ -0,0 +1,377 @@
...
+                  padStdev += pow(binvalue - padMean, 2);
+                }
+              }
+            }
+          }
+
+          if (runNo == 1 && padSum > 0.) { // no need for div by 0 check, because if padSum > 0 -> padsCount>0
+            padMean = padSum / padsCount;
+          }
+        }
```
wiechula, line 206: No need to loop twice
```suggestion
          for (int xBin = 1; xBin <= MaximumXBin; xBin++) {
            for (int yBin = 1; yBin <= MaximumYBin; yBin++) {
              const float binvalue = h->GetBinContent(xBin, yBin);
              if (binvalue != 0) {
                  padMean += binvalue;
                  padStdev += binvalue * binvalue;
                  padsCount++;
              }
            }
          }
        }
       if (padsCount>0) {
          padMean /= padsCount;
          padStdev = std::sqrt(std::abs(padStdev / padsCount - mean * mean));
        }
```

Modules/TPC/src/CheckOfPads.cxx
```diff
@@ -0,0 +1,377 @@
...
+        const int MaximumYBin = h->GetNbinsY();
+        if (titleh.find("IROC") != std::string::npos) {
+          totalPads = 5280;
+        } else if (titleh.find("OROC") != std::string::npos) {
+          totalPads = 9280;
+        } else {
+          return Quality::Null;
+        }
+
+        float padSum = 0.;
```
wiechula, line 180: Not needed
```suggestion
```

Modules/TPC/src/CheckOfPads.cxx
```diff
@@ -0,0 +1,377 @@
...
+          } else if (padsCount < mBadQualityLimit * totalPads) {
+            resultEmpty = Quality::Bad;
+          } else {
+            resultEmpty = Quality::Medium;
+          }
+          mSectorsQuality_Empty.push_back(resultEmpty);
+          mEmptyPadPercent.push_back(1. - (float)padsCount / totalPads);
+        }
+
+        padStdev = sqrt(padStdev / (padsCount - 1));
```
wiechula, line 220: Not needed, or to be modified above.
```suggestion
```

Modules/TPC/src/CheckOfPads.cxx
```diff
@@ -0,0 +1,377 @@
...
+        if (mEmptyCheck) {
+          if (padsCount > mMediumQualityLimit * totalPads) {
+            resultEmpty = Quality::Good;
+          } else if (padsCount < mBadQualityLimit * totalPads) {
+            resultEmpty = Quality::Bad;
+          } else {
+            resultEmpty = Quality::Medium;
+          }
+          mSectorsQuality_Empty.push_back(resultEmpty);
+          mEmptyPadPercent.push_back(1. - (float)padsCount / totalPads);
```
wiechula, line 217: totalPads can be `0` if title does not contain `IROC` or `OROC`. Either take care above and directly continue, or check for division by 0.

Modules/TPC/src/CheckOfPads.cxx
```diff
@@ -0,0 +1,377 @@
...
+        sumOfWeights += 1 / mPadStdev[it];
+      }
+      mTotalStdev = sqrt(1 / sumOfWeights); // standard deviation of the weighted average.
+      mTotalMean /= sumOfWeights;           // Weighted average (by standard deviation) of the total mean
+      // calculate the Qualities:
+
+      for (size_t it = 0; it < mPadMeans.size(); it++) { // loop over all pads
+
+        if (mExpectedValueCheck) {
+
```
wiechula, line 237:
```suggestion
        if (mExpectedValueCheck) {
```

Modules/TPC/src/CheckOfPads.cxx
```diff
@@ -0,0 +1,377 @@
...
+          } else if (std::abs(mPadMeans[it] - mExpectedValue) >= mPadStdev[it] * mExpectedValueMediumSigmas && std::abs(mPadMeans[it] - mExpectedValue) < mPadStdev[it] * mExpectedValueBadSigmas) {
+            resultEV = Quality::Medium;
+          } else {
+            resultEV = Quality::Bad;
+          }
+        }
+
+        mSectorsQuality_EV.push_back(resultEV);
+        if (mMeanCheck) {
+
```
wiechula, line 249:
```suggestion
        if (mMeanCheck) {
```

Modules/TPC/src/CheckOfPads.cxx
```diff
@@ -0,0 +1,377 @@
...
+        }
+
+        if (mExpectedValueCheck) {
+          qualities.push_back(resultEV);
+        }
+
+        if (mEmptyCheck) {
+          qualities.push_back(mSectorsQuality_Empty[it]);
+        }
+        if (qualities.size() > 0) {
```
wiechula, line 273:
```suggestion

        if (qualities.size() > 0) {
```

wiechula (changes requested):

Modules/TPC/src/CheckOfPads.cxx
```diff
@@ -0,0 +1,337 @@
...
+
+#include <iostream>
+
+namespace o2::quality_control_modules::tpc
+{
+void CheckOfPads::configure()
+{
+  mMediumQualityLimit = common::getFromConfig<float>(mCustomParameters, "mediumQualityPercentageOfWorkingPads", 0.7);
+  mBadQualityLimit = common::getFromConfig<float>(mCustomParameters, "badQualityPercentageOfWorkingPads", 0.3);
+  std::string CheckChoiceString = common::getFromConfig<std::string>(mCustomParameters, "CheckChoice", "Mean");
```
wiechula, line 38: Naming convention and logic
```suggestion
  std::string checkChoiceString = common::getFromConfig<std::string>(mCustomParameters, "CheckChoice", "Mean");
```
If the default value is `Mean`, the `mFoundCheck` will always be true. Is this the intended behaviour? If not, I think you should simple remove the `, "Mean"` part. Otherwise you can remove the lines 53-56 below, because this part of the code will never be executed...

Also `mFoundCheck` only seems to be used in this function, so no need for a member variable... Depending on the question above, it might even be eliminated completely.

Modules/TPC/src/CheckOfPads.cxx
```diff
@@ -0,0 +1,337 @@
...
+  if (size_t finder = CheckChoiceString.find("Empty"); finder != std::string::npos) {
+    mEmptyCheck = true;
+    mFoundCheck = true;
+  }
+
+  if (!mFoundCheck) {
+    mMeanCheck = true;
+    ILOG(Warning, Support) << "This Check requires a CheckChoice. There was no value or the given value is wrong or not readable. Chose between 'ExpectedValue'(Compare the pad mean to an expectedValue), 'Mean' (compare pad mean to global mean) and/or 'Empty' (check for a threshold of empty pads). As a default 'Mean' was selected." << ENDM;
+  }
+  // }
```
wiechula, line 57:
```suggestion
```

Modules/TPC/src/CheckOfPads.cxx
```diff
@@ -0,0 +1,337 @@
...
+  }
+  if (size_t finder = CheckChoiceString.find("Empty"); finder != std::string::npos) {
+    mEmptyCheck = true;
+    mFoundCheck = true;
+  }
+
+  if (!mFoundCheck) {
+    mMeanCheck = true;
+    ILOG(Warning, Support) << "This Check requires a CheckChoice. There was no value or the given value is wrong or not readable. Chose between 'ExpectedValue'(Compare the pad mean to an expectedValue), 'Mean' (compare pad mean to global mean) and/or 'Empty' (check for a threshold of empty pads). As a default 'Mean' was selected." << ENDM;
+  }
```
wiechula, line 56: Why not
```suggestion
  bool foundCheck = false;
  if (checkChoiceString == CheckChoiceExpectedValue) {
    mExpectedValueCheck = true;
    foundCheck = true;
  }
  if (checkChoiceString == CheckChoiceMean) {
    mMeanCheck = true;
    foundCheck = true;
  }
  if (checkChoiceString == CheckChoiceEmpty) {
    mEmptyCheck = true;
    foundCheck = true;
  }
  
    if (!foundCheck) {
    mMeanCheck = true;
    ILOG(Warning, Support) << fmt::format("This Check requires a CheckChoice. There was no value or the given value is wrong or not readable. Chose between '{}'(Compare the pad mean to an expectedValue), '{}' (compare pad mean to global mean) and/or '{}' (check for a threshold of empty pads). As a default '{}' was selected.", CheckChoiceExpectedValue, CheckChoiceMean, CheckChoiceEmpty, CheckChoiceMean) << ENDM;
  }
```

and remove `mFoundCheck`.

wiechula (changes requested):

Modules/TPC/src/CheckOfPads.cxx
wiechula: what about `const` and naming convention?

Modules/TPC/src/CheckOfPads.cxx
wiechula: what about the simplification using direct string comparison?

HorstMa: this would not work since we want to be able to run multiple checks at the same time (e.g. empty pad check and comparison against an expected value). So the checkChoiceString could look something like "ExpectedValue,Empty" so a direct string comparison would not work.

Barthelemy:
we need a new approval

Modules/TPC/src/CheckOfPads.cxx
wiechula: Ahh, thanks for the explanation! That makes sense.

wiechula (approved):

Barthelemy (approved):

PR #1406, opened 2022-08-12
[EMCAL-524] Add histograms montioring the bad cell calibration

mfasDa:
- Fraction of bad cells / event and supermodule
- Spectra bad cells / supermodule
- Occupancy histograms good and bad cells


mfasDa:
This is needed for monitoring the bad channel calibration. We need to know at least

- What fraction of the cells we record are masked bad by the online calibration, in which detector segment
- What is the spectrum of the cells flagged as bad (integrated) compared to the cells flagged as good
- What is the count rate of bad cells, do we see certain hot spots

This is need when putting the bad channel calibration into production mode, probably next week. We cannot let the calibration run blindly. However all the new histograms must be enabled manually via a task parameter.

mfasDa:
Tested offline, WIP label remove. Ready for merging.

knopers8 (approved):

PR #1407, opened 2022-08-12
[FIT] Refactoring by merging tasks

sbysiak:
WIP, to be merged after adding symmetric changes for FV0


Modules/FT0/include/FT0/DigitQcTask.h
```diff
@@ -107,6 +109,37 @@ class DigitQcTask final : public TaskInterface
...
+  };
+  enum TrgModeThresholdVar { kAmpl,
+                             kNchannels
+  };
+  enum TrgComparisonResult { kSWonly,
+                             kTCMonly,
+                             kNone,
+                             kBoth
+  };
+  const int mNChannelsA = 96;
```
sbysiak, line 126: do we have it defined somewhere already?

Modules/FT0/src/DigitQcTask.cxx
```diff
@@ -380,6 +475,19 @@ void DigitQcTask::monitorData(o2::framework::ProcessingContext& ctx)
       }
 
       setFEEmodules.insert(mChID2PMhash[chData.ChId]);
+
+      if (chData.QTCAmpl <= 0)
+        continue;
```
sbysiak, line 480: Is this correct to skip channels with negative amplitudes when re-computing triggers based on data from PMs?

sbysiak:
@mslupeck @afurs please comment on my questions above

Modules/FT0/src/DigitQcTask.cxx
afurs: this is for vertex calculations, but this condition is not enough.

Modules/FT0/include/FT0/DigitQcTask.h
afurs: https://github.com/AliceO2Group/AliceO2/blob/6061027f33dbf3332e41b5e5ef5f72373bf93d7f/Detectors/FIT/FT0/base/include/FT0Base/Geometry.h#L52-L53
Also there are should be limitation for C side for ref channel suppressing in case of vertex calculation.

Modules/FT0/src/DigitQcTask.cxx
mslupeck: Please talk to Jacek - he had a long discussion with Dima about it and I think he can share the email with you, in which it is clarified what conditions should be used for each trigger type.

Modules/FT0/include/FT0/DigitQcTask.h
mslupeck: @afurs should we put it into `FT0Base/Geometry.h`, or to consul config? I think config would be better because channel map can change, so we should just have a list of channels in consul that corresponds to reference PMT channels, and most QC tasks should ignore them completely

Modules/FT0/src/DigitQcTask.cxx
sbysiak: @jotwinow

Modules/FT0/include/FT0/DigitQcTask.h
sbysiak: constant replace with the ones from Geometry.h and limitation for C added

Modules/FT0/src/DigitQcTask.cxx
sbysiak: We will need to re-iterate on this, somehow I cannot get agreement between the logic you provided and what is recorded in TCM.

sbysiak:
I think we can merge this PR. The rest of current changes (checkers and adjustment of FDD) will come separately.

Regarding functionality from TriggerQcTask: current version works more less fine (for FV0 agreement in ~95% of cases for each trigger) and I had hard time improving this result in local tests, so it requires more effort and time.

afurs:
Hi @sbysiak , I left several comments. Some of them is important(e.g. double loops https://github.com/AliceO2Group/QualityControl/blob/0bdb85249e204df38ed3193e557ae1d52950a577/Modules/FT0/src/DigitQcTask.cxx#L573-L577) for processing at FLP side.

sbysiak:
Hi @afurs ,
I see only this one comment. Do you need more time for the review?

mslupeck:
Hi @Barthelemy @knopers8 
Could you merge it?

Barthelemy (approved):

PR #1412, opened 2022-08-16
Protect from having EoS callback invoked before Start

knopers8:


knopers8:
FST failure unrelated

PR #1422, opened 2022-08-26
ITS - fix bugs in QC checks from 1st review

iravasen:
Fix of simple bugs in FEE and track task


Barthelemy:
The error in the mac build seems legit.

iravasen:
@Barthelemy yes I forgot to push one thing. Now should be ok but better to wait for the test before merging.

iravasen:
@Barthelemy now it looks ok. To me it can be merged.

Barthelemy (approved):

PR #1428, opened 2022-08-31
FIT: new checker + timestamp usage in PP + adjustments

sbysiak:
Major changes:
- Add new checker `GenericCheck` for FT0/FV0/FDD
- use timestamps when retrieving filling scheme in PP (options for running online and offline)

Other minor changes:
- `AmpVsTime` histos can be preserved separately from other per channel histos as this correlation cannot be retrieved based on 2D histos like `AmpPerChannel`
- use common constants and proper number of channels for FV0
- fix missing obj handling in rebinHisto()


Barthelemy (approved):
ok for me but I still think that it should somehow be possible to avoid duplicating code between the 3 FIT components

PR #1432, opened 2022-09-02
[O2-3145] Add label ccdb to some tests

Barthelemy:
This way, one can easily exclude them in the PRs of alidist


knopers8:
there is a commit which you did not want to include here probably

knopers8 (approved):

PR #1434, opened 2022-09-05
MFT: fetch cluster dictionary from DPL and pattern iterator edit

diana0x0f:
- cluster dictionary is fetched from DPL
- task was crashing with the use of pattern iterator for grouped clusters, because the iterator was already used in convertCompactClusters -> a new iterator was added

### DIFF
--- a/Modules/MFT/src/QcMFTClusterTask.cxx
+++ b/Modules/MFT/src/QcMFTClusterTask.cxx
@@ -261,13 +261,11 @@ void QcMFTClusterTask::startOfCycle()
 
 void QcMFTClusterTask::monitorData(o2::framework::ProcessingContext& ctx)
 {
-  static bool initOnceDone = false;
-  if (!initOnceDone) { // the dictionary needs to be queried only once
+  if (mDict == nullptr){
     std::cout << "Getting dictionary from CCDB" << std::endl;
     ILOG(Info, Support) << "Getting dictionary from ccdb" << ENDM;
-    initOnceDone = true;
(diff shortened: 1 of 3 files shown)


Modules/MFT/src/QcMFTClusterTask.cxx
```diff
@@ -308,7 +307,8 @@ void QcMFTClusterTask::monitorData(o2::framework::ProcessingContext& ctx)
     if (oneCluster.getPatternID() != o2::itsmft::CompCluster::InvalidPatternID && !mDict->isGroup(oneCluster.getPatternID())) {
       mClusterSizeSummary->Fill(mDict->getNpixels(oneCluster.getPatternID()));
     } else {
-      o2::itsmft::ClusterPattern patt(patternIt);
+      auto patternIt2 = clustersPattern.begin();
```
shahor02, line 310: One should start from `clustersPattern.begin();` once and then increment the same iterator in the cluster loop. Could you move this line before https://github.com/AliceO2Group/QualityControl/pull/1434/files#diff-5a078c880ac63735b04899b111617d16021b9589832e1f8a90cd601273318c73R296? Or, you can simply reuse the patternIt after resetting it to `clustersPattern.begin()` instead of adding a new iterator.

shahor02 (changes requested):

Modules/MFT/src/QcMFTClusterTask.cxx
diana0x0f: Hi Ruben, thanks a lot, I edited the code.

shahor02 (approved):

tomas-herman (approved):

PR #1435, opened 2022-09-05
[EMCAL-524] CellTime histos for each BC, merged for all SM

cterrevo:


mfasDa (approved):
Looks good

mfasDa:
@Barthelemy This reduces 100 histograms to 5. They were up to now indeed needed per supermodule, because in run2 the so-called L1 phase was different between supermodules (not only in time), and with the data taken in the pilot beams we needed to verify that the phase is stable in time and per supermodule. Since this is now validated we can remove the histograms per supermodule and keep only the the projection over all supermodules for the 4 BC phases (+1 for the calibration trigger which has a fixed phase).

PR #1436, opened 2022-09-06
CPV: fix for merging IntensiveTH2F

sevdokim:
Hello! This is fix for proper merging of IntensiveTH2F in CPV/PhysicsTask.

### DIFF
--- a/Modules/CPV/include/CPV/PhysicsTask.h
+++ b/Modules/CPV/include/CPV/PhysicsTask.h
@@ -52,12 +52,22 @@ class IntensiveTH2F : public TH2F, public o2::mergers::MergeInterface
   }
 
   void setCycleNumber(uint32_t cycleNumber) { mCycleNumber = cycleNumber; }
+  uint32_t getCylceNumber() { return mCycleNumber; }
 
   void merge(MergeInterface* const other) override
   {
-    if (mCycleNumber < dynamic_cast<IntensiveTH2F*>(other)->mCycleNumber) {
-      LOG(info) << "IntensiveTH2F::merge() :" << GetName() << "is updating!";
-      *this = *(dynamic_cast<IntensiveTH2F*>(other));
+    if (IntensiveTH2F* cast = dynamic_cast<IntensiveTH2F*>(other); cast != nullptr) {
+      if (mCycleNumber < cast->mCycleNumber) {
+        mCycleNumber = cast->mCycleNumber;
+        for (int iX = 1; iX <= GetNbinsX(); iX++) {
+          for (int iY = 1; iY <= GetNbinsY(); iY++) {
+            SetBinContent(iX, iY, cast->GetBinContent(iX, iY));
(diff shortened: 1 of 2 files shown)


Modules/CPV/include/CPV/PhysicsTask.h
```diff
@@ -52,14 +52,28 @@ class IntensiveTH2F : public TH2F, public o2::mergers::MergeInterface
...
+      mCycleNumber = dynamic_cast<IntensiveTH2F*>(other)->mCycleNumber;
+      LOG(debug) << "other.fN = " << (*(dynamic_cast<IntensiveTH2F*>(other))).fN;
+      LOG(debug) << "other.binContent(1, 1) = " << (*(dynamic_cast<IntensiveTH2F*>(other))).GetBinContent(1, 1);
+      LOG(debug) << "this->fN = " << fN;
+      LOG(debug) << "this->binContent(1, 1) = " << GetBinContent(1, 1);
+      for (int iX = 1; iX <= GetNbinsX(); iX++) {
+        for (int iY = 1; iY <= GetNbinsY(); iY++) {
+          SetBinContent(iX, iY, dynamic_cast<IntensiveTH2F*>(other)->GetBinContent(iX, iY));
+        }
+      }
```
knopers8, line 73: Perhaps you could also do `dynamic_cast` just once at the beginning and then make sure it is not `nullptr`?
Also, will you need these debug logs for the future?

sevdokim: hi @knopers8 
I'm not sure how something can be `nullptr` in the merge method (asuming `this` and `other` are valid objects). But yes, you are right, it is better to use dynamic_cast once and check it for being `nullptr`.

knopers8: It is rather unlikely, but we've seen very unlikely events happening on grid, e.g. one of 10000 subjobs had memory issues, preventing ROOT to load correctly and generating corrupted objects. `other` could be invalid if it is `MergeInterface` but it is not `IntensiveTH2F`. It probably won't happen, but better safe than sorry.

Also, `dynamic_cast` has some overhead as compared to `reinterpret_cast`, so it's better to do it once for performance reasons.

sevdokim:
Hi @knopers8 
I tried to implement what you proposed. However O2fst check failed without logs. Is this foreseen?

TimoWilken:
Hi @sevdokim @knopers8, sorry, this is due to a connection issue (the error was: `Error on fetching 'EVTGEN': Error 128 from git ls-remote --heads --tags https://phab.hepforge.org/source/evtgen: fatal: unable to access 'https://phab.hepforge.org/source/evtgen/': Failed connect to phab.hepforge.org:443; Connection timed out. Aborting.`).

The check is currently re-running; it should be done soon.

Barthelemy (approved):

PR #1438, opened 2022-09-06
ITS FHR online task: filtering of noise pixel vector 

IsakovAD:
Fixed the problem that FHR task used too large RAM/CPU after increased IR

1) Vector of pixels _mHitPixelID_InStave_ now is pruned at each cycle of Monitoring function
2)  _mOccupancyPlot_ now shows 1D distribution of noise pixel occupancy
3) _mStaveHitmap_ now is filled only once

### DIFF
--- a/Modules/ITS/src/ITSFhrTask.cxx
+++ b/Modules/ITS/src/ITSFhrTask.cxx
@@ -516,12 +516,33 @@ void ITSFhrTask::monitorData(o2::framework::ProcessingContext& ctx)
       for (auto& digit : digVec[istave][0]) {
         int chip = digit.getChipIndex() % 9;
         mHitPixelID_InStave[istave][0][chip][1000 * digit.getColumn() + digit.getRow()]++;
+        if (mTFCount <= mCutTFForSparse) {
+              int pixelPos[2] = { digit.getColumn() + (1024 * chip) + 1, digit.getRow() + 1 };
+              int BinContent;
+              mStaveHitmap[istave]->GetBinContent(mStaveHitmap[istave]->GetBin(pixelPos),&BinContent); 
+              mStaveHitmap[istave]->SetBinContent(pixelPos,BinContent+1);
+        }
       }
     } else {
       for (int ihic = 0; ihic < nHicPerStave[mLayer]; ihic++) {
         for (auto& digit : digVec[istave][ihic]) {
           int chip = ((digit.getChipIndex() - ChipBoundary[mLayer]) % (14 * nHicPerStave[mLayer])) % 14;
           mHitPixelID_InStave[istave][ihic][chip][1000 * digit.getColumn() + digit.getRow()]++;
+          int ilink = ihic / (nHicPerStave[mLayer] / 2);
+          if (mTFCount <= mCutTFForSparse){
+          if (chip < 7) {
+                  int pixelPos[2] = { (ihic % (nHicPerStave[mLayer] / NSubStave[mLayer]) * ((nChipsPerHic[mLayer] / 2) * NCols)) + chip * NCols + digit.getColumn() + 1, NRows - digit.getRow() - 1 + (1024 * ilink) + 1 };
+                  int BinContent;
+                  mStaveHitmap[istave]->GetBinContent(mStaveHitmap[istave]->GetBin(pixelPos),&BinContent); 
+                  mStaveHitmap[istave]->SetBinContent(pixelPos,BinContent+1);
+          } else {
+                  int pixelPos[2] = { (ihic % (nHicPerStave[mLayer] / NSubStave[mLayer]) * ((nChipsPerHic[mLayer] / 2) * NCols)) + (nChipsPerHic[mLayer] / 2) * NCols - (chip - 7) * NCols - digit.getColumn(), NRows + digit.getRow() + (1024 * ilink) + 1 };
+              int BinContent;    
+              mStaveHitmap[istave]->GetBinContent(mStaveHitmap[istave]->GetBin(pixelPos),&BinContent); 
+              mStaveHitmap[istave]->SetBinContent(pixelPos,BinContent+1);   
+             }
+         }
+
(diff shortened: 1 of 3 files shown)


Modules/ITS/src/ITSFhrTask.cxx
```diff
@@ -599,29 +613,40 @@ void ITSFhrTask::monitorData(o2::framework::ProcessingContext& ctx)
         for (int ihic = 0; ihic < ((nHicPerStave[mLayer] / NSubStave[mLayer])); ihic++) {
           for (int ichip = 0; ichip < nChipsPerHic[mLayer]; ichip++) {
             if (GBTLinkInfo->statistics.nTriggers > 0) {
+
+              if (mDoHitmapFilter == 1 && mTFCount > mCutTFForSparse) {
```
iravasen, line 617: same as above

Modules/ITS/src/ITSFhrTask.cxx
```diff
@@ -599,29 +613,40 @@ void ITSFhrTask::monitorData(o2::framework::ProcessingContext& ctx)
...
+                for (auto iter = mHitPixelID_InStave[istave][ihic + ilink * ((nHicPerStave[mLayer] / NSubStave[mLayer]))][ichip].begin(); iter != mHitPixelID_InStave[istave][ihic + ilink * ((nHicPerStave[mLayer] / NSubStave[mLayer]))][ichip].end();) {
+                  if ((double)iter->second / GBTLinkInfo->statistics.nTriggers < 4.3e-5) { // 1 hits/cm^2 * 5 pixels/hits * 4.5 cm^2 / 1024 / 512 = 4.3e-5/pixel/event`
```
iravasen, line 619: I would put 4.3e-5 as parameter in the json

Modules/ITS/src/ITSFhrTask.cxx
```diff
@@ -562,24 +564,36 @@ void ITSFhrTask::monitorData(o2::framework::ProcessingContext& ctx)
...
+              } else
+                ++iter;
+            }
+          }
+
           for (iter = mHitPixelID_InStave[istave][0][ichip].begin(); iter != mHitPixelID_InStave[istave][0][ichip].end(); iter++) {
             if ((iter->second > mHitCutForNoisyPixel) &&
                 (iter->second / (double)GBTLinkInfo->statistics.nTriggers) > mOccupancyCutForNoisyPixel &&
                 ((double)GBTLinkInfo->statistics.nTriggers >= 1e6 && (double)GBTLinkInfo->statistics.nTriggers < 1e6 + 10000)) {
               mNoisyPixelNumber[mLayer][istave]++; // count only in 10000 events as soon as nTriggers is 1e6
```
iravasen, line 583: Since now you are removing the non-noisy pixels, we can count the noisy pixels w/o waiting for 1M events now (at least when filtering is applied) so to have an idea how many they are (i.e. how many survive the cut) from the TH2Poly.

Modules/ITS/src/ITSFhrTask.cxx
```diff
@@ -562,24 +564,36 @@ void ITSFhrTask::monitorData(o2::framework::ProcessingContext& ctx)
         if ((double)GBTLinkInfo->statistics.nTriggers >= 1e6 && (double)GBTLinkInfo->statistics.nTriggers < 1e6 + 10000) {
           mNoisyPixelNumber[mLayer][istave] = 0;
         }
-
         for (int ichip = 0 + (ilink * 3); ichip < (ilink * 3) + 3; ichip++) {
           std::unordered_map<unsigned int, int>::iterator iter;
+
+          if (mDoHitmapFilter == 1 && mTFCount > mCutTFForSparse) {
```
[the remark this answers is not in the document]
IsakovAD: We use the same _mHitPixelID_InStave_ vector of digits to draw THnSparse hitmaps. So I decided not to filter vector before the 2D hitmap will not be filled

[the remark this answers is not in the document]
iravasen: ah I see now! But then you need the `<` sign when filling the sparse. Probably it is still better to separate the variables but I let you think about it.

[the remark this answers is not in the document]
IsakovAD: If we have "<" sign there, it would mean that we fill ThnSparse by the same vector each TF until  _mTFCount_ will not became larger _mCutTFForSparse_. I think that it is enough to fill it just once when _mTFCount_ == _mCutTFForSparse_, because we are using function SetBinContent(), and not just a Fill()

Modules/ITS/src/ITSFhrTask.cxx
```diff
@@ -516,12 +516,33 @@ void ITSFhrTask::monitorData(o2::framework::ProcessingContext& ctx)
       for (auto& digit : digVec[istave][0]) {
         int chip = digit.getChipIndex() % 9;
         mHitPixelID_InStave[istave][0][chip][1000 * digit.getColumn() + digit.getRow()]++;
+        if (mTFCount <= mCutTFForSparse) {
+              int pixelPos[2] = { digit.getColumn() + (1024 * chip) + 1, digit.getRow() + 1 };
+              int BinContent;
+              mStaveHitmap[istave]->GetBinContent(mStaveHitmap[istave]->GetBin(pixelPos),&BinContent); 
+              mStaveHitmap[istave]->SetBinContent(pixelPos,BinContent+1);
```
iravasen: Why don't you use `mStaveHitmap[istave]->Fill(pixelPos)` ? Same for the other places.

IsakovAD: I thought that ThnSparse does not have Fill() function, but it appeared that It was necessary to fill *Double_t not *int_t. Fixed

iravasen:
Thanks @IsakovAD! Looks good to me

iravasen (approved):

PR #1441, opened 2022-09-07
[EMCAL-711] Fix parameter name for excluding gain errors from summary

mfasDa:
Copy/Paste error in parameter name leading to meaningless and confusing parameter name. Replacing with meaningfull name.


Barthelemy (approved):

Barthelemy (approved):

mfasDa:
There was a second bug in the condition for filling the histograms DDL vs. FEC ID that the error type was missing in the condition, therefore all histograms were showing the same FECs of all error types. Just realized this when looking at the raw error layout where the histogram for the Minor ALTRO errors was showing the FECs of the gain errors since the error counter of the Minor ALTRO errors was 0.

PR #1443, opened 2022-09-08
[WIP] ITS post-processing: fixed memory leaks in FHR and Cluster PP tasks

IsakovAD:
Fixed the problem with DrawClone() function, which was creating a clone of histogram without deleting it. Now we are using the multigraph


iravasen:
Hi @IsakovAD thanks a lot! Can you please apply the same fix to: https://github.com/AliceO2Group/QualityControl/blob/master/Modules/ITS/src/TrendingTaskITSThr.cxx? I see a DrawClone also there unfortunately.

Barthelemy:
@JianLIUhep @ktf any idea why O2fst/o2 is failing due to a crash in `qc-task-ITS-ITSClusterTask` ?

JianLIUhep:
Hi @Barthelemy, thanks for reporting this issue. I have just updated the QC json for the cluster task in O2DPG. It should work as soon as the PR is merged. https://github.com/AliceO2Group/O2DPG/pull/620

Barthelemy (approved):

PR #1447, opened 2022-09-12
Documentation: single data source of any kind allowed to beautify

knopers8:


Barthelemy:
@TimoWilken 
There is a problem with the build : 
```
## sw/BUILD/O2-full-system-test-latest/log
--
command alien-token-init had nonzero exit code 2
FATAL: No alien token present
--
```

Barthelemy (approved):

PR #1449, opened 2022-09-13
[QC-854] Add templating to the discard log file

Barthelemy:


Framework/include/QualityControl/runnerUtils.h
```diff
@@ -184,6 +186,24 @@ inline void overrideValues(boost::property_tree::ptree& tree, std::vector<std::p
   }
 }
 
+/**
+ * template the param infologgerDiscardFile (_ID_->[device-id])
+ * @param originalFile
+ * @param iCtx
+ * @return
+ */
+using namespace std;
```
knopers8, line 195: please do not use it in headers, it is going to leak out to all code in `o2::quality_control::core` which includes this file

knopers8 (approved):

Framework/include/QualityControl/runnerUtils.h
Barthelemy: correct, completely forgot to remove it, good catch

Barthelemy:
failure unrelated.

PR #1453, opened 2022-09-14
[QC-820] Fix Service Discovery by using blocking sockets

awegrzyn:
No need to use async operation as service discovery runs in dedicated thread.
The other approach would be to run in the main thread and use async operations.


Barthelemy (approved):

awegrzyn:
I don't think so, it should be all good.

PR #1456, opened 2022-09-14
IncreasingEntries: move the beautification code to the beautify method

Barthelemy:


Modules/Common/src/IncreasingEntries.cxx
```diff
@@ -59,6 +59,7 @@ void IncreasingEntries::configure()
 Quality IncreasingEntries::check(std::map<std::string, std::shared_ptr<MonitorObject>>* moMap)
 {
   Quality result = Quality::Good;
+  mFaultyObjectsNames.clear();
```
knopers8, line 62: If I am not wrong, if you use the "oneachseparately" update policy, you will invoke `check()` a few times, each time clearing this vector. Thus once beautifying, only the last object checked will potentially be in this vector.

knopers8 (changes requested):

Modules/Common/src/IncreasingEntries.cxx
Barthelemy: But when I looked at the code it seemed that for each `check` there would be zero or one `beautify`: 
```
  for (auto& moMapToCheck : moMapsToCheck) {
    std::vector<std::string> monitorObjectsNames;
    boost::copy(moMapToCheck | boost::adaptors::map_keys, std::back_inserter(monitorObjectsNames));

    Quality quality;
    try {
      quality = mCheckInterface->check(&moMapToCheck);
    } catch (...) {
      std::string diagnostic = boost::current_exception_diagnostic_information();
      ILOG(Error, Ops) << "Unexpected exception in user code (check):\n"
                       << diagnostic << ENDM;
      continue;
    }

    ILOG(Info, Support) << "Check '" << mCheckConfig.name << "', quality '" << quality << "'" << ENDM;
    // todo: take metadata from somewhere
    qualityObjects.emplace_back(std::make_shared<QualityObject>(
      quality,
      mCheckConfig.name,
      mCheckConfig.detectorName,
      UpdatePolicyTypeUtils::ToString(mCheckConfig.policyType),
      stringifyInput(mCheckConfig.inputSpecs),
      monitorObjectsNames));
    beautify(moMapToCheck, quality);
  }
```
Or am I misreading this ?

knopers8: yes, you are right, i remembered it differently, sorry.

knopers8: can you then add a comment there that "we promise Checks that beautify() is invoked after check() of the same object"? just in case we refactor this and forget about it.

Barthelemy: yes, I have added more doc and also tried to make it clearer.

knopers8 (approved):

PR #1461, opened 2022-09-15
Script to delete objects versions for a list of runs

Barthelemy:


Framework/CMakeLists.txt
```diff
@@ -424,7 +424,8 @@ install(FILES example-default.json
         DESTINATION etc)
 
 install(PROGRAMS script/RepoCleaner/o2-qc-repo-cleaner
-                 script/RepoCleaner/o2-qc-repo-deletion
+                 script/RepoCleaner/o2-qc-repo-delete-runs
+                 script/RepoCleaner/o2-qc-repo-delete-period
```
knopers8, line 428: I was thinking that you mean data taking period (as the metadata `PeriodName`). Could you please clarify this in the executable name? E.g.  `o2-qc-repo-delete-interval`, o2-qc-repo-delete-time-period`?

Framework/script/RepoCleaner/o2-qc-repo-delete-runs
```diff
@@ -0,0 +1,126 @@
...
+                    logging.debug("")
+                elif answer.lower() in ["n", "no"]:
+                    logging.info("   skipping")
+                else:
+                    logging.error("   wrong input, skipping")
+            else:
+                ccdb.deleteVersion(v)
+                logging.debug("")
+
+        logging.info(f"Deleted items: {len(versions)}")
```
knopers8, line 91: the count will also include the versions where you said "no". That could be misleading and scare us in the future.

knopers8 (approved):
thanks, it's good like this, but please see my suggestions for now or for the next release.

Barthelemy:
I agree on both comments and will modify the script tomorrow

PR #1465, opened 2022-09-19
PrimVtx and ITSTPC matching json to run in multinode; includes also json to be added to consul, to run at P2

chiarazampolli:
@knopers8 : can you check? This is what we discussed. 
Local tests seem to indicate a race condition: the remote node cannot receive the data, it only sees End Of Stream. But since there is no other way to test it, maybe it is ok to commit it?
I include here also [GLO_mnl_EPN.json](https://github.com/AliceO2Group/QualityControl/compare/master...chiarazampolli:json_P2?expand=1#diff-a49b23af8eea2badf5f942a67b0db906bd92c59102a4d369ac99f1b2ef78535e), which is what I will put in consul, if you agree. 

To test:

PrimVertex:
local node:
```
o2-primary-vertex-reader-workflow --hbfutils-config o2_tfidinfo.root -b | o2-qc --config json:///home/zampolli/SOFT/alibuild/ali-o2-dev1/QualityControl/Modules/GLO/vertexing-qc_mnl.json -b --local --host localhost
```

remote node, from another terminal:
```
o2-qc --config json:///home/zampolli/SOFT/alibuild/ali-o2-dev1/QualityControl/Modules/GLO/vertexing-qc_mnl.json -b --remote
```

Matching:
local node:
```
o2-global-track-cluster-reader --track-types "ITS-TPC,ITS,TPC" --cluster-types none -b --hbfutils-config o2_tfidinfo.root | o2-qc --config json:////home/zampolli/SOFT/alibuild/ali-o2-dev1/QualityControl/Modules/GLO/ITSTPCmatchedTracks_mnl.json -b --local --host localhost
```

remote node, from another terminal:
o2-qc --config json:////home/zampolli/SOFT/alibuild/ali-o2-dev1/QualityControl/Modules/GLO/ITSTPCmatchedTracks_mnl.json -b --remote
```

(of course, you need the prim vtx, and tracks files).

### CODE REVIEW

File: Modules/GLO/GLO_mnl_EPN.json [hunk 1]
```diff
@@ -0,0 +1,118 @@
...
+	"taskParameters" : {
+          "isMC": "false"
+        },
+        "location" : "local",
+	"localMachines": [
+            "epn",
+	    "localhost"
+        ],
+        "remoteMachine": "alio2-cr1-qc03.cern.ch",
+	"remotePort": "46000",
```
knopers8, line 51: please use a port in range of 47700-47799 which is not yet used by a different task. You can grep ControlWorkflows/scripts/etc for those.

This also applies to the same fields in other tasks/config files.

Modules/GLO/GLO_mnl_EPN.json
```diff
@@ -0,0 +1,118 @@
...
+        },
+	"taskParameters" : {
+          "isMC": "false"
+        },
+        "location" : "local",
+	"localMachines": [
+            "epn",
+	    "localhost"
+        ],
+        "remoteMachine": "alio2-cr1-qc03.cern.ch",
```
knopers8, line 50: I will assign you `alio2-cr1-qc07.cern.ch`.

Modules/GLO/GLO_mnl_EPN.json
```diff
@@ -0,0 +1,118 @@
...
+      "taskParameters" : {
+        "GID" : "ITS-TPC,ITS",
+        "verbose" : "false",
+        "minPtCut" : "0.1f",
+        "etaCut" : "1.4f",
+        "minNTPCClustersCut" : "60",
+        "minDCACut" : "100.f",
+        "minDCACutY" : "10.f",
+        "grpFileName" : "o2sim_grp.root",
+        "geomFileName" : "o2sim_geometry-aligned.root"
```
knopers8, line 75: will these files be accessible on QC nodes?

Modules/GLO/ITSTPCmatchedTracks_mnl.json
```diff
@@ -0,0 +1,82 @@
...
+      },
+      "conditionDB" : {
+        "url" : "ccdb-test.cern.ch:8080"
+      },
+      "infologger" : { "" : "Configuration of the Infologger (optional).",
+                       "filterDiscardDebug" : "false",
+                       "" : "Set to true to discard debug and trace messages (default: false)",
+                       "filterDiscardLevel" : "21",
+                       "" : "Message at this level or above are discarded (default: 21 - Trace)" }
+    },
```
knopers8, line 29: if you want to use also this fie at P2, please use the same qc.config section as in `GLO_mnl_EPN.json`

Modules/GLO/GLO_mnl_EPN.json
```diff
@@ -0,0 +1,118 @@
...
+        "url" : "influxdb-unix:///tmp/telegraf.sock"
+      },
+      "consul" : {
+        "url" : "http://ali-consul.cern.ch:8500"
+      },
+      "conditionDB" : {
+        "url" : "http://localhost:8084"
+      },
+      "infologger" : { "" : "Configuration of the Infologger (optional).",
+                       "filterDiscardDebug" : "1",
```
knopers8, line 25: do you have a lot of debug logs? if not, you could keep it at `false`.

knopers8 (changes requested):

Modules/GLO/GLO_mnl_EPN.json
chiarazampolli: Done.

Modules/GLO/GLO_mnl_EPN.json
chiarazampolli: Thanks.

Modules/GLO/ITSTPCmatchedTracks_mnl.json
chiarazampolli: No, this and `vertexing-qc_mnl.json` are for local testing only.

Modules/GLO/GLO_mnl_EPN.json
chiarazampolli: Yes, it would print too much if I used "false".

Modules/GLO/GLO_mnl_EPN.json
chiarazampolli: I don't know, but we need those. Also TOF should use them, and maybe someone else. How is it done?
This is related also to https://alice.its.cern.ch/jira/browse/QC-810

chiarazampolli: BTW, these files are needed for processing, no merging, so it does not matter if they are not on the merger node.

knopers8: indeed, correct

knopers8 (approved):

PR #1468, opened 2022-09-20
[EMCAL-524] added 2D histo for Ampl and Time CAL per SM

cterrevo:


mfasDa (approved):
Looks good!

Barthelemy:
@TimoWilken is there a problem with the checkers ? there are two that seem stuck.

Merging for the release as the rest seems ok

TimoWilken:
You're right, sorry; fixed.

PR #1470, opened 2022-09-22
[QC-861] Add script move objects

Barthelemy:


Framework/script/RepoCleaner/o2-qc-repo-move-objects
```diff
@@ -0,0 +1,99 @@
...
+def parseArgs():
+    """Parse the arguments passed to the script."""
+    logging.info("Parsing arguments")
+    parser = argparse.ArgumentParser(description='Move all objects in path to new-path')
+    parser.add_argument('--url', dest='url', action='store', help='URL of the CCDB, with http[s]://', required=True)
+    parser.add_argument('--log-level', dest='log_level', action='store', default="20",
+                        help='Log level (CRITICAL->50, ERROR->40, WARNING->30, INFO->20,DEBUG->10)')
+    parser.add_argument('--dry-run', action='store_true',
+                        help='Dry run, no actual deletion nor modification to the CCDB.')
+    parser.add_argument('--path', dest='path', action='store', default="",
```
knopers8, line 19: can you add an example of such a path? So that we do not add `/.*` because it is added automatically.

Framework/script/RepoCleaner/o2-qc-repo-move-objects
```diff
@@ -0,0 +1,99 @@
...
+        exit(0)
+
+    logging.warning("****** ARE YOU ABSOLUTELY SURE YOU WANT TO CONTINUE ? ******")
+    answer = input("Yes/No \n  ")
+    if answer.lower() not in ["y", "yes"]:
+        exit(0)
+
+    for v in versions:
+        logging.info(f"Ready to move {v}")
+        # here we need to make sure that we preserve the subdirectory structure and replace the only the matching part
```
knopers8, line 67:
```suggestion
        # here we need to make sure that we preserve the subdirectory structure and replace only the matching part
```

knopers8 (commented):
thanks, all good, just small doc improvement suggestions

Framework/script/RepoCleaner/o2-qc-repo-move-objects
Barthelemy: agreed and done

knopers8 (approved):

PR #1475, opened 2022-09-24
[EMCAL-524] Fix swapping of cell status

mfasDa:
Wrong comparator used, therefore good cells consisted of
bad cells and vice versa.


Barthelemy:
@TimoWilken @sawenzel o2sim_hepmc is timing out

PR #1482, opened 2022-10-02
add calibration from ccdb in TOF digits QC (applyCalib=true)

noferini:
@njacazio @ercolessi


njacazio:
Thanks Francesco! The workflow  configuration stays unchanged right?

njacazio (approved):

PR #1484, opened 2022-10-03
[EMCAL-530] Fix uninitialized geometry

mfasDa:
Geometry pointer was not initialized with 0 in the
constructor leading to teh geometry not being properly
initialized on certain systems particularly under linux,
resulting in a crash in the internal clusterizer.


Barthelemy:
@TimoWilken @ktf  There seems to be a problem with the macOS build but also the other builders are stuck since yesterday evening.

mfasDa:
CI errors unrelated

PR #1496, opened 2022-10-06
ITS - Improved lane status flag tracking

nicolovalle:
- FHR: Small correction of decoding error plot. Added dynamic number of bins to the 1D plot
- FEE: Added lane status flag that are never reset: they count the number of orbits missed by a given lane
- FEE: Added option to reset lane status and payload plot independently


nicolovalle:
In the second commit the IB triangles in the plots with geometry has been enlarged.

IsakovAD (approved):
All changes are fine. Thanks

PR #1497, opened 2022-10-06
[WIP] TPC: add independently configurable data source for each PP-QC task

tklemenz:
Added the config option `dataSourceURL` to the PP-QC tasks so that multiple tasks running in one workflow can have individual data sources instead of all being fixed to `conditionDB`.

Caveat: When the task is updated on `newObject`, this does not look at `dataSourceURL`. Here you need to choose either ccdb or qcdb ([see docu](https://github.com/AliceO2Group/QualityControl/blob/master/doc/PostProcessing.md#triggers-configuration)). `ccdb` looks at `qc.config.conditioDB.url` and `qcdb` looks at `qc.config.database.host`.

In addition, there are some minor adjustments.


tklemenz:
The error in `o2-dataflow` is unrelated.

PR #1498, opened 2022-10-06
cmake: remove usage of deprecated BASENAME parameter

aphecetche:


iravasen:
@aphecetche thanks! Yes that's old. Fine with me (ITS) if no objections from @IsakovAD

iravasen (approved):

Barthelemy (approved):

PR #1499, opened 2022-10-07
Trigger class rates and trigger class rates per run

mbombara:
Added trigger class rates (64) + trigger class rates per run (in principle 64 classes are divided into (or assigned to) 16 runs). Data source are counters provided by CTP.


iravasen:
Hi @mbombara, thank you. I guess you need to rebase on the master first. I see several changes which are already merged in the master.

mbombara:
@iravasen rebase is done

Barthelemy (approved):

PR #1502, opened 2022-10-07
MFT: timing info edit

diana0x0f:
- set orbit of the first TF for relative time calculation (before, it was updated with every new TF)
- change binning for digits, clusters and tracks per BC to one bin per BC, instead of the number of triggers per orbit


Barthelemy (approved):

Barthelemy:
I am merging to include it in the extra release of today and because it seems to be innocuous.

PR #1503, opened 2022-10-08
Migrate to use ServiceRegistryRef

ktf:
For now it's just an alias, but this will simplify the multithreading migration.


njacazio (approved):
Thanks Giulio

iravasen (approved):

ktf:
@Barthelemy @knopers8 I would need a new patch release with this in as well to continue merging the rest of my multithreading related work.

Notice how I changed something which was using a nullptr to indicate a missing ServiceRegistry to use `ServiceRegistryRef::active<DataTakingContex>()`. If that's not acceptable, let's discuss some other alternative solution.

Barthelemy:
I will not include it in today's release because it is a large change. I will then patch the release for you.

knopers8 (approved):

PR #1504, opened 2022-10-08
Modifications in Tracklet Task

shsharma1:


Modules/TRD/src/TrackletsTask.cxx
```diff
@@ -294,92 +385,116 @@ void TrackletsTask::drawHashedOnHistsPerLayer(int iLayer)
...
+          ctx.inputs().get<gsl::span<o2::trd::Tracklet64>>("tracklets");
+      auto triggerrecords =
+          ctx.inputs().get<gsl::span<o2::trd::TriggerRecord>>("triggers");
       // std::cout << "Tracklets per time frame: " << tracklets.size();
       mTrackletsPerTimeFrame->Fill(tracklets.size());
       mTrackletsPerTimeFrameCycled->Fill(tracklets.size());
       mTriggersPerTimeFrame->Fill(triggerrecords.size());
-      for (auto& trigger : triggerrecords) {
+      if (!digits.size() == 0) {
+        mEventsPerTimeFrame->Fill(triggerrecords.size());
```
bazinski, line 445: You are only counting calibration triggers, what about calibration triggers (digits size >0) ?
If you want to only count digit events then it should be in the digits task, not the tracklets task.

bazinski: also conventionally you would do != instead of the if statement preceding this line.

Modules/TRD/src/TrackletsTask.cxx
```diff
@@ -100,114 +94,182 @@ void TrackletsTask::drawTrdLayersGrid(TH2F* hist)
...
   getObjectsManager()->startPublishing(mTriggersPerTimeFrame);
+  mEventsPerTimeFrame = new TH1F("eventswithdigitspertimeframe",
+                                 "Number of Events with Digits per Time "
+                                 "Frame;Event with digits in Time Frame;Counts",
+                                 100, 0, 100);
+  getObjectsManager()->startPublishing(mEventsPerTimeFrame);
+  mDigitsPerEvent = new TH1F(
+      "digitsperevent", "Number of Digits per event;Digits in Event;Counts",
+      100, 0, 50000);
+  getObjectsManager()->startPublishing(mDigitsPerEvent);
```
bazinski, line 262: Why is this in the TrackletsTask ? When it is solely dealing with digits?
Also in DigitTask is the same spectrum : https://github.com/AliceO2Group/QualityControl/blob/72c5618f40f3529a21b26b6751a32bc8cf3f1f44/Modules/TRD/src/DigitsTask.cxx#L214

bazinski (changes requested):

bazinski:
@shsharma1 take a look through https://github.com/AliceO2Group/CodingGuidelines, particularly the part about git-clang, that will fix your formatting errors.

Modules/TRD/src/TrackletsTask.cxx
shsharma1: > Why is this in the TrackletsTask ? When it is solely dealing with digits? Also in DigitTask is the same spectrum :
> 
> https://github.com/AliceO2Group/QualityControl/blob/72c5618f40f3529a21b26b6751a32bc8cf3f1f44/Modules/TRD/src/DigitsTask.cxx#L214

Yes I checked the digit task now. This is already in the digit task. Will not include it.

Modules/TRD/src/TrackletsTask.cxx
shsharma1: > You are only counting calibration triggers, what about pure physics triggers (digits size ==0 ) ? If you want to only count digit events then it should be in the digits task, not the tracklets task.

Hello, There is already one histogram in Tracklet task where we have events per time frame (not any condition on digits). But here we need to do events with digits per time frame. That is why I have included check on digits. But now I checked digit task and there I see that digit check is already there. So I only need to fill the histogram in digit task w/o digits check?
image

bazinski: yes exactly. The lines after the continue have a non zero digit vector size.

shsharma1: Ok thanks for your reply

PR #1513, opened 2022-10-17
[EMCAL-792] Fix reset of all errors and histo for unknown type

mfasDa:
- Missing reset fixed for counter of any error type, leading
  to same errors merged multiple times
- Add histograms to monitor unkown (errorous) error codes


mfasDa:
We found that the number of counts in the general error overview was increasing in time even though the problem was fixed while the counter of the dedicated error type stayed constant after the fix (see i.e https://ali-qcg.cern.ch/?page=layoutShow&layoutId=62d54366c907df2f546331ac&RunNumber=527446). We suspect this comes from the missing reset of the general overview histogeam which merges the same error code multiple times.

The additional new histogram is for safety monitoring corrupted messages in case the component would receive such.

Barthelemy (approved):

PR #1519, opened 2022-10-19
More work to merge the multithreading PR in O2

ktf:


ktf:
@Barthelemy @knopers8 can we fast track a release with this in? The behaviour is the same unless there is the file Framework/Features.h, which is not there yet.

PR #1523, opened 2022-10-19
TOF: Fix TOF hits vs crate object

ercolessi:


njacazio (approved):
Thanks Francesca

ercolessi:
Hi @njacazio, a small fix in the histograms of tof hits vs crates, this was tested thanks!

PR #1529, opened 2022-10-21
[WIP] [QC-882] Add option to suffix the log file name

Barthelemy:


Barthelemy:
The suffix is passed by argument because it is not easy for EPN to differentiate the config files per NUMA nodes. Yet, I am not 100% happy about having two different ways of setting and modifying the log file.

Barthelemy:
we will use `--override-values` instead.

PR #1532, opened 2022-10-24
[EMCAL-792] Define error histograms based on settings in O2

mfasDa:
Getters for number of error codes, error code titles and
converters from error code to error type have been added
for all error classes in O2. Using these ones to define error
histograms rather than internal hard-coded settings.

### DIFF
--- a/Modules/EMCAL/src/RawErrorTask.cxx
+++ b/Modules/EMCAL/src/RawErrorTask.cxx
@@ -23,6 +23,9 @@
 #include "EMCALBase/Geometry.h"
 #include "EMCALBase/Mapper.h"
 #include "EMCALReconstruction/AltroDecoder.h"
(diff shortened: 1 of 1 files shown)


mfasDa:
Needs https://github.com/AliceO2Group/AliceO2/pull/10068

mfasDa:
Information defining the histograms now in one place together with the error codes itself. Makes modifications of the error codes easier. Like this we can simply i.e. add error codes or move error codes from one type to another since only one place has to be modified and there will be no dependency conflict - before both O2 and QC needed to be modified which depend on each other.

mfasDa:
Error unrelated

mfasDa:
Re-triggered CI

knopers8 (approved):

PR #1533, opened 2022-10-25
Name Service Discovery health thread

awegrzyn:
I'm not sure this is 100% correct but I'd be in favour that all threads in QC/DPL are named....
FYI @ktf


Barthelemy:
format issue

Barthelemy:
Going through old PRs. Why is it still WIP ?

awegrzyn:
I was expecting @ktf feedback on that. In general `gdb` debugging a lot easier if threads are named.

ktf:
Yes, it should be fine. For the record, the OpenBSD man page explicitly says that the string is copied, and some of the linux versions mention the fact the string is written to /proc/self/task/[tid]/comm, so the lifetime of the string should be guaranteed by the fact that there is a copy happening at some point.

PR #1534, opened 2022-10-26
[QC-873 + QC-874] ITS QC: read geometry file from ccdb, correct reading .json parameters

IsakovAD:
Added optional possibility to read geometry from ccdb

### DIFF
--- a/Modules/ITS/src/ITSClusterTask.cxx
+++ b/Modules/ITS/src/ITSClusterTask.cxx
@@ -97,7 +97,18 @@ void ITSClusterTask::initialize(o2::framework::InitContext& /*ctx*/)
 
   getJsonParameters();
 
-  o2::base::GeometryManager::loadGeometry(mGeomPath.c_str());
+  if (mLocalGeometryFile == 1) {
+    ILOG(Info, Support) << "Getting geometry from local file" << ENDM;
+    o2::base::GeometryManager::loadGeometry(mGeomPath.c_str());
+  } else {
+    ILOG(Info, Support) << "Getting geometry from ccdb - timestamp: " << mGeoTimestamp << ENDM;
+    auto& mgr = o2::ccdb::BasicCCDBManager::instance();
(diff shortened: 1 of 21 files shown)


Barthelemy:
Could you check the O2fst error ? I think that it is due to https://alice.its.cern.ch/jira/browse/QC-874

Modules/ITS/src/ITSClusterTask.cxx
```diff
@@ -97,7 +97,18 @@ void ITSClusterTask::initialize(o2::framework::InitContext& /*ctx*/)
...
+  if (mLocalGeometryFile == 1) {
+    ILOG(Info, Support) << "Getting geometry from local file" << ENDM;
+    o2::base::GeometryManager::loadGeometry(mGeomPath.c_str());
+  } else {
+    ILOG(Info, Support) << "Getting geometry from ccdb - timestamp: " << mGeoTimestamp << ENDM;
+    auto& mgr = o2::ccdb::BasicCCDBManager::instance();
+    mgr.setTimestamp(mGeoTimestamp);
+    mgr.get<TGeoManager>("GLO/Config/GeometryAligned");
+    if (!o2::base::GeometryManager::isGeometryLoaded()) {
+      ILOG(Error, Support) << "Can't retrive geometry from ccdb: " << mgr.getURL() << " timestamp: " << mGeoTimestamp << ENDM;
```
Barthelemy, line 109: can it survive such a condition? what will happen later ? should it be a fatal and it exits ?

IsakovAD:
@Barthelemy, @iravasen 

The new commit includes the following updates:

1) .json parameters now obtained with function from utils/common which should provide default values in case of problem with .json + added default values for all parameters
2) Updated handling of error when there is no geomtry in the CCDB

iravasen (approved):

iravasen:
Thanks a lot @IsakovAD ! Very good. 
@Barthelemy if it looks good also for you, I let you merge the code directly.

PR #1536, opened 2022-10-26
Report mismatching run number when merging Monitor Objects

knopers8:
...but just once for a collection, so we do not spam logs for each of 1000 objects.


knopers8:
@ercolessi

knopers8:
To clarify the purpose of this, it can help us e.g. to detect if we get any spurious objects during physics runs from test setups at P2.

Barthelemy (approved):

PR #1538, opened 2022-10-26
Make sure the ports are unique when running the multinode test

ktf:


Framework/script/o2-qc-multinode-test.sh
```diff
@@ -42,29 +49,20 @@ function delete_data() {
...
-if [ -z "$UNIQUE_PORT_1" ]
-then
-  echo "UNIQUE_PORT_1 must be set when calling o2-qc-multinode-test.sh"
-  exit 1
-fi
+# mkdir in /tmp is guaranteed to be atomic
+mkdir /tmp/${UNIQUE_TEST_NAME} || { echo "Concurrent usage of the same port ${UNIQUE_PORT_1} detected, exiting"; exit 1; }
+pushd /tmp/${UNIQUE_TEST_NAME}
+
+UNIQUE_PORT_2=$((UNIQUE_PORT_1+1))
```
ktf, line 62: To avoid conflicts, the cmake code should be changed to have only odd / even ports.

ktf:
Well, at least we can reproduce the issue.

ktf:
For the record, one thing I noticed on my mac is that the port is always the same...

knopers8:
It is generated at cmake time, not execution time. Do you mean it is the same when running cmake over and over?

ktf:
Can we squash & merge this one and have a new release?

Barthelemy:
@knopers8 I let you review.

knopers8 (approved):

PR #1547, opened 2022-10-31
changes in ranges of histograms

shsharma1:


knopers8 (approved):
For the future, I would advise using clearer commit/PR titles, which at least indicate the concerned detector.

PR #1550, opened 2022-11-01
ITS post-processing: fixed crash for empty CCDB object

IsakovAD:
Small fix to avoid crash of the workflow in case if there is no MO in the CCDB


IsakovAD:
@iravasen 

I have removed ThnSparse lines from the .json, and I confirm that FHR PP works well without it. 

I would keep the line "if (mo == nullptr) continue" to avoid crashes in case of similar issues in future.

iravasen:
@IsakovAD ok with me, thanks a lot! Approving.

iravasen (approved):

PR #1551, opened 2022-11-01
[EMCAL-524] Add counters for good/bad cells

mfasDa:
- Adding counters for good/bad cells for the full detector,
  EMCAL/DCAL and vs. supermodule ID
- Increase number of cell range as many cell counter
  histograms are cut in the distribution


mfasDa:
For trending: Since we have the online bad channel calibration enabled we need to monitor the time behaviour of the masked signals.

Barthelemy (approved):

PR #1554, opened 2022-11-03
[FOCAL] Update pixel mapping for IB and OB configuration

mfasDa:
- Provide second mapping with IB configuration
- Make base class for mapping in order to determine mapping version on-the-fly
- Swap row indices for OB mapping


mfasDa:
2 different options for pixel setups are available for the testbeam Nov. 2022:
- ALPIDE IB chip, 3x6 matrix, 2 layers
- ALPIDE OB chip, 7x6 matrix (4x6 in beam), 2 layers
The testbeam task must support both of them in order to be able to change the setup during the TB campaign. Selection of the pixel setup for the moment is steered via the task parameters.

Barthelemy (approved):

PR #1555, opened 2022-11-03
Avoid copy

ktf:
This triggers some of the copies / deletions which you see.


ktf:
Proper fix to avoid the copy completely coming up in the Framework.

knopers8 (approved):
I don't understand, but I trust you.

As far as I can tell, `DeviceSpec` does not contain any `framework::Task`, how does it help?

ktf:
It contains the AlgorithmSpec, which contains the captured task. I am changing a few things in DPL to prevent any copy of those.

knopers8:
OK, indeed, thanks

ktf:
Sorry guys, I need a patch also with this...

PR #1556, opened 2022-11-03
TOF: PID plots per track source and event time 

ercolessi:
@noferini @njacazio

### DIFF
--- a/Modules/PID/src/TaskFT0TOF.cxx
+++ b/Modules/PID/src/TaskFT0TOF.cxx
@@ -48,9 +48,11 @@ using namespace o2::tof;
 namespace o2::quality_control_modules::pid
 {
 
+float MyTrack::t0maxp = 1.5; // default p threshold for tracks for event time computation
+
 bool MyFilter(const MyTrack& tr)
 {
-  return (tr.getP() < tr.getT0MaxP());
+  return (tr.getP() < MyTrack::getT0MaxP());
 }
 
 TaskFT0TOF::~TaskFT0TOF()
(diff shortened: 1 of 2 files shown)


Modules/PID/src/TaskFT0TOF.cxx
```diff
@@ -50,23 +50,27 @@ namespace o2::quality_control_modules::pid
 
 bool MyFilter(const MyTrack& tr)
 {
-  return (tr.getP() < 2.0);
+  return (tr.getP() < 1.5);
```
njacazio, line 53: Maybe we want this configurable without having to change the code?

njacazio (commented):
Ciao Francesca, huge work, thanks!

Modules/PID/src/TaskFT0TOF.cxx
ercolessi: Ciao @njacazio indeed thanks for spotting this! I changed it to configurable

njacazio (approved):

Modules/PID/src/TaskFT0TOF.cxx
noferini: I guess that
float t0maxp = 1.5;
should be 
static float t0maxp;
(the same for the getter)

ercolessi: Ciao @noferini thanks a lot for spotting this I changed it to static!

njacazio (approved):

njacazio (approved):

njacazio (approved):

Barthelemy (approved):

PR #1557, opened 2022-11-03
[FOCAL] Additional customization

mfasDa:
- Add debug mode in order to decrease verbosity in production
- Make PAD window duration configurable

### DIFF
--- a/Modules/FOCAL/src/TestbeamRawTask.cxx
+++ b/Modules/FOCAL/src/TestbeamRawTask.cxx
@@ -80,7 +80,20 @@ TestbeamRawTask::~TestbeamRawTask()
 
 void TestbeamRawTask::initialize(o2::framework::InitContext& /*ctx*/)
 {
-  mPadDecoder.setTriggerWinDur(20); // MF @TODO make taskParameter
+  auto get_bool = [](const std::string_view input) -> bool {
+    return input == "true";
+  };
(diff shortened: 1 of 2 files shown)


Modules/FOCAL/src/TestbeamRawTask.cxx
```diff
@@ -181,7 +194,9 @@ void TestbeamRawTask::monitorData(o2::framework::ProcessingContext& ctx)
       int currentpos = 0;
       while (currentpos < databuffer.size()) {
         auto rdh = reinterpret_cast<const o2::header::RDHAny*>(databuffer.data() + currentpos);
-        o2::raw::RDHUtils::printRDH(rdh);
+        if (mDebugMode) {
```
Barthelemy, line 197: I want to make you aware that the debug mode will probably flood the logs and create a huge back pressure.

Barthelemy (approved):

Modules/FOCAL/src/TestbeamRawTask.cxx
mfasDa: That's why I added the option as "debug mode" - it is only for local debugging and switched off in runs. We will still need it for a while in tests to prepare i.e. different setups (i.e. different pixel layout), but we will remove the debug once the setup is stable. Keep in mind that the detector is still in the development phase.

PR #1567, opened 2022-11-10
Adding Reductor of Cluster CalPad Objects for TPC QC

mlesch:


Modules/TPC/src/CalPadClusterReductor.cxx
```diff
@@ -0,0 +1,110 @@
...
+
+          // Remove pads which are empty from any calculations
+          data.erase(std::remove_if(data.begin(), data.end(), [](const auto& value) { return (std::isnan(value) || value <= 0); }), data.end());
+
+          Float_t(*Quantity)[72] = getArrayPointer(iType);
+          Quantity[0][iROC] = static_cast<float>(data.size());
+          Quantity[1][iROC] = TMath::Mean(data.begin(), data.end());
+          Quantity[2][iROC] = TMath::StdDev(data.begin(), data.end());
+          Quantity[3][iROC] = TMath::Median(data.size(), data.data());
+          Quantity[4][iROC] = TMath::RMS(data.size(), data.data());
```
wiechula, line 62: I think `TMath::RMS` and TMath::Stddev` are identical https://root.cern.ch/doc/master/TMath_8h_source.html#l00525, so no need to have both. Since `StdDev` is the correct term, I would only keep that.

wiechula (changes requested):

Modules/TPC/src/CalPadClusterReductor.cxx
mlesch: Done

knopers8 (approved):

PR #1571, opened 2022-11-18
TPC: Fix for IDC displaying in the QCG

tklemenz:
This fixes the problem that only the very first IDC QC objects in a run were properly displayed in the QCG.

### DIFF
--- a/Modules/TPC/run/tpcQCIDCs.json
+++ b/Modules/TPC/run/tpcQCIDCs.json
@@ -17,7 +17,7 @@
         "url": ""
       },
       "conditionDB": {
-        "url": "ccdb-test.cern.ch:8080"
+        "url": "http://alice-ccdb.cern.ch"
       }
     },
     "postprocessing": {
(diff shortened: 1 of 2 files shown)


tklemenz:
The error is unrelated. Can this be merged?

wiechula (approved):

wiechula:
Fine from my side. But I cannot force merge. @Barthelemy , can you have a look?

Barthelemy:
@ktf is trying to fix the macos error. We will force merge tomorrow or Thursday when preparing the next release if it has not been fixed by this time. Is that ok for you ?

tklemenz:
Yes, I guess it will anyway make it to the next release, right?

Barthelemy:
yes it will

tklemenz:
Very good. Then it is absolutely no problem.

ktf:
For the record, the alleged fix for the macOS issue was merged. Let me know if that is the case.

Barthelemy:
how do we restart the failed check ?

PR #1575, opened 2022-11-20
[WIP] Track matching and PH per chamber

vikasLstar:
Tracks in TRD are matched with TPC and ITS. PH spectra were obtained for 540 chambers. Parsing time for each timeframe is also included.


Barthelemy:
Add 'WIP' in the title will block the merge.

Barthelemy:
is it still work in progress ? otherwise I would propose to close the PR in order not to clutter the list of opened PRs.

PR #1576, opened 2022-11-21
L1phase QC

peressounko:


peressounko:
Dear @Barthelemy, something strange with macOS  test: first time it failed in hepmc test, then I pushed dummy commit to restart test and it again failed in macOS check but in installation. Is it known feature?

Barthelemy:
yes, it is.

Barthelemy:
the problem is being fixed by Giulio but I will force merge because I need to release.

Barthelemy (approved):

PR #1579, opened 2022-11-25
Matching of TRD tracks with ITS and TPC

vikasLstar:
ADC signals were obtained for the digits containing TRD tracks matched with TPC and ITS.

### DIFF
--- a/Modules/TRD/src/PulseHeightTrackMatch.cxx
+++ b/Modules/TRD/src/PulseHeightTrackMatch.cxx
@@ -11,7 +11,7 @@
 
 ///
 /// \file   PulseHeightTrackMatch.cxx
-/// \author My Name
+/// \author Vikash Sumberia
 ///
 
 #include <TCanvas.h>
@@ -59,23 +59,26 @@ void PulseHeightTrackMatch::retrieveCCDBSettings()
 
 void PulseHeightTrackMatch::buildHistograms()
 {
-   mParsingTimePerTF.reset(new TH1F("parsingtimeperTF", "Time for processing each TF", 100000, 0, 100000));
-   mParsingTimePerTF->GetXaxis()->SetTitle("Time in #mus");
-   mParsingTimePerTF->GetYaxis()->SetTitle("Counts");
-   getObjectsManager()->startPublishing(mParsingTimePerTF.get());
-  
-   mDigitsPerEvent.reset(new TH1F("digitsperevent", "Digits per Event", 10000, 0, 10000));
-   getObjectsManager()->startPublishing(mDigitsPerEvent.get());
-  
-   mTracksPerEvent.reset(new TH1F("tracksperevent", "Matched TRD tracks per Event", 100, 0, 10));
-   getObjectsManager()->startPublishing(mTracksPerEvent.get());
+  mParsingTimePerTF.reset(new TH1F("parsingtimeperTF", "Time for processing each TF", 100000, 0, 100000));
+  mParsingTimePerTF->GetXaxis()->SetTitle("Time in #mus");
+  mParsingTimePerTF->GetYaxis()->SetTitle("Counts");
+  getObjectsManager()->startPublishing(mParsingTimePerTF.get());
 
-   mTriggerPerTF.reset(new TH1F("triggerpertimeframe", "Triggers per TF", 1000, 0, 1000));
-   getObjectsManager()->startPublishing(mTriggerPerTF.get());
+  mDigitsPerEvent.reset(new TH1F("digitsperevent", "Digits per Event", 10000, 0, 10000));
+  getObjectsManager()->startPublishing(mDigitsPerEvent.get());
 
-   mTriggerWDigitPerTF.reset(new TH1F("triggerwdpertimeframe", "Triggers with Digits per TF", 10, 0, 10));
-   getObjectsManager()->startPublishing(mTriggerWDigitPerTF.get());
-   
+  mTracksPerEvent.reset(new TH1F("tracksperevent", "Matched TRD tracks per Event", 100, 0, 10));
+  getObjectsManager()->startPublishing(mTracksPerEvent.get());
+
+  mTrackletsPerMatchedTrack.reset(new TH1F("trackletspermatchedtrack", "Tracklets per matched TRD track", 100, 0, 10));
+  getObjectsManager()->startPublishing(mTrackletsPerMatchedTrack.get());
+
+  mTriggerPerTF.reset(new TH1F("triggerpertimeframe", "Triggers per TF", 1000, 0, 1000));
+  getObjectsManager()->startPublishing(mTriggerPerTF.get());
+
+  mTriggerWDigitPerTF.reset(new TH1F("triggerwdpertimeframe", "Triggers with Digits per TF", 10, 0, 10));
+  getObjectsManager()->startPublishing(mTriggerWDigitPerTF.get());
+
   mPulseHeightpro.reset(new TProfile("PulseHeight/mPulseHeightpro", "Pulse height profile  plot;Timebins;Counts", 30, -0.5, 29.5));
   mPulseHeightpro.get()->Sumw2();
   getObjectsManager()->startPublishing(mPulseHeightpro.get());
@@ -84,7 +87,6 @@ void PulseHeightTrackMatch::buildHistograms()
   mPulseHeightperchamber.get()->Sumw2();
   getObjectsManager()->startPublishing(mPulseHeightperchamber.get());
   getObjectsManager()->setDefaultDrawOptions(mPulseHeightperchamber.get()->GetName(), "colz");
- 
 }
 
 void PulseHeightTrackMatch::initialize(o2::framework::InitContext& /*ctx*/)
@@ -93,7 +95,6 @@ void PulseHeightTrackMatch::initialize(o2::framework::InitContext& /*ctx*/)
 
   retrieveCCDBSettings();
   buildHistograms();
-  
 }
 
 void PulseHeightTrackMatch::startOfActivity(Activity& /*activity*/)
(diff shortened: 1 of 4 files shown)


Barthelemy (approved):

Barthelemy:
hello, 
Could you fixx the format please ? 
Thank you

vikasLstar:
Hi, 
Formatting has been done. Also, some more lines have been added to get the number of tracklets from the matched tracks.
Thanks and regards,
Vikash

knopers8:
a legitimate problem from macOS:
```
/System/Volumes/Data/build/alice-ci-workdir/qualitycontrol-o2/sw/SOURCES/QualityControl/1579/0/Modules/TRD/src/PulseHeightTrackMatch.cxx:188:85: error: format specifies type 'int' but the argument has type 'o2::dataformats::GlobalTrackID' [-Werror,-Wformat]
          printf("Got track with %i tracklets and ID %i \n", track.getNtracklets(), track.getRefGlobalTrackId());
```

vikasLstar:
Problem with macOS should be fixed now.

Many thanks.

knopers8:
It seems there is still something:
```

Undefined symbols for architecture x86_64:
  "o2::trd::Tracklet64::getPadCol() const", referenced from:
--
```

vikasLstar:
probably should be fixed now.
Many thanks.

vikasLstar:
could you please what is the error now?
Thanks.

knopers8:
it was a temporary, unrelated issue, looks good now, thanks!

Modules/TRD/src/PulseHeightTrackMatch.cxx
```diff
@@ -0,0 +1,308 @@
...
+  if (a->getPadRow() > b->getPadRow()) {
+    return 0;
+  }
+  if (a->getPadCol() < b->getPadCol()) {
+    return 1;
+  }
+  return 0;
+}
+
+void PulseHeightTrackMatch::monitorData(o2::framework::ProcessingContext& ctx)
```
tdietel, line 134: Please comment the code, this makes it much more readable.

Modules/TRD/src/PulseHeightTrackMatch.cxx
```diff
@@ -0,0 +1,308 @@
...
+{
+  for (auto&& input : ctx.inputs()) {
+    auto digits = ctx.inputs().get<gsl::span<o2::trd::Digit>>("digits");
+    auto tracklets = ctx.inputs().get<gsl::span<o2::trd::Tracklet64>>("tracklets");
+    auto triggerrecords = ctx.inputs().get<gsl::span<o2::trd::TriggerRecord>>("triggers");
+    auto tracks = ctx.inputs().get<gsl::span<o2::trd::TrackTRD>>("tracks");
+    auto triggerrecordstracks = ctx.inputs().get<gsl::span<o2::trd::TrackTriggerRecord>>("trigrectrk");
+
+    std::vector<o2::trd::Digit> digitv(digits.begin(), digits.end());
```
tdietel, line 143: I do not see the need for `digitv`: it seems to be instantiated as a copy of the digits for one event, then the size is checked, gets sorted, but I do not see where it is used after that.

Modules/TRD/src/PulseHeightTrackMatch.cxx
```diff
@@ -0,0 +1,308 @@
...
+
+      if (trigger.getNumberOfDigits() == 0) {
+        continue;
+      }
+
+      else {
+        nTrgWDigits++;
+      }
+
+      if (trigger.getNumberOfDigits() > 10000) {
```
tdietel, line 167: TH1 has an overflow bin, so you do not need this if statement.

Modules/TRD/src/PulseHeightTrackMatch.cxx
```diff
@@ -0,0 +1,308 @@
...
+        mDigitsPerEvent->Fill(9999);
+      } else {
+        mDigitsPerEvent->Fill(trigger.getNumberOfDigits());
+      }
+
+      // now sort digits to det,row,pad
+      std::sort(std::begin(digitsIndex) + trigger.getFirstDigit(), std::begin(digitsIndex) + trigger.getFirstDigit() + trigger.getNumberOfDigits(),
```
tdietel, line 174: @bazinski or @martenole should comment here.
I understand the digits should be sorted in the message, so this sorting should not be necessary. And if they are not sorted, it should be safe to sort the digits, instead of introducing one level in indirection with the digitsIndex.

Modules/TRD/src/PulseHeightTrackMatch.cxx
```diff
@@ -0,0 +1,308 @@
...
+
+    int nTrgWDigits = 0;
+    int nTotalTrigger = 0;
+    auto crustart = std::chrono::high_resolution_clock::now();
+
+    for (const auto& trigger : triggerrecords) {
+      // printf("-------------Trigger----------------------BCData for Trigger %i, %i \n ",trigger.getBCData().orbit,trigger.getBCData().bc);
+      nTotalTrigger++;
+
+      if (trigger.getNumberOfDigits() == 0) {
```
tdietel, line 159: This if statement would explain why you never see 0 triggers with digits in a timeframe: you already reject timeframes without digits here, before you fill the histogram in line 267

Modules/TRD/src/PulseHeightTrackMatch.cxx
```diff
@@ -0,0 +1,308 @@
...
+      for (const auto& triggerTracks : triggerrecordstracks) {
+
+        if (trigger.getBCData() != triggerTracks.getBCData()) {
+          continue;
+        }
+
+        int nTracks = 0;
+        for (int iTrack = triggerTracks.getFirstTrack(); iTrack < triggerTracks.getFirstTrack() + triggerTracks.getNumberOfTracks(); iTrack++) {
+          nTracks++;
+          const auto& track = tracks[iTrack];
```
tdietel, line 186:
```suggestion
        for (const auto& track : triggerTracks.getTrackRefs()) {
          nTracks++;
```
Please test, but it should work and make the code more readable.

Modules/TRD/src/PulseHeightTrackMatch.cxx
```diff
@@ -0,0 +1,308 @@
...
+
+          // printf("Got track with %i tracklets \n", track.getNtracklets());
+          int ntracklets = track.getNtracklets();
+          if (ntracklets > 10)
```
tdietel, line 190: As mentioned above, TH1 has an overflow bin.

Modules/TRD/src/PulseHeightTrackMatch.cxx
```diff
@@ -0,0 +1,308 @@
...
+            int det = trklt.getDetector();
+            int sec = det / 30;
+            int row = trklt.getPadRow();
+            // int col = trklt.getPadCol();
+
+            // obtain pad number relative to MCM center
+            int padLocal = trklt.getPositionBinSigned() * o2::trd::constants::GRANULARITYTRKLPOS;
+            // MCM number in column direction (0..7)
+            int mcmCol = (trklt.getMCM() % o2::trd::constants::NMCMROBINCOL) + o2::trd::constants::NMCMROBINCOL * (trklt.getROB() % 2);
+            int col = 6.f + mcmCol * ((float)o2::trd::constants::NCOLMCM) + padLocal;
```
tdietel, line 213: I remember a discussion with Felix that these calculations should be provided centrally, either in the tracklet class or as a helper function. Can you check with him?

Modules/TRD/src/PulseHeightTrackMatch.cxx
tdietel: And I do not see that digitsIndex is used anywhere, so it should be removed before we merge this PR.

Modules/TRD/src/PulseHeightTrackMatch.cxx
```diff
@@ -0,0 +1,308 @@
...
+
+            for (int iDigit = trigger.getFirstDigit() + 1; iDigit < trigger.getFirstDigit() + trigger.getNumberOfDigits() - 1; ++iDigit)
+
+            {
+              const auto& digit = digits[iDigit];
+              const auto& digitBefore = digits[iDigit - 1];
+              const auto& digitAfter = digits[iDigit + 1];
```
tdietel, line 220: Here you assume that the digits are sorted, but you only sorted the digitsIndex above. Please clarify with @bazinski if the digits are already sorted, or if they are safe to be sorted.

tdietel: It seems you also do not consider shared pads that appear twice in the data stream.

Modules/TRD/src/PulseHeightTrackMatch.cxx
```diff
@@ -0,0 +1,308 @@
...
+
+              std::tuple<unsigned int, unsigned int, unsigned int> aa, ba, ca;
+              aa = std::make_tuple(digitBefore.getDetector(), digitBefore.getPadRow(), digitBefore.getPadCol());
+              ba = std::make_tuple(digit.getDetector(), digit.getPadRow(), digit.getPadCol());
+              ca = std::make_tuple(digitAfter.getDetector(), digitAfter.getPadRow(), digitAfter.getPadCol());
```
tdietel, line 225: I do not see the need for intermediate tuples. The values can be calculated when needed.
It's also not necessary to process the adjacent pads at this point, because for now you are only looking at the coordinates of the central pad. Once you found a match for that, you can check the neighbors.

Modules/TRD/src/PulseHeightTrackMatch.cxx
```diff
@@ -0,0 +1,308 @@
...
+              bool consecutive = false;
+              bool digitmatch = false;
+
+              if (det == det2 && row == row2) {
+
+                int coldif = col2 - col;
+                if (TMath::Abs(coldif) < 4)
+                  digitmatch = true;
+              }
+              if (det1 == det2 && det2 == det3 && row1 == row2 && row2 == row3 && col2 + 1 == col1 && col3 + 1 == col2)
```
tdietel, line 240: There is some redundancy here, because the digits are sorted. For example, det1==det3 automatically implies det1==det2 and det2==det3.

tdietel (changes requested):
I left some comments inline. 
A general comment is that the matching of digits to tracklets does not seem to be done in an optimal way. Both digits and tracklets have the information from which MCM they originate, and this can help to disambiguate which digits were used for the calculation of a tracklet. Otherwise you could end up with using the shared digits from a neighboring MCM.

I think It can happen that the tracklet position is calculated to be on one of the outermost digits, because the tracklets have a slope and the position is given at the edge of a chamber.  Please check the channel number of the first and the last digit from the same MCM that sent the tracklet. There should be a very strong correlation with the tracklet position, but it woud be good to see how well this works.

PR #1581, opened 2022-11-25
[EMCAL-524, O2-3383] Force trigger BC of LED triggers to be 0

mfasDa:
LED triggers must be forced to BC phase 0 since
histograms are only allocated for that phase.


mfasDa:
Fixes crash related to cell task in async reconstruction of PbPb reported in https://alice.its.cern.ch/jira/browse/O2-3383

Barthelemy (approved):

PR #1583, opened 2022-11-29
[EMCAL-530] Distinguish phys and calib trigger in cluster task

mfasDa:
- Add histograms for calib trigger
- Add optional control histograms at cell level
- Outsource histogram filling for phys. trigger
  to dedicated function


mfasDa:
CI error unrelated

Barthelemy (approved):

PR #1585, opened 2022-12-01
[QC-469] Refactor repocleaner to have a proper python module

Barthelemy:


Framework/script/RepoCleaner/qcrepocleaner/Ccdb.py
```diff
@@ -57,15 +57,15 @@ def __init__(self, url):
...
+    def getObjectsList(self, added_since: int = 0, path: str = "") -> List[str]:
         '''
         Get the full list of objects in the CCDB that have been created since added_since.
 
         :param added_since: if specified, only return objects added since this timestamp in epoch milliseconds.
         :return A list of strings, each containing a path to an object in the CCDB.
         '''
         logger.debug(f"added_since : {added_since}")
-        url_for_all_obj = self.url + '/latest/.*'
+        url_for_all_obj = self.url + '/latest/' + path + '/.*'
```
knopers8, line 68: if the path is empty, we get double "/", is this fine?

Framework/script/RepoCleaner/qcrepocleaner/config.yaml
```diff
@@ -25,9 +25,9 @@ Rules:
 #  - object_path: Test
 #    delay: 240
 #    policy: none_kept
-#  - object_path: .*
-#    delay: 1440
-#    policy: skip
+  - object_path: qc/DAQ/MO/daqTask/numberRDHs
+    delay: 1440
+    policy: none_kept
```
knopers8, line 30: did you enable this only for a test or this is actually needed?

knopers8 (commented):
I do not know much about proper python packaging, but it looks sane.

Framework/script/RepoCleaner/qcrepocleaner/Ccdb.py
Barthelemy: yes, I tested this case

Framework/script/RepoCleaner/qcrepocleaner/config.yaml
Barthelemy: good catch, that was just for a test, I will remove it

knopers8 (approved):

PR #1588, opened 2022-12-08
 ITS Online: Updated TGeometry loader

IsakovAD:
Fixed retrieving of TGeometry file. 

Now URL for ccdb with configs is taken from the "conditionDB" JSON parameter


iravasen:
Hi @IsakovAD thanks a lot, I approve. Can you please do the same modification for the cluster json in O2DPG and tell them that it has to go together with this PR?

iravasen (approved):

PR #1589, opened 2022-12-09
add TOF residuals vs pt

noferini:
- Adding TOF residuals also vs pT
- Minor fix in track selection to avoid border effect in filling histos with eta=0.8 and pt=0.3

@njacazio @ercolessi


njacazio (approved):

njacazio:
I think that for the PID part we need the approval of @knopers8 and @Barthelemy

noferini:
All checks passed. It would be nice to have it in as soon as possible (december reco)

knopers8 (approved):

PR #1590, opened 2022-12-12
[FOCAL-8] Use pixel decoding from O2

mfasDa:
Pixel decoding has been ported to O2 in
https://github.com/AliceO2Group/AliceO2/pull/10444.
In favour of a common package the decoding
within QualityControl is dropped.


mfasDa:
Needs https://github.com/AliceO2Group/AliceO2/pull/10469

PR #1591, opened 2022-12-13
[WIP] attempt at fixing the crash with the updated root

Barthelemy:


Barthelemy:
@mlesch We have got an issue with the postprocessing. Details here: https://alice.its.cern.ch/jira/browse/QC-901

Basically, we were using `Draw()` several times on the same object. It adds the objects twice to the pad and when the pad gets deleted it will try to delete all the pointers (hence double deletion). The former ROOT version issued and error but did not carry out with the double deletion. The new ROOT just does it and we crash. 

I don't really know the code in `SliceTrendingTask`. I did what I could to remove all the double `drawing`. One should use `pad->Modified; pad->Update()` instead. Could you check what I did ? 

Thank you

Barthelemy:
@TimoWilken @ktf any idea why it says that there is a merge conflict ? I don't see that on my setup

Framework/src/SliceTrendingTask.cxx
```diff
@@ -187,7 +187,7 @@ void SliceTrendingTask::generatePlots()
           beautifyGraph(multigraph, plot, c);
           multigraph->Draw("A pmc plc");
           c->cd(2);
```
mlesch, line 189: I think we can remove this. We don't need to switch to the second pad anymore as the legend was already drawn.

Framework/src/SliceTrendingTask.cxx
```diff
@@ -187,7 +187,7 @@ void SliceTrendingTask::generatePlots()
           beautifyGraph(multigraph, plot, c);
           multigraph->Draw("A pmc plc");
```
mlesch, line 188: Can be removed, the multigraph was already drawn in `drawCanvasMO`

Framework/src/SliceTrendingTask.cxx
```diff
@@ -198,6 +198,7 @@ void SliceTrendingTask::generatePlots()
...
```
mlesch, line 199: Can be removed, the multigraph was already drawn in `drawCanvasMO`

Framework/src/SliceTrendingTask.cxx
```diff
@@ -215,7 +216,7 @@ void SliceTrendingTask::generatePlots()
           getUserAxisRange(plot.graphYRange, yMin, yMax);
           histo->SetMinimum(yMin);
           histo->SetMaximum(yMax);
-          histo->Draw(plot.option.data()); // redraw and update to force changes on y-axis
+          c->Modified();
           c->Update();
         }
```
mlesch, line 222: The `histo->Draw(plot.option.data());` three lines below can also be removed

mlesch (commented):
Hello @Barthelemy 

Thanks a lot for the changes! Please correct me if I am wrong, but as I understand the situation we should draw an object only once, but if we modify the already drawn object, then we need to call the `pad->Modified; pad->Update()`. If this is the case, then the function `drawCanvasMO` (called in line 172) does all the drawing. Everything after that just modifies the drawn objects, e.g. adding axis-labels. I added a view comments where I think we should remove the `Draw()`. I can unfortunately not directly test the code for now, as I don't have the new ROOT version yet. However, if you want to make a quick test and see if it crashes, the SliceTrendingTask can be tested with QualityControl/Framework/postprocessing.json with the task `ExampleTrendExtended`.

Barthelemy:
Thank you @mlesch, I have applied your proposed changes and tested it. No crashes ! 

I propose to merge it as is.

Barthelemy:
@mlesch could you review again and approve ?

mlesch (approved):
Hello @Barthelemy 

Thanks a lot for the changes! I'm happy to hear that there are no crashes. I looked again at the code and didn't find anything suspicious. 

Cheers,
Marcel

PR #1595, opened 2022-12-15
ITS: Ranges updated to be compatible with back-bias calibration

iravasen:


iravasen:
Hi @IsakovAD , trivial fix for the ranges so to be good also when calibrating with back-bias voltage applied to the staves (cc @nicolovalle )

iravasen:
Hi @Barthelemy, can you please force-merge this PR? The errors are not related to the code modifications here.

PR #1597, opened 2022-12-19
TPC: Add Trending Reductor for IDC0 Scale Factor

mlesch:
Reductor for the scale factor of IDC0 for A side and C side of the TPC. This code needs the changes done in this PR https://github.com/AliceO2Group/AliceO2/pull/10493

### DIFF
--- a/Modules/TPC/src/IDCScaleReductor.cxx
+++ b/Modules/TPC/src/IDCScaleReductor.cxx
@@ -38,7 +38,7 @@ void IDCScaleReductor::update(TObject* obj)
     auto canvas = dynamic_cast<TCanvas*>(obj);
 
     if (canvas) {
-      TGraph* g = (TGraph*)canvas->GetListOfPrimitives()->FindObject("g_IDC0ScaleFactor");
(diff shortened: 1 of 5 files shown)


Modules/TPC/src/IDCScaleReductor.cxx
```diff
@@ -0,0 +1,52 @@
...
+  return "IDCScaleFactorASide/F:IDCScaleFactorCSide";
+} // const char* IDCScaleReductor::getBranchLeafList()
+
+void IDCScaleReductor::update(TObject* obj)
+{
+  if (obj) {
+    auto canvas = dynamic_cast<TCanvas*>(obj);
+
+    if (canvas) {
+      TGraph* g = (TGraph*)canvas->GetListOfPrimitives()->FindObject("g_IDC0ScaleFactor");
```
wiechula, line 41: You could do dynamic cast here as well.

wiechula (changes requested):

wiechula (approved):

PR #1601, opened 2023-01-02
bug fix in tof digit qc

noferini:
@chiarazampolli , @ercolessi , @njacazio 
Fix of a bug causing a crash in async reconstruction (apass2). 
A protection added for rare case, not spotted in apass1 probably because we move to 100% sampling only in apass2.

### DIFF
--- a/Modules/TOF/src/TaskDigits.cxx
+++ b/Modules/TOF/src/TaskDigits.cxx
@@ -364,6 +364,10 @@ void TaskDigits::monitorData(o2::framework::ProcessingContext& ctx)
         bcCorrCable += o2::constants::lhc::LHCMaxBunches;
       }
 
+      if (bcCorrCable >= o2::constants::lhc::LHCMaxBunches) {
+        bcCorrCable -= o2::constants::lhc::LHCMaxBunches;
+      }
+
       ndigitsPerBC[row.mFirstIR.orbit % nOrbits][bcCorrCable]++;
 
       ndigitsPerCrate[crateECH]++;


chiarazampolli:
Thanks! How often did the problem happen? I don't think that this was affecting significantly apass2, was it?

noferini:
> Thanks! How often did the problem happen? I don't think that this was affecting significantly apass2, was it?

Accordingly to Catalin, in LHC22m (80/182 errors analyzed) 62/80 errors were due to TOF digit QC.

chiarazampolli:
Thanks. What is important though for TPC calibration is that we have anyway a high success rate. Was it the case? I cannot check right now. Otherwise we might need to reprocess some runs/masterjobs (if there are any significantly affected).

noferini:
> Thanks. What is important though for TPC calibration is that we have anyway a high success rate. Was it the case? I cannot check right now. Otherwise we might need to reprocess some runs/masterjobs (if there are any significantly affected).

Yes, it is the case > 98% success rate.
Cheers
Francesco

chiarazampolli:
Perfect, i was not sure there was some run especially affected.

thanks!

chiara

knopers8:
@noferini is this blocking us? should we wait for Nicolo for a review or shall i force-merge?

noferini:
Hi @knopers8 ,
the PR is quite simple and I think it can be merged also now (no need of a review). It is not super-urgent but it is crucial for the reconstruction starting next week.
Thanks!

PR #1602, opened 2023-01-03
[QC-911] Propagate provenance to trigger

mfasDa:
Provenance was not propagated to trigger in case
trigger is foreachobject.


mfasDa:
Provenance now properly propagated to trendValues.

knopers8 (approved):

knopers8:
Thanks!

PR #1604, opened 2023-01-05
[QC-912] Fix storing user metadata in Checks

knopers8:


Barthelemy (approved):

Barthelemy:
As there are some positive tests and the other ones fail for unrelated reasons, I merge.

PR #1605, opened 2023-01-05
Fix timing info in ITS cluster and track tasks

iravasen:
To retrieve the cluster dictionary from CCDB, the timestamp is now retrieved by using the `current` timestamp when option `0` for the `dicttimestamp` in the json is used. In this PR, we set the TF timestamp (using `TimingInfo.creation`) instead of the current timestamp. 

Requested by reconstruction experts. 

cc @chiarazampolli @shahor02


shahor02 (approved):

Barthelemy:
@ktf I am getting new errors this morning: 
> /sw/SOURCES/O2/build_QualityControl_o2-dataflow-cs8-slc8_x86-64/0/Framework/PhysicsSupport/src/Plugin.cxx:16:10: fatal error: Framework/O2DatabasePDGPlugin.h: No such file or directory

IsakovAD (approved):
Changes are fine; 

Sorry for the delay, I though that this PR already was reviewed by Ivan:

PR #1606, opened 2023-01-05
[QC-913] Support run number as TrendingType in SliceTrendingTask

mfasDa:
Add tree reader query for run number based on
meta.runNumber and the following two trending types
for trending as function of the run number
- run
- multigraphrun

### DIFF
--- a/Framework/src/SliceTrendingTask.cxx
+++ b/Framework/src/SliceTrendingTask.cxx
@@ -172,7 +172,7 @@ void SliceTrendingTask::generatePlots()
     drawCanvasMO(c, plot.varexp, plot.name, plot.option, plot.graphErrors, mAxisDivision[varName]);
 
     int NumberPlots = 1;
-    if (plot.varexp.find(":time") != std::string::npos) { // we plot vs time, multiple plots on canvas possible
+    if (plot.varexp.find(":time") != std::string::npos || plot.varexp.find(":run") != std::string::npos) { // we plot vs time, multiple plots on canvas possible
       NumberPlots = mNumberPads[varName];
     }
     for (int p = 0; p < NumberPlots; p++) {
@@ -243,9 +243,9 @@ void SliceTrendingTask::drawCanvasMO(TCanvas* thisCanvas, const std::string& var
   getTrendErrors(err, errXName, errYName);
 
   // Divide the canvas into the correct number of pads.
-  if (trendType == "time") {
+  if (trendType == "time" || trendType == "run") {
     thisCanvas->DivideSquare(mNumberPads[varName]); // trending vs time: multiple plots per canvas possible
-  } else if (trendType == "multigraphtime") {
+  } else if (trendType == "multigraphtime" || trendType == "multigraphrun") {
     thisCanvas->Divide(2, 1);
   } else {
     thisCanvas->DivideSquare(1);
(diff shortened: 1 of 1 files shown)


mfasDa:
@mlesch please review!

This is needed in case of trending with respect to the run number and not the timestamp, which is relevant for the async QC.

mfasDa:
Any good reason why the slice reductor should require at least 2 bins in x? I.e. we might have histograms for counting purpose which have exactly 1 bin (event counters). If I want to trend the bin content I would need a regular trending task just for that purpose ...

Modules/Common/src/TH1SliceReductor.cxx
```diff
@@ -141,7 +141,7 @@ void TH1SliceReductor::GetTH1StatsY(TH1* hist, float stats[3],
     ILOG(Error, Support) << "Error: Negative bin in TH1ReducterTPC::GetTH1StatsY" << ENDM;
     exit(0);
   }
-  if (upperBin <= lowerBin) {
```
mlesch, line 144: Please remove this, I will update the TH1SliceReductor in the next couple of days as there are more changes to be made

mlesch (changes requested):
Hello @mfasDa , 

Thanks a lot for the changes to support the trending vs runnumber! I appreciate it a lot. 
To me, the changes in the ```SliceTrendingTask.cxx``` look fine. Did you test it with some test asynch-postprocessing?

About the ```TH1SliceReductor.cxx```: What is passed in the json file are the upper and lower boundary of the slice. Hence, one always needs two entries but it is possible to trend one single bin. However, I see the thing you changed in ```TH1SliceReductor.cxx``` and indeed, at this part there is a bug that prevents the trending of a single bin for statistics along the y axis. As there are more changes I want to do on the ```TH1SliceReductor.cxx```, I would like to ask if you could revert your changes on it. I will then pick it up in the next couple of days and change it myself. Thanks a lot!

mlesch (changes requested):
Sorry for the additional comment, but on a second look I did find something in the `SliceTrendingTask.cxx` that would need to be changed. In line 175 one has to change 

`if (plot.varexp.find(":time") != std::string::npos) { // we plot vs time, multiple plots on canvas `

to 

`if (plot.varexp.find(":time") != std::string::npos || plot.varexp.find(":run") != std::string::npos) { // we plot vs time, multiple plots on canvas `

Barthelemy:
Hi, 
Where are we with this ? 
Cheers,

Modules/Common/src/TH1SliceReductor.cxx
mfasDa: Hi @mlesch , 

Thanks a lot for the review!

Yes, I tested the changes using the in a postprocessing of apass2 data from LHC22m on EMCAL output.

OK, I will revert the changes in the TH1SliceReductor.

mfasDa:
> Sorry for the additional comment, but on a second look I did find something in the `SliceTrendingTask.cxx` that would need to be changed. In line 175 one has to change
> 
> `if (plot.varexp.find(":time") != std::string::npos) { // we plot vs time, multiple plots on canvas `
> 
> to
> 
> `if (plot.varexp.find(":time") != std::string::npos || plot.varexp.find(":run") != std::string::npos) { // we plot vs time, multiple plots on canvas `

Indeed I overlooked this - thanks for spotting! Fixed.

mfasDa:
@Barthelemy @mlesch I have pushed an update. Both comments are addressed.

mlesch (approved):

Barthelemy (approved):

PR #1608, opened 2023-01-09
Mch geometry from ccdb fetcher

aphecetche:
@aferrero2707 here are a couple of changes for the track task : 

- remove custom fetching of the geometry, in favour of using the CCDB Fetcher like in the rest of AliceO2
- ensure we request MID tracks when MCH-MID matching is needed

(plus some printout changes)


Modules/MUON/Common/src/TracksTask.cxx
```diff
@@ -144,32 +158,34 @@ void TracksTask::monitorData(o2::framework::ProcessingContext& ctx)
...
   auto rofs = mRecoCont.getMCHTracksROFRecords();
   auto clusters = mRecoCont.getMCHTrackClusters();
   auto digits = ctx.inputs().get<gsl::span<o2::mch::Digit>>("mchtrackdigits");
 
+  ctx.inputs().get<TGeoManager*>("mchgeo");
+
   if (mSrc[GID::MCH] == 1) {
     ILOG(Info, Devel) << "Debug: MCH requested" << ENDM;
     if (true || mRecoCont.isTrackSourceLoaded(GID::MCH)) {
-      ILOG(Info, Devel) << "Debug: MCH source loaded" << ENDM;
```
aferrero2707, line 165:
```
if (true || mRecoCont.isTrackSourceLoaded(GID::MCH))
```
the `true` in the `if` statement makes the `isTrackSourceLoaded()` check useless. I suspect this is a leftover of some debugging code that we forgot to cleanup.

Modules/MUON/Common/src/TracksTask.cxx
```diff
@@ -144,32 +158,34 @@ void TracksTask::monitorData(o2::framework::ProcessingContext& ctx)
...
+      ILOG(Info, Devel) << "Debug: MCH source loaded " << mRecoCont.isTrackSourceLoaded(GID::MCH) << ENDM;
       mTrackPlotters[GID::MCH]->fillHistograms(mRecoCont);
     }
   }
   if (mSrc[GID::MCHMID] == 1) {
     ILOG(Info, Devel) << "Debug: MCHMID requested" << ENDM;
     if (true || mRecoCont.isMatchSourceLoaded(GID::MCHMID)) {
-      ILOG(Info, Devel) << "Debug: MCHMID source loaded" << ENDM;
```
aferrero2707, line 172: See previous comment.

Modules/MUON/MCH/src/TracksTask.cxx
```diff
@@ -115,10 +115,6 @@ void TracksTask::initialize(o2::framework::InitContext& /*ic*/)
 {
   ILOG(Info, Support) << "initialize TracksTask" << ENDM; // QcInfoLogger is used. FairMQ logs will go to there as well.
 
-  if (!o2::base::GeometryManager::isGeometryLoaded()) {
-    TaskInterface::retrieveConditionAny<TObject>("GLO/Config/Geometry");
-  }
-
   createTrackHistos();
```
aferrero2707, line 118: Do we also need the `ctx.inputs().get<TGeoManager*>("mchgeo");` statement in the `monitorData()` method of `o2::quality_control_modules::muonchambers::TracksTask`?

Modules/MUON/Common/src/TracksTask.cxx
```diff
@@ -144,32 +158,34 @@ void TracksTask::monitorData(o2::framework::ProcessingContext& ctx)
...
-
   if (!assertInputs(ctx)) {
     return;
   }
 
+  mRecoCont.collectData(ctx, *mDataRequest.get());
+
+  ILOG(Info, Devel) << "Debug: Collected data" << ENDM;
+
   ILOG(Info, Devel) << "Debug: Asserted inputs" << ENDM;
```
aferrero2707, line 169: This message should go before the call to `mRecoCont.collectData()`

aferrero2707 (changes requested):
In addition to the comments specific to the new commits, there is a bug in the computation of RAbs here: https://github.com/aphecetche/QualityControl/blob/mch-geometry-from-ccdb-fetcher/Modules/MUON/MCH/src/TracksTask.cxx#L308
The starting point for the extrapolation should be the original MCH track parameters, and not the parameters at the origin.

In a future PR we need to remove the code duplication between [MUON/MCH/src/TracksTask.cxx](https://github.com/aphecetche/QualityControl/blob/mch-geometry-from-ccdb-fetcher/Modules/MUON/MCH/src/TracksTask.cxx) and [MUON/Common/src/TrackPlotter.cxx](https://github.com/aphecetche/QualityControl/blob/mch-geometry-from-ccdb-fetcher/Modules/MUON/Common/src/TrackPlotter.cxx), but for the time being I would propose to fix the RAbs computation, since the required changes are rather simple.

aphecetche:
@aferrero2707 concerning `MUON/MCH/src/TracksTask.cxx` vs `MUON/Common/src/TrackTask.cxx`, could you remind me why we're still using the MCH one instead of just the Common one ?

Modules/MUON/Common/src/TracksTask.cxx
aphecetche: removed

Modules/MUON/Common/src/TracksTask.cxx
aphecetche: removed

Modules/MUON/MCH/src/TracksTask.cxx
aphecetche: hum, yes, we probably should. Assuming we still want to use that Task (see other comment/question).

Modules/MUON/Common/src/TracksTask.cxx
aphecetche: done

aphecetche:
btw, note that once merged this PR will require a change in the QC configuration(s) (the query for the task must now contain `GLO/Config/0?lifetime=condition&ccdb-path=GLO/Config/GeometryAligned`)

chiarazampolli:
Hello @aphecetche , 
Are you also aware of this development? 
https://github.com/AliceO2Group/QualityControl/pull/1586
I don't think you are using it, are you?
Chiara

aphecetche:
@chiarazampolli no, indeed I was not aware of that development... 
Will have a look and change this PR accordingly then...

knopers8:
OK, I will force-merge #1586 then, so you are not blocked.

aferrero2707:
> concerning MUON/MCH/src/TracksTask.cxx vs MUON/Common/src/TrackTask.cxx, could you remind me why we're still using the MCH one instead of just the Common one ?

The idea was to let each detector (MFT/MCH/MID) have its own specific task for the tracks QC, with the plots ending up in the detector-specific part of the QCDB tree. The rationale being that each detector probably wants some specific plots that are not produced by the generic task for global forward tracks.

It is clear that the current structure is far from optimal, with a lot of code duplication, so we need to find a more elegant way to implement this...

aphecetche:
@aferrero2707 PR updated with latest way of fetching the geometry from the CCDB (using #1586). The example configurations have been updated accordingly. For the record, the required change in the configurations is, in the end, to add the following for the task(s) requiring the geometry from the CCDB : 

```
  "grpGeomRequest" : {
           "geomRequest": "Aligned",
           "askGRPECS": "false",
           "askGRPLHCIF": "false",
           "askGRPMagField": "false",
           "askMatLUT": "false",
           "askTime": "false",
           "askOnceAllButField": "false",
           "needPropagatorD":  "false"
         },
```

I also fixed the Rabs computation for the "legacy" `MCH/TrackTasks` while we decide what to do exactly with that task.

aferrero2707:
@aphecetche @chiarazampolli is there anything (apart from the CI) preventing this one from being reviewed and eventually approved?

aphecetche:
@aferrero2707 actually I guess what's missing is approval(s) from @Barthelemy and/or @knopers8 ?

knopers8 (approved):

knopers8:
yes, and passing tests, though probably the failures are unrelated as usual

PR #1609, opened 2023-01-11
[QC-810] Fix the enabling logic for GRP requests

knopers8:
@aphecetche here it is fixed


aphecetche:
@knopers8 for some reason it seems the o2-dataflow CI misses the "new" ITSResponse dependency : is its alidist up-to-date ?

knopers8:
the usual alpide test is failing, unrelated, merging

PR #1618, opened 2023-01-20
Update ITCClusterTask

lbariogl:
Whit this PR:
- all TH*D are changed to TH*F
- unused plots are removed (monitor plots)
- z-phi cluster occupancy and size plots are added
- detailed (finer binning) cluster occupancy and size plots are added

Example of new QC layout can be found here: https://qcg-test.cern.ch/?page=layoutShow&layoutId=63c7b384f53a750fbab2821a

### DIFF
--- a/Modules/ITS/include/ITS/ITSClusterTask.h
+++ b/Modules/ITS/include/ITS/ITSClusterTask.h
@@ -57,45 +57,47 @@ class ITSClusterTask : public TaskInterface
   void addObject(TObject* aObject);
   void getJsonParameters();
   void createAllHistos();
-  void updateOccMonitorPlots();
 
   static constexpr int NLayer = 7;
   static constexpr int NLayerIB = 3;
-  TH2F* hClusterVsBunchCrossing;
+
   std::vector<TObject*> mPublishedObjects;
-  TH1F* hClusterSizeSummaryIB[7][48][9];
+
+  // Inner barrel
   TH1F* hClusterTopologySummaryIB[7][48][9];
   TH1F* hGroupedClusterSizeSummaryIB[7][48][9];
 
-  TH1F* hClusterSizeLayerSummary[7];
-  TH1F* hClusterTopologyLayerSummary[7];
-  TH1F* hGroupedClusterSizeLayerSummary[7];
-
   TH2F* hAverageClusterOccupancySummaryIB[7];
-  TH2F* hAverageClusterOccupancyMonitorIB[7]; // will be used in online data monitoring, showing occupation for the last N ROFs
   TH2F* hAverageClusterSizeSummaryIB[7];
-  TH2F* hAverageClusterSizeMonitorIB[7];
 
-  Int_t mClusterOccupancyIB[7][48][9];
-  Int_t mClusterOccupancyIBmonitor[7][48][9];
+  int mClusterOccupancyIB[7][48][9];
 
+  // Outer barrel
   TH1F* hGroupedClusterSizeSummaryOB[7][48];
   TH1F* hClusterSizeSummaryOB[7][48];
   TH1F* hClusterTopologySummaryOB[7][48];
 
   TH2F* hAverageClusterOccupancySummaryOB[7];
-  TH2F* hAverageClusterOccupancyMonitorOB[7]; // will be used in online data monitoring, showing occupation for the last N ROFs
   TH2F* hAverageClusterSizeSummaryOB[7];
-  TH2F* hAverageClusterSizeMonitorOB[7];
 
-  //  THnSparseD *sClustersSize[7];
+  int mClusterOccupancyOB[7][48][28];
+
+  // Layer synnary
+  TH1F* hClusterSizeLayerSummary[7];
+  TH1F* hClusterTopologyLayerSummary[7];
+  TH1F* hGroupedClusterSizeLayerSummary[7];
+
+  // General
+  TH2F* hClusterVsBunchCrossing;
   TH2F* mGeneralOccupancy;
+
+  int mClusterSize[7][48][28]; //[#layers][max staves][max lanes / chips]
+  int nClusters[7][48][28];
 
   const int mOccUpdateFrequency = 100000;
   int mDoPublish1DSummary = 0;
   int mNThreads = 1;
   int mNRofs = 0;
-  int mNRofsMonitor = 0;
   int nBCbins = 103;
   long int mTimestamp = -1;
   std::string mGeomPath = "./";
@@ -112,11 +114,6 @@ class ITSClusterTask : public TaskInterface
   const std::string mYlabels[NLayer * 2] = { "L6B(S24#rightarrow47)", "L5B(S21#rightarrow41)", "L4B(S15#rightarrow29)", "L3B(S12#rightarrow23)", "L2B(S10#rightarrow19)", "L1B(S08#rightarrow15)", "L0B(S06#rightarrow11)", "L0T(S00#rightarrow05)", "L1T(S00#rightarrow07)", "L2T(S00#rightarrow09)", "L3T(S00#rightarrow11)", "L4T(S00#rightarrow14)", "L5T(S00#rightarrow20)", "L6T(S00#rightarrow23)" };
 
   int mEnableLayers[7];
-  int mClusterSize[7][48][28] = { { { 0 } } }; //[#layers][max staves][max lanes / chips]
-  double mClusterSizeMonitor[7][48][28] = { { { 0 } } };
-  int nClusters[7][48][28] = { { { 0 } } };
-  Int_t mClusterOccupancyOB[7][48][28] = { { { 0 } } };
-  Int_t mClusterOccupancyOBmonitor[7][48][28] = { { { 0 } } };
 
   o2::itsmft::TopologyDictionary* mDict;
   o2::its::GeometryTGeo* mGeom;
--- a/Modules/ITS/include/ITS/ITSClusterTask.h
+++ b/Modules/ITS/include/ITS/ITSClusterTask.h
@@ -58,41 +58,47 @@ class ITSClusterTask : public TaskInterface
   void getJsonParameters();
   void createAllHistos();
 
+
   static constexpr int NLayer = 7;
   static constexpr int NLayerIB = 3;
 
   std::vector<TObject*> mPublishedObjects;
 
   // Inner barrel
-  TH1F* hClusterTopologySummaryIB[7][48][9];
-  TH1F* hGroupedClusterSizeSummaryIB[7][48][9];
+  TH1F* hClusterTopologySummaryIB[NLayer][48][9];
+  TH1F* hGroupedClusterSizeSummaryIB[NLayer][48][9];
 
-  TH2F* hAverageClusterOccupancySummaryIB[7];
-  TH2F* hAverageClusterSizeSummaryIB[7];
+  TH2F* hAverageClusterOccupancySummaryIB[NLayer];
+  TH2F* hAverageClusterSizeSummaryIB[NLayer];
 
-  int mClusterOccupancyIB[7][48][9];
+  int mClusterOccupancyIB[NLayer][48][9];
 
   // Outer barrel
-  TH1F* hGroupedClusterSizeSummaryOB[7][48];
-  TH1F* hClusterSizeSummaryOB[7][48];
-  TH1F* hClusterTopologySummaryOB[7][48];
+  TH1F* hGroupedClusterSizeSummaryOB[NLayer][48];
+  TH1F* hClusterSizeSummaryOB[NLayer][48];
+  TH1F* hClusterTopologySummaryOB[NLayer][48];
 
-  TH2F* hAverageClusterOccupancySummaryOB[7];
-  TH2F* hAverageClusterSizeSummaryOB[7];
+  TH2F* hAverageClusterOccupancySummaryOB[NLayer];
+  TH2F* hAverageClusterSizeSummaryOB[NLayer];
 
-  int mClusterOccupancyOB[7][48][28];
+  int mClusterOccupancyOB[NLayer][48][28];
 
   // Layer synnary
-  TH1F* hClusterSizeLayerSummary[7];
-  TH1F* hClusterTopologyLayerSummary[7];
-  TH1F* hGroupedClusterSizeLayerSummary[7];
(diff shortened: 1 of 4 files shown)


Modules/ITS/src/ITSClusterTask.cxx
```diff
@@ -339,73 +328,88 @@ void ITSClusterTask::reset()
...
 
-    hClusterTopologyLayerSummary[iLayer] = new TH1D(Form("Layer%d/ClusterTopologySummary", iLayer), Form("Layer%dClusterTopologySummary", iLayer), 300, 0, 300);
+    hClusterTopologyLayerSummary[iLayer] = new TH1F(Form("Layer%d/ClusterTopologySummary", iLayer), Form("Layer%dClusterTopologySummary", iLayer), 300, 0, 300);
     hClusterTopologyLayerSummary[iLayer]->SetTitle(Form("Cluster topology summary for Layer %d", iLayer));
     addObject(hClusterTopologyLayerSummary[iLayer]);
     formatAxes(hClusterTopologyLayerSummary[iLayer], "Cluster Topology (ID)", "counts", 1, 1.10);
     hClusterTopologyLayerSummary[iLayer]->SetStats(0);
 
-    if (iLayer < 3) {
+    hAverageClusterOccupancySummaryZPhi[iLayer] = new TH2F(Form("Layer%d/ClusterOccupancyZPhi", iLayer), Form("Cluster occupancy on Layer %d;z (cm);#phi (#circ);", iLayer), 200, -100., 100., 360, -180., 180.);
```
iravasen, line 392: `circ` does not seem to appear correctly on the GUI. It should be a GUI problem. Maybe put `degree`. Same for the other lines where you use `circ`

Modules/ITS/src/ITSClusterTask.cxx
```diff
@@ -339,73 +328,88 @@ void ITSClusterTask::reset()
...
+    if (mDoPublishDetailedSummary == 1) {
+      addObject(hAverageClusterOccupancySummaryZPhi[iLayer]);
+    }
+
+    hAverageClusterSizeSummaryZPhi[iLayer] = new TH2F(Form("Layer%d/ClusterSizeZPhi", iLayer), Form("Cluster size on Layer %d;z (cm);#phi (#circ);", iLayer), 200, -100., 100., 360, -180., 180.);
+    hAverageClusterSizeSummaryZPhi[iLayer]->SetStats(0);
+    hAverageClusterSizeSummaryZPhi[iLayer]->SetBit(TH1::kIsAverage);
+    hAverageClusterSizeSummaryZPhi[iLayer]->GetYaxis()->SetLabelSize(0.02);
+    hAverageClusterSizeSummaryZPhi[iLayer]->GetXaxis()->SetLabelSize(0.02);
+    if (mDoPublishDetailedSummary == 1) {
```
iravasen, line 406: here we could extend the `if` to the full definition of the histo. I mean, we can allocate memory only if we decide to use the histo. The same `if` will have to be added also in `monitorData` when filling the plots. This is more for the `async` usage, to save memory (if not needed then we will always enable these plots). Same for all other `if`. But here I would wait for @IsakovAD answer.

Modules/ITS/src/ITSClusterTask.cxx
```diff
@@ -279,50 +299,19 @@ void ITSClusterTask::monitorData(o2::framework::ProcessingContext& ctx)
           mGeneralOccupancy->SetBinError(xbin, ybin, 1e-15);
         }
       }
-    }
-  }
+      hAverageClusterSizeSummaryZPhi[iLayer]->Divide(hAverageClusterOccupancySummaryZPhi[iLayer]);
+      hAverageClusterOccupancySummaryZPhi[iLayer]->Scale(1. / mNRofs);
```
iravasen, line 303: This is correct but at the beginning of `monitorData` before starting to fill the histos again you have to add:
```
hAverageClusterSizeSummaryZPhi[iLayer]->Multiply(hAverageClusterOccupancySummaryZPhi[iLayer]);
hAverageClusterOccupancySummaryZPhi[iLayer]->Scale(mNRofs);
```
This is due to the fact that `monitorData` runs continuously on different TFs and hence, before refilling you have to remove the scaling factors. On the other side you would rescale too many times with these lines.

Modules/ITS/src/ITSClusterTask.cxx
```diff
@@ -239,23 +248,34 @@ void ITSClusterTask::monitorData(o2::framework::ProcessingContext& ctx)
...
+      mGeom->fillMatrixCache(o2::math_utils::bit2Mask(o2::math_utils::TransformType::L2G));
+      auto gloC = mGeom->getMatrixL2G(ChipID) * locC;
+      float phi = (float)TMath::ATan2(gloC.Y(), gloC.X());
+
+      phi = (float)(phi * 180 / TMath::Pi());
+      hAverageClusterOccupancySummaryZPhi[lay]->Fill(gloC.Z(), phi);
+      hAverageClusterSizeSummaryZPhi[lay]->Fill(gloC.Z(), phi, (float)npix);
+
+      hAverageClusterOccupancySummaryCoarse[lay]->Fill(getHorizontalBin(locC.Z(), chip, lay, lane), getVerticalBin(locC.X(), sta, lay));
+      hAverageClusterSizeSummaryCoarse[lay]->Fill(getHorizontalBin(locC.Z(), chip, lay, lane), getVerticalBin(locC.X(), sta, lay), (float)npix);
```
iravasen, line 262: I would add also a `SetBinError` for these 4 guys (see the others TH2 with the bit `kisAverage` set). So far we did not care about the error, hence we set it to `1e-15` (a random small value). This is important for the QC merger that in this way will make an average of the plot "without" taking the bin errors as weights.

Modules/ITS/include/ITS/ITSClusterTask.h
```diff
@@ -51,75 +51,100 @@ class ITSClusterTask : public TaskInterface
...
+  TH2F* hClusterVsBunchCrossing = nullptr;
+  TH2F* mGeneralOccupancy = nullptr;
 
-  TH2D* hAverageClusterOccupancySummaryIB[7];
-  TH2D* hAverageClusterOccupancyMonitorIB[7]; // will be used in online data monitoring, showing occupation for the last N ROFs
-  TH2D* hAverageClusterSizeSummaryIB[7];
-  TH2D* hAverageClusterSizeMonitorIB[7];
+  // Coarse checks
+  TH2F* hAverageClusterOccupancySummaryCoarse[NLayer] = { nullptr };
+  TH2F* hAverageClusterSizeSummaryCoarse[NLayer] = { nullptr };
```
iravasen, line 105: a curiosity, why `coarse` if they have actually a finer binning wrt the ones we have already in the task?

Modules/ITS/src/ITSClusterTask.cxx
```diff
@@ -239,23 +248,34 @@ void ITSClusterTask::monitorData(o2::framework::ProcessingContext& ctx)
           hGroupedClusterSizeLayerSummary[lay]->Fill(npix);
         }
       }
+
+      // Transformation to the local --> global
+      mGeom->fillMatrixCache(o2::math_utils::bit2Mask(o2::math_utils::TransformType::L2G));
```
IsakovAD, line 253: Is it necessary to update matrix for each cluster?

Modules/ITS/src/ITSClusterTask.cxx
IsakovAD: Indeed it would help a lot with the memory usage; 

Just one point: we need to be accurate since we should also add an "if" case during the reset and delete of these objects as well.

Modules/ITS/itsCluster.json
```diff
@@ -37,18 +36,17 @@
...
-              "geomstamp": "1640991600000"
-      
+        "taskParameters": {
+          "layer": "1111111",
+          "nThreads": "1",
+          "nBCbins": "103",
+          "dicttimestamp": "0",
+          "geomPath": "./",
+          "publishSummary1D": "0",
+          "publishDetailedSummary": "1",
```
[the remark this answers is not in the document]
lbariogl: OK, I will do it

Modules/ITS/src/ITSClusterTask.cxx
```diff
@@ -279,50 +299,19 @@ void ITSClusterTask::monitorData(o2::framework::ProcessingContext& ctx)
...
+      hAverageClusterSizeSummaryZPhi[iLayer]->Divide(hAverageClusterOccupancySummaryZPhi[iLayer]);
+      hAverageClusterOccupancySummaryZPhi[iLayer]->Scale(1. / mNRofs);
 
-  if (mNRofsMonitor >= mOccUpdateFrequency) {
-    updateOccMonitorPlots();
-    mNRofsMonitor = 0;
-    memset(mClusterOccupancyIBmonitor, 0, sizeof(mClusterOccupancyIBmonitor));
-    memset(mClusterOccupancyOBmonitor, 0, sizeof(mClusterOccupancyOBmonitor));
+      hAverageClusterSizeSummaryZPhi[iLayer]->Divide(hAverageClusterOccupancySummaryZPhi[iLayer]);
+      hAverageClusterOccupancySummaryZPhi[iLayer]->Scale(1. / mNRofs);
```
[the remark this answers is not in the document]
lbariogl: Yes, it is indeed a bug. Thanks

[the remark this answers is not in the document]
lbariogl: First you need to evaluate the total size and then to divide by the number of clusters.
It is the first time that I scale `hAverageClusterOccupancySummaryZPhi`. It is not done before

Modules/ITS/src/ITSClusterTask.cxx
```diff
@@ -339,73 +328,88 @@ void ITSClusterTask::reset()
...
-      hAverageClusterSizeMonitorOB[iLayer]->Reset();
-      for (Int_t iStave = 0; iStave < mNStaves[iLayer]; iStave++) {
-
+      for (int iStave = 0; iStave < mNStaves[iLayer]; iStave++) {
         hClusterTopologySummaryOB[iLayer][iStave]->Reset();
         hGroupedClusterSizeSummaryOB[iLayer][iStave]->Reset();
       }
     }
+    hAverageClusterOccupancySummaryZPhi[iLayer]->Reset();
+    hAverageClusterSizeSummaryZPhi[iLayer]->Reset();
```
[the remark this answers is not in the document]
lbariogl: Ok

Modules/ITS/src/ITSClusterTask.cxx
lbariogl: `#degree` does not exist in ROOT LatEx symbols. `#circ` should be the correct one: I also tried locallu on a canvas and it works. I cannot understand what's going wrong :/

Modules/ITS/src/ITSClusterTask.cxx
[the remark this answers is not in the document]
lbariogl: Ok, now I get it: I do it twice for Z-Phi. Once it should be for fine binning

Modules/ITS/src/ITSClusterTask.cxx
lbariogl: Done and now fine-binned plots are alway within if guards. I thounk though that in a separate PR we should make all the histograms `unique_ptr`, so that wa are more safe with delete operations

Modules/ITS/src/ITSClusterTask.cxx
iravasen: I meant to use the full word "degree" without any symbol. It's probably the GUI which cannot display this symbol.

Modules/ITS/src/ITSClusterTask.cxx
lbariogl: OK, I did the same also for fine-binned histograms

Modules/ITS/src/ITSClusterTask.cxx
lbariogl: You're right. I have moved it at the beginning

Modules/ITS/src/ITSClusterTask.cxx
lbariogl: Done

Modules/ITS/src/ITSClusterTask.cxx
lbariogl: Done

iravasen:
@lbariogl thanks a lot! To me it looks all fine, @IsakovAD what do you think? 
@lbariogl have you checked whether the z-axis values of the TH2 plots make sense now? I guess that with the previous normalization at every QC cycle, the z values were off quite a bit.

Modules/ITS/src/ITSClusterTask.cxx
```diff
@@ -180,6 +182,15 @@ void ITSClusterTask::monitorData(o2::framework::ProcessingContext& ctx)
 #endif
   // Filling cluster histogram for each ROF by open_mp
 
+  if (mDoPublishDetailedSummary == 1) {
```
IsakovAD, line 186: @lbariogl 

When I am building your code, I have the error: 

>error: for statement expected before 'if'
This error comes from the fact that OpenMP would require a "for loop" after the OpenMP pragma expression. Do you have a such error? Interestingly, it appeared only for my case. 

Could you move this new if statement before the OpenMP parallelization? For my case, it solved the issue

lbariogl: @IsakovAD I have not tried, but you are right. I have moved it

IsakovAD:
@iravasen for me all is fine

iravasen:
ok also for me all is fine.

iravasen (approved):

iravasen:
@knopers8 O2fst fails I guess because we need to add the same option to the jsons of O2DPG (@IsakovAD is doing it). Is there a way to run the two remaining tests?

knopers8:
The only way I know is to try retriggering them by adding some dummy commit :/

Perhaps you would be able to make the new options backwards-compatible, i.e. nothing fails when these are missing?

iravasen:
Thanks @knopers8, I try with a dummy commit. Typically (learned from David) O2fst fails when the option in the json is not present in the same jsons in O2DPG. This time there was an error also about TPC so we are not so sure anymore. Let's see how it goes now.

PR #1619, opened 2023-01-20
fix QC-904

bazinski:
Fix for jira ticket QC-904 
also dropped Task from the classes for Digits and Trackletsl as per request from @Barthelemy  in spreadsheet : https://docs.google.com/spreadsheets/d/1uMAdnoya5RSdBsKCZ-fXIIgrBMnFWYrCS_KBD-1ccqk/edit#gid=478496474


bazinski:
@knopers8 I am not sure if this reinstantiates a bug with multilple log entries. I need to have multiple monitoring objects and merge them from the qc objects produced by this pr. Do you have an example of how to do that?

knopers8:
`o2-qc-file-merger` is what you need. It's fairly simple, `--help` should give you an idea how to use it. You can use the same input object and merge it a few times.

Concerning the renaming, only task names are being changed, so the string that is used as the key of the config structures (or `taskName` if you use it) and is used in the paths in QCDB. On the other hand the class name can stay as it is if you prefer. If you decide to change it as well, you will have to coordinate these changes with O2DPG and consul at P2.

bazinski:
@knopers8 I thought that only took monitoring objects, and the things I save from my json are histograms. How do I convert to monitoring objects? or am I missing things completely.

knopers8:
If you used `--local-batch` in your QC command, then you produced monitor objects. Otherwise, you actually can produce them this way, see [this](https://github.com/AliceO2Group/QualityControl/blob/master/doc/Advanced.md#batch-processing) for more details.

Otherwise, you can just open your object with ROOT shell and use the plain `Merge()` methods on the raw objects.

bazinski:
Those errors dont seem to be related to the code in this PR, and I tested with the merging of monitoring objects and previous pile of log messages are not there.

knopers8 (approved):
Indeed, thank you!

PR #1622, opened 2023-01-25
[QC-932] add a new repo cleaner policy

Barthelemy:
because the production policy was too complicated and this is good enough according to detectors.

### DIFF
--- a/Framework/script/RepoCleaner/qcrepocleaner/policies_utils.py
+++ b/Framework/script/RepoCleaner/qcrepocleaner/policies_utils.py
@@ -1,0 +1,32 @@
+from datetime import datetime
+from datetime import timedelta
+import logging
+from typing import DefaultDict, List
+
+from qcrepocleaner.Ccdb import Ccdb, ObjectVersion
+
+logger = logging  # default logger
+
+def in_grace_period(version: ObjectVersion, delay: int):
+    return not (version.validFromAsDt < datetime.now() - timedelta(minutes=delay))
+
+
+def get_run(v: ObjectVersion) -> str:
+    run = "none"
+    if "Run" in v.metadata:
+        run = str(v.metadata['Run'])
+    elif "RunNumber" in v.metadata:
+        run = str(v.metadata['RunNumber'])
+    return run
+
+
+def group_versions(ccdb, object_path, period_pass, versions_buckets_dict: DefaultDict[str, List[ObjectVersion]]):
+    # Find all the runs and group the versions (by run or by a combination of multiple attributes)
+    versions = ccdb.getVersionsList(object_path)
+    for v in versions:
+        logger.debug(f"Assigning {v} to a bucket")
+        run = get_run(v)
+        period_name = v.metadata.get("PeriodName") or ""
+        pass_name = v.metadata.get("PassName") or ""
+        key = run + period_name + pass_name if period_pass else run
+        versions_buckets_dict[key].append(v)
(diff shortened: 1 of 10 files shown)


Framework/script/RepoCleaner/qcrepocleaner/rules/multiple_per_run.py
```diff
@@ -0,0 +1,130 @@
...
+    update_list: List[ObjectVersion] = []
+    versions_buckets_dict: DefaultDict[str, List[ObjectVersion]] = defaultdict(list)
+    metadata_for_preservation = {'preservation': 'true'}
+
+    # config parameters
+    period_pass = (extra_params.get("period_pass", False) == True)
+    logger.debug(f"period_pass : {period_pass}")
+    interval_between_versions = int(extra_params.get("interval_between_versions", 30))
+    logger.debug(f"interval_between_versions : {interval_between_versions}")
+    migrate_to_EOS = (extra_params.get("migrate_to_EOS", False) is True)
```
knopers8, line 62: Is there a reason why you compare `period_pass` using `== True`, but `migrate_to_EOS` with `is True`?

Framework/script/RepoCleaner/qcrepocleaner/rules/multiple_per_run.py
```diff
@@ -0,0 +1,130 @@
...
+def process(ccdb: Ccdb, object_path: str, delay: int,  from_timestamp: int, to_timestamp: int,
+            extra_params: Dict[str, str]):
+    '''
+    Process this deletion rule on the object. We use the CCDB passed by argument.
+    Objects who have been created recently are spared (delay is expressed in minutes).
+
+    This specific policy, multiple_per_run, operates like this : Keep the first and the last version of a run plus
+    1 version every X minutes.
+
+    Extra parameters:
```
knopers8, line 23: `period_pass` and `object_attribute` are not explained as extra parameters

knopers8 (commented):
Some minor questions, but it looks good overall, thank you

Framework/script/RepoCleaner/qcrepocleaner/policies_utils.py
```diff
@@ -0,0 +1,32 @@
...
+from datetime import timedelta
+import logging
+from typing import DefaultDict, List
+
+from qcrepocleaner.Ccdb import Ccdb, ObjectVersion
+
+logger = logging  # default logger
+
+def in_grace_period(version: ObjectVersion, delay: int):
+    return not (version.validFromAsDt < datetime.now() - timedelta(minutes=delay))
```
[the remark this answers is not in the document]
Barthelemy: indeed, we should perhaps move to createdAt here. I have already used it in the new policy.

Framework/script/RepoCleaner/qcrepocleaner/rules/multiple_per_run.py
Barthelemy: good point. Modern Python dictates to use `is True`. I will change the other one.

Framework/script/RepoCleaner/qcrepocleaner/rules/multiple_per_run.py
Barthelemy: yes. and actually `object_attribute` is not used anymore.

knopers8 (approved):
thank you

PR #1623, opened 2023-01-26
Make sure fmt is included

ktf:
Currently relies on Logger.h being included in a place where it should not be.


ktf:
@Barthelemy could you cut a PR with this?

Barthelemy:
@ktf  yes

Barthelemy:
@ktf It seems that o2-dataflow is stuck...

PR #1624, opened 2023-01-27
Allow multiple rules

Barthelemy:
Depends on https://github.com/AliceO2Group/QualityControl/pull/1622


Framework/script/RepoCleaner/qcrepocleaner/rules/1_per_hour.py
```diff
@@ -27,7 +27,7 @@ def process(ccdb: Ccdb, object_path: str, delay: int,  from_timestamp: int, to_t
     :return a dictionary with the number of deleted, preserved and updated versions. Total = deleted+preserved.
     '''
     
-    logger.debug(f"Plugin 1_per_hour processing {object_path}")
+    logger.debug(f"aPlugin 1_per_hour processing {object_path}")
```
knopers8, line 30: typo?

knopers8 (approved):
It looks good, I would just clarify in the doc, that having multiple rules implies deleting objects that match to any of the provided rules. One could understand this the opposite - if any of the rules mark the object to preserve, it is preserved.

PR #1627, opened 2023-01-30
Use unique bindings in ROOT file source outputs

knopers8:
fixes the problem of objects from different detectors arriving to different check sinks


knopers8:
most of the checks passed, o2-dataflow and o2fst are stuck, merging it get it into the nightly

PR #1629, opened 2023-02-01
[WIP] TPC: add occupancy plot to Cluster task

tklemenz:


tklemenz:
This needs https://github.com/AliceO2Group/AliceO2/pull/10658.

wiechula (commented):
Looks ok so far. I guess we need to wait for the PR in O2.

Barthelemy:
@wiechula what should we do with this PR ? do you want to fix the conflict and we merge ?

wiechula:
@Barthelemy , it seems the O2 dependency was not merged, I'll check with @tklemenz .

tklemenz:
Exactly, the O2 dependency is missing, so merging now probably doesn't make much sense. I won't be in the office for a while but I'll move the task on someone else. So maybe it's the best if we close this PR opened by me and the one finishing the task re-opens.

PR #1634, opened 2023-02-06
Fix warnings and missing headers

ktf:


ktf:
The real issue I am trying to fix is the missing `#include <set>` which breaks https://github.com/AliceO2Group/AliceO2/pull/10659. I need .3 I guess..

knopers8 (approved):

ktf:
Can I get a tag with this as well?

PR #1637, opened 2023-02-10
[QC-916] Improvements in ROOT file sink

knopers8:


Barthelemy:
For reference: https://root-forum.cern.ch/t/should-the-result-of-tdirectory-getdirectory-be-deleted/53427

Barthelemy (approved):

PR #1641, opened 2023-02-14
MCH: fix clustermap filling in QC tracks

aphecetche:
Also remove unused lines of code (referencing Digit where they are not needed).
As far as I can tell the Digit are not used in any of the two QC tracks (MUON/Common or MCH) so I've removed them from the respective queries in the sample jsons as well.

Tested mainly starting from a `mchtracks.root` file (instead of a full reco workflow) to check we're only using information we actually have in the file.

```
o2-mch-tracks-reader-workflow --hbfutils-config $HOME/Downloads/o2_tfidinfo.root --infile $HOME/Downloads/mchtracks.root | o2-qc --config json://./mch-tracks-2.json --local-batch="qc-2.root" | o2-dpl-run -b
```

@aferrero2707 we'll definitely need to consolidate our MCH track QC, but this PR should solve the immediate problem of no longer having the cluster map available _and_ correctly filled.

### DIFF
--- a/Modules/MUON/Common/etc/qc-tracks-mch.json
+++ b/Modules/MUON/Common/etc/qc-tracks-mch.json
@@ -23,7 +23,7 @@
         "maxNumberCycles": "-1",
         "dataSource": {
           "type": "direct",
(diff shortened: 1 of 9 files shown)


aferrero2707:
@aphecetche looks ok to me. Indeed we do not need the digits in the tracks QC.

Also I agree that we need to merge the tracks QC between the global part and the MCH-specific part, and avoid code duplication. But that's for a future PR...

PR #1645, opened 2023-02-21
[IMPORTANT] RDH v6 to v7 in ITS FEE

iravasen:
Important - this task crashes right now with FLP suite 0.89.1

### DIFF
--- a/Modules/ITS/src/ITSFeeTask.cxx
+++ b/Modules/ITS/src/ITSFeeTask.cxx
@@ -329,7 +329,10 @@ void ITSFeeTask::monitorData(o2::framework::ProcessingContext& ctx)
   resetLanePlotsAndCounters(); // action taken depending on mResetLaneStatus and mResetPayload
 
   for (auto it = parser.begin(), end = parser.end(); it != end; ++it) {
-    auto const* rdh = it.get_if<o2::header::RAWDataHeaderV6>();
+    auto const* rdh = it.get_if<o2::header::RAWDataHeaderV7>();
+    if(!rdh) {
+      ILOG(Debug, Devel) << "We ask RDHv7 but you are running with another version, please fix it";
+    }
     // Decoding data format (RDHv6)
     int istave = (int)(rdh->feeId & 0x00ff);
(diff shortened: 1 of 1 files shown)


Modules/ITS/src/ITSFeeTask.cxx
```diff
@@ -329,7 +329,10 @@ void ITSFeeTask::monitorData(o2::framework::ProcessingContext& ctx)
   resetLanePlotsAndCounters(); // action taken depending on mResetLaneStatus and mResetPayload
 
   for (auto it = parser.begin(), end = parser.end(); it != end; ++it) {
-    auto const* rdh = it.get_if<o2::header::RAWDataHeaderV6>();
+    auto const* rdh = it.get_if<o2::header::RAWDataHeaderV7>();
+    if(!rdh) {
+      ILOG(Debug, Devel) << "We ask RDHv7 but you are running with another version, please fix it";
```
knopers8, line 334:
```suggestion
      ILOG(Debug, Devel) << "We ask RDHv7 but you are running with another version, please fix it" << ENDM;
```

iravasen: too fast, thanks! Give me few min, I wanted to run a test

shahor02: @iravasen One can have it in a version-independent way as:
```
const auto& rdh = *reinterpret_cast<const header::RDHAny*>(it.raw());
auto feeID = o2::raw::RDHUtils::getFEEID(rdh);
...
```
See other getters of https://github.com/AliceO2Group/AliceO2/blob/90a2345c0c322dafcfa6c33b2fa2a341649ff754/Detectors/Raw/include/DetectorsRaw/RDHUtils.h#L36

iravasen:
I tested it and looks good

knopers8:
Do we wait for @IsakovAD review? I will release new QC today afternoon, it will go to the next FLP suite.

IsakovAD (approved):
Sorry for the delay. All is fine

PR #1653, opened 2023-02-27
Apply clang-tidy suggestions to the framework

knopers8:


Barthelemy (approved):
`const`, `const` everywhere !
thank you and I will trust clang-tidy on this one.

PR #1654, opened 2023-02-27
SliceTrending: Add option to add branch to ongoing trending

mlesch:
Changes:
- SliceTrendingTask:  Possibility to add a new histogram into an ongoing trending. 
  - Remaining issue: If someone removes a histogram (as datasource) from an ongoing trending and then adds it back at some later point, then the drawing does not match with the time. 
  - @knopers8: I can add something similar for TrendingTask. However, I am not sure how to manipulate TTree Draw in such a way that I only draw the new datapoints for the newly added histogram. Maybe via the "selection" option?
- TH1SliceReductor: Changes so that one can trend single bin along y axis

### DIFF
--- a/Framework/src/SliceTrendingTask.cxx
+++ b/Framework/src/SliceTrendingTask.cxx
@@ -79,8 +79,13 @@ void SliceTrendingTask::initialize(Trigger, framework::ServiceRegistryRef servic
     mTrend->SetBranchAddress("meta", &(mMetaData.runNumber)); // TO-DO: Find reason why simply &mMetaData does not work
     mTrend->SetBranchAddress("time", &mTime);
     for (const auto& source : mConfig.dataSources) {
+      bool existingBranch = mTrend->GetBranchStatus(source.name.c_str());
       mSources[source.name] = new std::vector<SliceInfo>();
-      mTrend->SetBranchAddress(source.name.c_str(), &mSources[source.name]);
+      if (existingBranch) {
+        mTrend->SetBranchAddress(source.name.c_str(), &mSources[source.name]);
+      } else {
+        mTrend->Branch(source.name.c_str(), &mSources[source.name]);
+      }
     }
   }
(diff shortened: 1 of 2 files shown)


knopers8 (approved):
Thanks, looks good!

Regarding TrendingTask, what actually happens if we just ask `TTree::Draw` to draw some incomplete points? If it just ignored them, then perhaps it would be enough to modify `initialize` like you did?

mlesch:
I'm not sure what would happen in that case. The problem is, that the `time` branch and the datasource branch has different number of entries if its corresponding datasource/histogram is added later then some others. As far as I understand it, apart from certain special cases, one can not say which entry in the datasource branch then matches with which entry in the `time` branch. That's why I was thinking if for example the new histogram branch has `n` entries, and the time branch as `m` (with `m>n`), then we only take the `n` newest entries from time branch for drawing via the selection option.

knopers8:
Indeed, you are right, we cannot know how entries in different branches match to each other or actually ROOT assumes that their index is how they should be matched. I just tried using the `selection`, but in such case `TTree::Draw` just extrapolates the last points of the incomplete branches instead of taking `n` last of each... thus we would have to completely reimplement the drawing in `TrendingTask` by avoiding `Draw`, which would essentially make it `SliceTrendingTask`.

So I would just add a check in `TrendingTask` if the number and names of branches match and if not, spit an error and create a tree from scratch. Otherwise we risk creating plots which do not make sense and it might not be easy to detect.

mlesch:
I agree, this might be the best and safest solution for now. I will also have to think about some additional safety protection for the `SliceTrendingTask`. Because if we trend a histogram, then remove it but continue the trending task and then add it back again, then also in this case the plots will not make any sense.

knopers8:
Indeed, that's also a risk. So I guess we want to start from scratch whenever we remove a column.

PR #1664, opened 2023-03-06
[QC-942] [QC-909] Use Bookkeeping in QC

Barthelemy:


Framework/include/QualityControl/Activity.h
```diff
@@ -67,12 +68,15 @@ class Activity
...
 
+  void importFromRun(const std::shared_ptr<o2::bookkeeping::Run>& run);
+
   int mId{ 0 };
   int mType{ 0 };
   std::string mPeriodName{};
   std::string mPassName{};
   std::string mProvenance{ "qc" };
   ValidityInterval mValidity{ gFullValidityInterval };
+  std::string mBeamType{};
```
knopers8, line 79: We could actually take the opportunity to enforce enumerated beam types. It would be risky to require from the users to know if the string is going to be "PROTON-PROTON" or "pp", or "p-p".

Framework/include/QualityControl/Activity.h
```diff
@@ -22,6 +22,7 @@
 #include "QualityControl/ValidityInterval.h"
 
 #include "Rtypes.h"
+#include "BookkeepingApi/run.pb.h" // TODO beurk
```
knopers8, line 25: I am wondering if we should have a bunch of our structures have some interface to BK directly in their classes or better have one BK interface which can read/write all the concerned structures. If we do the latter, the advantages are:
- BK API will not be scattered everywhere in the QC code - one unit which communicates with BK. Easier to update the code related to BK if it is one place.
- probably faster compilation due to less inclusion of the BK headers

Framework/src/TaskRunner.cxx
```diff
@@ -414,16 +416,34 @@ void TaskRunner::printTaskConfig() const
                     << " / Save to file : " << mTaskConfig.saveToFile << ENDM;
 }
 
+Activity TaskRunner::createActivityObject()
```
knopers8, line 419: I would not put it to `TaskRunner`, other kinds of QC actors might also need to do it.

Barthelemy: yes, I was planning to move it to the `runnerUtils`

Framework/include/QualityControl/Activity.h
Barthelemy: I am not sure that we should do it on our side. Rather in BK

knopers8: Whoever does it, I think that the user code should not rely on comparing hardcoded strings.

knopers8: Somewhere, maybe in `Common` or in `AliceO2` we could have some class to define the collision system. Maybe there is one already? (cc @shahor02 ?)

Barthelemy:
this needs https://github.com/alisw/alidist/pull/4899

Framework/src/Bookkeeping.cxx
```diff
@@ -0,0 +1,61 @@
...
+{
+
+void Bookkeeping::init(std::string url)
+{
+  if (url.empty()) {
+    ILOG(Warning, Support) << "No URL provided for Bookkeeping. Nothing will be stored nor retrieved." << ENDM;
+    return;
+  }
+  try {
+    mClient = BkpProtoClientFactory::create(url);
```
knopers8, line 33: Can `create` return a nullptr instead of throwing an exception? Are we guaranteed that this does not happen in the future? I am afraid we might end up with a code path which has `mInitialized == true`, but `mClient == nullptr`.

Framework/src/Bookkeeping.cxx
```diff
@@ -0,0 +1,61 @@
...
+#include "QualityControl/Bookkeeping.h"
+#include "QualityControl/QcInfoLogger.h"
+#include "BookkeepingApi/BkpProtoClientFactory.h"
+
+using namespace o2::bkp::api::proto;
+
+namespace o2::quality_control::core
+{
+
+void Bookkeeping::init(std::string url)
```
knopers8, line 26: Once DPL is able to run several Data Processors multithreaded, we will initialize the BK several times for no reason. Perhaps you can add a check if Bookkeeping was already initialized with the same url and do nothing then, plus if the url is different than before, add a warning, but proceed?

knopers8 (changes requested):

Framework/include/QualityControl/Bookkeeping.h
```diff
@@ -0,0 +1,51 @@
...
+///
+/// @file    Bookkeeping.h
+/// @author  Barthelemy von Haller
+
+#ifndef QC_CORE_BOOKKEEPING_H
+#define QC_CORE_BOOKKEEPING_H
+
+#include <string>
+#include "QualityControl/Activity.h"
+#include "BookkeepingApi/BkpProtoClient.h"
```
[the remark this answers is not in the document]
Barthelemy: It fails because I have `  Bookkeeping() = default;` in this class and it tries to instantiate the unique_ptr.

Framework/src/Bookkeeping.cxx
Barthelemy: yes, done

Barthelemy:
I applied all your comments but one.

knopers8 (approved):

PR #1669, opened 2023-03-08
TPC: Update laser trending

mlesch:


Modules/TPC/src/LtrCalibReductor.cxx
```diff
@@ -61,6 +62,9 @@ void LtrCalibReductor::update(TObject* obj)
 
 double LtrCalibReductor::getValue(TText* line)
 {
+  if (!line)
+    return 0.;
```
wiechula, line 66: Always use braces, also for one line statements!
```suggestion
  if (!line) {
    return 0.;
    }
```

wiechula (changes requested):

wiechula (approved):

PR #1671, opened 2023-03-08
[QC-384] Flexible cycles

Barthelemy:


Barthelemy:
needs https://github.com/AliceO2Group/AliceO2/pull/10930

Framework/basic.json
```diff
@@ -39,7 +39,11 @@
         "className": "o2::quality_control_modules::skeleton::SkeletonTask",
         "moduleName": "QcSkeleton",
         "detectorName": "TST",
-        "cycleDurationSeconds": "10",     "": "10 seconds minimum",
+        "": "The last item in cycleDurations will be used for the rest of the duration whatever the period",
+        "cycleDurations": [
+          {"cycleDuration": 10, "period": 35},
+          {"cycleDuration": 12, "period": 1}
```
knopers8, line 45: as in O2, can you use `validity` instead of `period`? You could also consider adding back `seconds` somewhere.

Framework/src/InfrastructureGenerator.cxx
```diff
@@ -214,9 +214,15 @@ o2::framework::WorkflowSpec InfrastructureGenerator::generateRemoteInfrastructur
...
-
-      generateMergers(workflow, taskSpec.taskName, numberOfLocalMachines, cycleDurationSeconds, taskSpec.mergingMode,
+      std::vector<std::pair<size_t, size_t>> cycleDurationsMultiplied;
+      if (taskSpec.cycleDurationSeconds > 0) { // old, simple, style
+        cycleDurationsMultiplied = { { taskSpec.cycleDurationSeconds, 1 } };
+      } else { // new style
+        cycleDurationsMultiplied = taskSpec.cycleDurations;
+      }
+      std::for_each(cycleDurationsMultiplied.begin(), cycleDurationsMultiplied.end(),
+                    [taskSpec](std::pair<size_t, size_t>& p) { p.second *= taskSpec.mergerCycleMultiplier; });
```
knopers8, line 224: Are you sure it is the `second` to be multiplied? I assume that the validity should stay the same, but the cycle length should be longer.

Framework/src/InfrastructureSpecReader.cxx
```diff
@@ -93,7 +93,14 @@ TaskSpec InfrastructureSpecReader::readSpecEntry<TaskSpec>(const std::string& ta
   ts.className = taskTree.get<std::string>("className");
   ts.moduleName = taskTree.get<std::string>("moduleName");
   ts.detectorName = taskTree.get<std::string>("detectorName");
-  ts.cycleDurationSeconds = taskTree.get<int>("cycleDurationSeconds");
+  ts.cycleDurationSeconds = taskTree.get<int>("cycleDurationSeconds", -1);
+  if (taskTree.count("cycleDurations") > 0) {
+    for (const auto& cycleConfig : taskTree.get_child("cycleDurations")) {
+      auto duration = cycleConfig.second.get<size_t>("cycleDuration");
+      auto nbCycles = cycleConfig.second.get<size_t>("period");
+      ts.cycleDurations.push_back(std::pair{ duration, nbCycles });
```
knopers8, line 101: didn't you decide to use period/validity instead of `nbCycles`? I am confused now

Framework/src/TaskRunnerFactory.cxx
```diff
@@ -62,18 +64,21 @@ TaskRunnerConfig TaskRunnerFactory::extractConfig(const CommonSpec& globalConfig
...
-    ILOG(Error, Support) << "Cycle duration is too short (" << cycleDurationSeconds << "), replaced by a duration of 10 seconds." << ENDM;
-    cycleDurationSeconds = 10;
+
+  // cycle duration
+  // Two ways of configuring, incompatible.
+  // 1. simple, old, way: cycleDurationSeconds is the duration in seconds for all cycles
+  // 2. complex, new, way: cycleDurations: a list of tuples specifying different durations to be applied for a certain time
+  if (taskSpec.cycleDurationSeconds > 0 && taskSpec.cycleDurations.size() > 0) {
+    ILOG(Error, Ops) << "Both cycleDurationSeconds and cycleDurations have been defined. Pick one. Sheepishly exiting out." << ENDM;
+    exit(1);
```
knopers8, line 74: perhaps you can throw an exception to be consistent with other fatal errors? this way DPL/FairMQ may handle the exit more gracefully.

Framework/src/TaskRunnerFactory.cxx
```diff
@@ -152,6 +157,32 @@ TaskRunnerConfig TaskRunnerFactory::extractConfig(const CommonSpec& globalConfig
...
+                                                  const std::string& detectorName, const std::string& taskName)
+{
+  // This is to check that the durations are not below 10 seconds except when using a dummy database
+  auto dummyDatabaseUsed = globalConfig.database.count("implementation") > 0 && globalConfig.database.at("implementation") == "Dummy";
+  for (auto& [cycleDuration, period] : cycleDurations) {
+    if (!dummyDatabaseUsed && cycleDuration < 10) {
+      ILOG(Error, Support) << "Cycle duration is too short (" << cycleDuration << "), replaced by a duration of 10 seconds." << ENDM;
+      cycleDuration = 10;
+    }
+  }
```
knopers8, line 170: Since you are doing something in that loop only if `dummyDatabaseUsed` is true, why don't you not execute the loop at all in such case? ...and get rid of `!dummyDatabaseUsed` in the inner if statement?

Framework/src/TaskRunnerFactory.cxx
```diff
@@ -152,6 +157,32 @@ TaskRunnerConfig TaskRunnerFactory::extractConfig(const CommonSpec& globalConfig
...
+
+  // Create the TimerSpec for cycleDurations
+  std::vector<TimerSpec> timers;
+  for (auto& [cycleDuration, period] : cycleDurations) {
+    timers.push_back({ cycleDuration * 1000000000 /*µs*/, period });
+  }
```
knopers8, line 176:
```suggestion
  for (auto& [cycleDuration, validity] : cycleDurations) {
    timers.push_back({ cycleDuration * 1000000000 /*µs*/, validity });
  }
```

knopers8 (changes requested):

Framework/basic.json
Barthelemy: yes

Framework/src/InfrastructureGenerator.cxx
Barthelemy: correct

Framework/src/TaskRunnerFactory.cxx
Barthelemy: yes

Framework/src/InfrastructureSpecReader.cxx
Barthelemy: yes that was a leftover

Framework/src/TaskRunnerFactory.cxx
Barthelemy: of course

Barthelemy:
I tried to address all the comments.

knopers8 (approved):

PR #1676, opened 2023-03-14
ITS Track: added option to normalize by nROF + 6 new plots

iravasen:
Added in this PR:
- option `doNorm` to do: 0 = no normalization, 1 = normalization by nVertices, 2 = normalization by nROFs. The last one has to be used with cosmics. 
- 1 new plot with #clusters_per_tracks vs phi (was existing only the one vs eta)
- 5 new plots with "Layer with 1st hit" (check where a track has the first hit) vs phi for tracks with 4,5,6,7 clusters and any number of clusters

Plots intended for ITS experts and introduced to better understand the track distributions in case of holes in the acceptance (i.e. missing stave(s)).


IsakovAD (approved):
All is fine, not comments

iravasen:
@IsakovAD thanks!
@knopers8 the error for macOS test is not related to this PR, can you force merge? Otherwise I go with a dummy commit.

PR #1677, opened 2023-03-16
TRD remove bug causing labels QC-941

bazinski:
Remove the labels from the layer plots of digits and tracklets.
I still cant replicate the problem causing the qc05 to fill the disk with logs. It is clearly coming from the labels of these graphs.
I can replicate the same error message but it would be on the epn and only after adding code to fill all bins with labels, unlike the commented out sections of the code in this PR.

A more permanent solution will come later.


Barthelemy (approved):

bazinski:
@Barthelemy shot for the formatting, for some reason my clang-format is not finding all.

PR #1678, opened 2023-03-16
Optimizing tracklets and digits matching

vikasLstar:
The task has been modified to optimize the matching of tracklets and digits.

### DIFF
--- a/Modules/TRD/src/PulseHeightTrackMatch.cxx
+++ b/Modules/TRD/src/PulseHeightTrackMatch.cxx
@@ -64,20 +64,11 @@ void PulseHeightTrackMatch::buildHistograms()
   mParsingTimePerTF->GetYaxis()->SetTitle("Counts");
   getObjectsManager()->startPublishing(mParsingTimePerTF.get());
 
-  mDigitsPerEvent.reset(new TH1F("digitsperevent", "Digits per Event", 10000, 0, 10000));
-  getObjectsManager()->startPublishing(mDigitsPerEvent.get());
-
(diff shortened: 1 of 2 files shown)


Barthelemy (approved):

Modules/TRD/src/PulseHeightTrackMatch.cxx
```diff
@@ -107,172 +98,89 @@ void PulseHeightTrackMatch::startOfCycle()
...
+            if (digit.getPadRow() != trklt.getPadRow())
+              continue;
+            coldif = digit.getPadCol() - trklt.getPadCol();
+            if (TMath::Abs(coldif) > 1)
+              continue; // PadCol may be off by +-1 due to the slope of tracklet
+            const auto& digitBefore = digits[iDigit - 1];
+            const auto& digitAfter = digits[iDigit + 1];
+
+            bool consecutive = false;
+            if (digitBefore.getDetector() == digitAfter.getDetector() && digitBefore.getPadRow() == digit.getPadRow() && digit.getPadRow() == digitAfter.getPadRow() && digit.getPadCol() + 1 == digitBefore.getPadCol() && digitAfter.getPadCol() + 1 == digit.getPadCol())
```
martenole, line 164: Did you check whether this flag `consecutive` at some point is true? And please don't forget the curly braces even around one liners
```
if (...) {
}
```

vikasLstar: yes, after applying this condition what remains are only those digits that are consecutive in space. I also added a comment on this condition and put curly braces on one-liners also.
Thanks.

martenole (approved):
Thanks, looks good to me

PR #1680, opened 2023-03-17
Script to delete objects not in a list of runs

Barthelemy:
There is the opposite script already. I thought I could just add an option to invert its behaviour but it was actually not trivial. It is simpler and clearer to have a dedicated script.


knopers8 (approved):
It's also OK like this, but I am wondering if it was not easier to just invert some logic in `o2-qc-repo-delete-objects-in-runs` with an additional flag?

PR #1681, opened 2023-03-17
If possible, use --id argument instead runQC as IL facility

knopers8:
This a rather hacky way to retrieve the identity of a Data Processor before it defines its workflow.

This does not work, probably because "--id" is not one of the expected parameters for the user workflow:
```
auto id = config.options().hasOption("id") && config.options().isSet("id") ? config.options().get<std::string>("id") : "runQC";
```

While this improves the situation, we will still have e.g. "qc-task-ITS-ITSFEE" (device id) at the beginning of the task lifetime, but "task/ITSFEE" once it fully initializes. We could consider using device ID throughout the lifetime, to be consistent.


Barthelemy (approved):
agreed

Barthelemy:
@TimoWilken There seems to be a problem with `QC/o2-dataflow/macOS` : 
```
455/455 Test  #97: CcdbApi-MultiHandle ..................................................................***Timeout 1500.04 sec
```

PR #1684, opened 2023-03-20
MID::Digits Display Normalization

ValerieRamillien:


Barthelemy (changes requested):
Thank you for your PR. I have just 2 small suggestions.

Barthelemy:
@ValerieRamillien This is not on you. There is an issue in O2 that triggers this problem. I will force merge your changes. Thank you for applying the comments.

PR #1685, opened 2023-03-20
Pass activity to checks

Barthelemy:
The users' code did not have access to the Activity in the Check.


Barthelemy:
@knopers8 This is to give access to the Activity from the users' checks. This is a step in the direction of having users' custom parameters depending on the activity type.

knopers8 (changes requested):
Actually, this allows for any Check to modify the Activity for the whole CheckRunner. Can we avoid this?

PR #1686, opened 2023-03-20
FIT PostProc: hotfix, check for nullptr

afurs:


Barthelemy (approved):

Barthelemy:
Do you need a patch version with that ? if so, please contact me on MM.

afurs:
> Do you need a patch version with that ? if so, please contact me on MM.

yes, okay

PR #1687, opened 2023-03-21
A helper to aggregate the minimal matching Activity

knopers8:
This will be needed to obtain the correct validity of:
- a QualityObject obtained in a Check applied on more than one object
- a result of merging.


Barthelemy (approved):
very nice

PR #1691, opened 2023-03-27
[QC-946] Users params run type

Barthelemy:


knopers8:
It seems there are some legit test failures:
```
 7/42 Test  #9: testCheckInterface ......................***Failed    1.63 sec
27/42 Test #33: testCustomParameters ....................***Failed    1.54 sec
31/42 Test #37: testMeanIsAbove .........................***Failed    1.83 sec
```

Barthelemy:
yes, indeed, I'll fix it

Barthelemy:
I realized that I did not test the custom parameters for the check and aggregator. I put back the PR in WIP for the time being.

knopers8:
still broken.

tomas-herman:
Hi @Barthelemy, will this PR enable also setting different values of data sampling in the json files based on the run type? E.g. running our digit task with 0.1% sampling for cosmic/noise runs and 0.001% sampling for pp?

Barthelemy:
@tomas-herman no, it won't but that's a good idea. I would put it in a different PR though. Could you create a ticket perhaps ?

tomas-herman:
Sure, here it is: https://alice.its.cern.ch/jira/browse/QC-962

Framework/include/QualityControl/CustomParameters.h
```diff
@@ -0,0 +1,222 @@
...
+   * @return the value for the given key. If it does not exist, returns the provided default value.
+   */
+  std::string atOrDefaultValue(const std::string& key, std::string defaultValue) const
+  {
+    try {
+      return mCustomParameters.at("default").at("default").at(key);
+    } catch (const std::out_of_range& exc) {
+      return defaultValue;
+    }
+  }
```
knopers8, line 126: Maybe these two can be joined together by:
```
std::string atOrDefaultValue(const std::string& key, std::string defaultValue = "", const std::string& runType = "default", const std::string& beamType = "default")
```

knopers8: Then you could consider joining it also with `at` if you used `optional`.

Framework/src/InfrastructureSpecReader.cxx
```diff
@@ -107,9 +107,18 @@ TaskSpec InfrastructureSpecReader::readSpecEntry<TaskSpec>(const std::string& ta
...
+  if (taskTree.count("extendedTaskParameters") > 0) {
+    for (const auto& [runtype, subTreeRunType] : taskTree.get_child("extendedTaskParameters")) {
+      for (const auto& [beamtype, subTreeBeamType] : subTreeRunType) {
+        for (const auto& [key, value] : subTreeBeamType) {
+          ts.customParameters.set(key, value.get_value<std::string>(), runtype, beamtype);
+        }
+      }
+    }
+  }
   if (taskTree.count("taskParameters") > 0) {
```
knopers8, line 119: maybe we should complain if users use both at the same time?

doc/Advanced.md
```diff
@@ -992,8 +992,77 @@ The same approach can be applied to other actors in the QC framework, like Check
 
 ## Definition and access of task-specific configuration
 
+A task can access custom parameters declared in the configuration file at `qc.tasks.<task_id>.extendTaskParameters` or `qc.tasks.<task_id>.taskParameters`. They are stored inside an object of type `CustomParameters` named `mCustomParameters`, which is a protected member of `TaskInterface`.
```
knopers8, line 995:
```suggestion
A task can access custom parameters declared in the configuration file at `qc.tasks.<task_id>.extendedTaskParameters` or `qc.tasks.<task_id>.taskParameters`. They are stored inside an object of type `CustomParameters` named `mCustomParameters`, which is a protected member of `TaskInterface`.
```

doc/Advanced.md
```diff
@@ -992,8 +992,77 @@ The same approach can be applied to other actors in the QC framework, like Check
...
+          },
+          "physics": {
+            "default": {
+              "myOwnKey1": "myOwnValue1b",
+              "myOwnKey2": "myOwnValue2b"
+            },
+            "pp": {
+              "myOwnKey1": "myOwnValue1c"
+            },
+            "PbPb": {
```
knopers8, line 1027: Are "pp" and "PbPb" the correct examples of the strings to be used? The same comment for "physics" and "cosmics".

knopers8 (changes requested):
Thanks, please see my comments.

One thing that I don't understand is how the string for run type can be compared to `Activity.type` which is int. Are the users supposed to use numbers actually?

doc/Advanced.md
```diff
@@ -992,8 +992,77 @@ The same approach can be applied to other actors in the QC framework, like Check
 
 ## Definition and access of task-specific configuration
```
knopers8, line 993: Please update the index.

Framework/include/QualityControl/CustomParameters.h
```diff
@@ -0,0 +1,222 @@
...
+/// \file   CustomParameters.h
+/// \author Barthelemy von Haller
+///
+
+#ifndef QC_CUSTOM_PARAMETERS_H
+#define QC_CUSTOM_PARAMETERS_H
+
+#include <string>
+#include <unordered_map>
+#include <iostream>
```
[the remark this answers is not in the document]
Barthelemy: indeed, done.

Framework/src/InfrastructureSpecReader.cxx
Barthelemy: yes, I have put a warning and I actually only use `extendedTaskParameters` in such case.

doc/Advanced.md
Barthelemy: I have added an example for the Run Type that is defined in O2 in "DataFormatsParameters/ECSDataAdapters.h". 
Concerning the beam type I am still waiting for the information.

Barthelemy:
I agree that it is not extensible and this is to address the request from RC concerning the Run type and the beam type. 

I had to maintain the interface of CustomParameters similar to `vector` to avoid updating hundreds of files. 

The beam type is still not defined, ie. I don't have a translation between the int we get from BK and the string. Once it is there I will update the example and the doc.

doc/Advanced.md
Barthelemy: ok, found it, added doc.

Barthelemy:
@knopers8 I have addressed your comments as best as I could. Beam type is defined in the doc. Run type is defined in a class in O2. I have added a link to the doc. 
I also added an example in SkeletonTask.

doc/Advanced.md
```diff
@@ -992,8 +997,80 @@ The same approach can be applied to other actors in the QC framework, like Check
...
+          },
+          "COSMICS": {
+            "myOwnKey1": "myOwnValue1e",
+            "myOwnKey2": "myOwnValue2e"
+          }
+        },
+```
+It allows to have variations of the parameters depending on the run and beam types. The proper run types can be found here: https://github.com/AliceO2Group/AliceO2/blob/dev/DataFormats/Parameters/include/DataFormatsParameters/ECSDataAdapters.h#L54. The `default` can be used 
+to ignore the run or the beam type. 
+The beam type is one of the following: `PROTON-PROTON`, `Pb-Pb`, `Pb-PROTON`
```
knopers8, line 1044: Can you use these strings in the examples as well?

Framework/include/QualityControl/CustomParameters.h
```diff
@@ -0,0 +1,222 @@
...
+   * @param beamType
+   * @return a map of the key-value pairs
+   * @throw std::out_of_range if no key-value pair correspond to these beamType and runType
+   */
+  const std::unordered_map<std::string, std::string>& getAllForRunBeam(const std::string& runType, const std::string& beamType)
+  {
+    if (mCustomParameters.count(runType) > 0 && mCustomParameters[runType].count(beamType) > 0) {
+      return mCustomParameters[runType][beamType];
+    }
+    throw std::out_of_range("Unknown beam or run: " + runType + ", " + beamType);
```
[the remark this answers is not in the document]
knopers8: this is not resolved yet

Framework/include/QualityControl/CustomParameters.h
```diff
@@ -0,0 +1,213 @@
...
+ *   CustomParameters cp;
+ *   cp.set("key", "value"); // for the default run and beam type
+ *   cout << "value for key `key` : " << cp.at("key") << endl;
+ *
+ *   CustomParameters cp2;
+ *   cp2.set("key", "value_run1_beam1", "physics", "pp");
+ *   cout << "value for key `key` : " << cp.at("key", "physics", "pp") << endl;
+ *
+ */
+class CustomParameters
```
knopers8, line 44: I don't quite understand why we should recompile these methods for each time someone includes this header. Perhaps everything could be moved to a source file?

Barthelemy: ok I had missed your previous comment
fixed.

doc/Advanced.md
Barthelemy: I found one instance that I  changed

knopers8 (approved):

knopers8 (approved):

PR #1692, opened 2023-03-28
Do not get B field from external file anymore, but from CCDB - QualityControl

chiarazampolli:
@knopers8 , @shahor02 , goes with 
https://github.com/AliceO2Group/O2DPG/pull/977
https://github.com/AliceO2Group/AliceO2/pull/11044

### DIFF
--- a/Modules/ITS/src/ITSClusterTask.cxx
+++ b/Modules/ITS/src/ITSClusterTask.cxx
@@ -31,6 +31,7 @@
 #include "CCDB/BasicCCDBManager.h"
 #include "CCDB/CCDBTimeStampUtils.h"
 #include <Framework/InputRecord.h>
+#include "Framework/TimingInfo.h"
 #include <TH1F.h>
 #include <TH2F.h>
 #include "Common/Utils.h"
@@ -137,9 +138,10 @@ void ITSClusterTask::startOfCycle()
 
 void ITSClusterTask::monitorData(o2::framework::ProcessingContext& ctx)
 {
-
-  mGeom = o2::its::GeometryTGeo::Instance();
-  mGeom->fillMatrixCache(o2::math_utils::bit2Mask(o2::math_utils::TransformType::L2G));
+  if (ctx.services().get<o2::framework::TimingInfo>().globalRunNumberChanged) {
+    mGeom = o2::its::GeometryTGeo::Instance();
+    mGeom->fillMatrixCache(o2::math_utils::bit2Mask(o2::math_utils::TransformType::L2G));
+  }
 
   if (mTimestamp == -1) { // get dict from ccdb
     mTimestamp = std::stol(o2::quality_control_modules::common::getFromConfig<string>(mCustomParameters, "dicttimestamp", "0"));
--- a/Modules/ITS/src/ITSClusterTask.cxx
+++ b/Modules/ITS/src/ITSClusterTask.cxx
@@ -102,21 +102,6 @@ void ITSClusterTask::initialize(o2::framework::InitContext& /*ctx*/)
 
   getJsonParameters();
 
-  if (mLocalGeometryFile == 1) {
-    ILOG(Info, Support) << "Getting geometry from local file" << ENDM;
-    o2::base::GeometryManager::loadGeometry(mGeomPath.c_str());
-  } else {
-    ILOG(Info, Support) << "Getting geometry from ccdb - timestamp: " << std::stol(mGeoTimestamp) << ENDM;
-    std::map<std::string, std::string> metadata;
-    TaskInterface::retrieveConditionAny<TGeoManager>("GLO/Config/GeometryAligned", metadata, std::stol(mGeoTimestamp));
-    if (!o2::base::GeometryManager::isGeometryLoaded()) {
-      ILOG(Fatal, Support) << "Can't retrive geometry from ccdb with timestamp: " << std::stol(mGeoTimestamp) << ENDM;
-      throw std::runtime_error("Can't retrive geometry from ccdb!");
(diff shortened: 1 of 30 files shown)


tomas-herman:
Hi @chiarazampolli,

We need the geometry information because we are using a method to access additional information about the clusters: https://github.com/AliceO2Group/AliceO2/blob/346604524e24649cf599da2d0757759234bcaca1/Detectors/ITSMFT/MFT/alignment/src/AlignPointHelper.cxx#L206 

And this helper needs the geometry information. As far as I understand the geometry needs to be available in the QC task for the helper to be able to work. I will comment in the O2DPG about the second PR.

Cheers,
Tomas

chiarazampolli:
Hello @tomas-herman , 
I tried and everything worked, because to my understanding, changing the json adding the GeomRequest makes the geometry available. 
Can you please try with the PRs from O2, O2DPG, QC locally, and see if it behaves correctly? I checked a few MFT plots and before and after the change they were identical, but I did not check them all.
Chiara

Barthelemy (approved):

tomas-herman:
Hi @chiarazampolli 

I will try to do a local test but I need to rebuild first as I have old versions of the repositories.

Just from looking at the code it seems to me that the Geometry Request in the json i.e.
`	"grpGeomRequest" : {
          "geomRequest": "Aligned",
          "askGRPECS": "false",
          "askGRPLHCIF": "false",
          "askGRPMagField": "true",
          "askMatLUT": "false",
          "askTime": "false",
          "askOnceAllButField": "false",
          "needPropagatorD":  "false"`

needs to be added here: https://github.com/AliceO2Group/QualityControl/blob/master/Modules/MFT/mft-clusters.json

Modules/ITS/src/ITSClusterTask.cxx
```diff
@@ -153,6 +138,9 @@ void ITSClusterTask::startOfCycle()
 void ITSClusterTask::monitorData(o2::framework::ProcessingContext& ctx)
 {
 
+  mGeom = o2::its::GeometryTGeo::Instance();
+  mGeom->fillMatrixCache(o2::math_utils::bit2Mask(o2::math_utils::TransformType::L2G));
+
```
iravasen, line 143: is there a reason to move these two lines in `monitorData` instead of `initialize` (where they were before)? The same for the ITSFhrTask.cxx

iravasen:
@chiarazampolli thanks! I only have the comment above, the rest is fine! I will take care of modifying the ITS jsons at P2 once this will be there.

Modules/ITS/src/ITSClusterTask.cxx
Barthelemy: good point, that does not seem optimal

chiarazampolli: The geometry is passed only when the task calls "run", which for the QC is the monitorData. Unfortunately, I don't see a better way. You can protect this: if already done, don't redo it. Assuming the geometry does not change. 
I would let you take care once the PR is merged.

chiarazampolli:
> Hi @chiarazampolli
> 
> I will try to do a local test but I need to rebuild first as I have old versions of the repositories.
> 
> Just from looking at the code it seems to me that the Geometry Request in the json i.e. ` "grpGeomRequest" : { "geomRequest": "Aligned", "askGRPECS": "false", "askGRPLHCIF": "false", "askGRPMagField": "true", "askMatLUT": "false", "askTime": "false", "askOnceAllButField": "false", "needPropagatorD": "false"`
> 
> needs to be added here: https://github.com/AliceO2Group/QualityControl/blob/master/Modules/MFT/mft-clusters.json

Ciao @tomas-herman ,
Added. 
Chiara

Modules/ITS/src/ITSClusterTask.cxx
shahor02: Note that with the DPL CCDB fetcher the geometry will not be loaded until the 1st TF processing. And the methods above to need TGeometry loaded. 
As far as I understand, this means that they should be executed in the monitor. 
Another question is that they should be executed only once in the beginning of the run, i.e. should be changed to 
```
if (pc.services().get<o2::framework::TimingInfo>().globalRunNumberChanged) {
   mGeom = o2::its::GeometryTGeo::Instance();
   mGeom->fillMatrixCache(o2::math_utils::bit2Mask(o2::math_utils::TransformType::L2G));
}
```

chiarazampolli: Hello @shahor02 , 
Exactly, this is what I meant. I did not know about the way to check that the run number is changed, I would have checked with a `bool` :-)
Implemented now, please check.
For the B field, it is different though, since for that the DP processing is continuously ongoing. 
Chiara

iravasen: Thanks a lot @shahor02, @chiarazampolli.

iravasen: @chiarazampolli can you please apply the same also for the ITSClusterTask.cxx?

chiarazampolli: I did actually, but I forgot to add it :-) Now done.

iravasen (approved):

shahor02 (approved):

njacazio (approved):
Thanks Chiara!

Barthelemy:
@tomas-herman do you sign off the changes ?

tomas-herman (approved):

Barthelemy:
merging as people have approved and only 1 check is missing

PR #1693, opened 2023-03-28
TPC: Update QualityObserver

mlesch:
Update includes: 
- New option in json called `qualityDetailChoice`. This allows to select for which qualities the reasons and comments of the check result are shown. Example `"qualityDetailChoice": "Medium, Bad"` will only show reasons and comments for checks with medium or bad quality. Full list of options: `"qualityDetailChoice": "Null, Good, Medium, Bad"`
- Fix to (hopefully) solve the EOR crash. `finalize` function should not delete pointer to canvas

### DIFF
--- a/Modules/TPC/src/QualityObserver.cxx
+++ b/Modules/TPC/src/QualityObserver.cxx
@@ -35,6 +35,7 @@ void QualityObserver::configure(const boost::property_tree::ptree& config)
   auto& id = getID();
   mObserverName = config.get<std::string>("qc.postprocessing." + id + ".qualityObserverName");
   mViewDetails = config.get<bool>("qc.postprocessing." + id + ".observeDetails", true);
+  mQualityDetailChoice = config.get<std::string>("qc.postprocessing." + id + ".qualityDetailChoice", "Null, Good, Medium, Bad");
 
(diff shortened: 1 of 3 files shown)


wiechula (approved):

Modules/TPC/src/QualityObserver.cxx
```diff
@@ -90,8 +109,6 @@ void QualityObserver::update(Trigger t, framework::ServiceRegistryRef services)
 void QualityObserver::finalize(Trigger t, framework::ServiceRegistryRef)
 {
   generatePanel();
-  delete mCanvas;
-  mCanvas = nullptr;
```
knopers8, line 94: This is indeed correct to be removed, but please delete `mCanvas` in `~QualityObserver` instead.

mlesch: Putting it into  `~QualityObserver` somehow then still had some issues for me. That's why I just removed this lines and treated `mCanvas` similar to `mPlots` in the `Framework/TrendingTask`

knopers8: I see, thanks for the explanation. To be understood.

PR #1695, opened 2023-03-28
TPC: Update RatioGenerator

mlesch:
Update to fix EOR crash.

### DIFF
--- a/Modules/TPC/src/RatioGeneratorTPC.cxx
+++ b/Modules/TPC/src/RatioGeneratorTPC.cxx
@@ -64,14 +64,12 @@ void RatioGeneratorTPC::update(Trigger t, framework::ServiceRegistryRef services
 
 void RatioGeneratorTPC::finalize(Trigger t, framework::ServiceRegistryRef)
 {
-  generatePlots();
   for (const auto& source : mConfig) {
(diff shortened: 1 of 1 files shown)


wiechula (approved):

Modules/TPC/src/RatioGeneratorTPC.cxx
```diff
@@ -64,14 +64,12 @@ void RatioGeneratorTPC::update(Trigger t, framework::ServiceRegistryRef services
 
 void RatioGeneratorTPC::finalize(Trigger t, framework::ServiceRegistryRef)
 {
-  generatePlots();
   for (const auto& source : mConfig) {
     if (mRatios.count(source.nameOutputObject)) {
       getObjectsManager()->stopPublishing(source.nameOutputObject);
-      delete mRatios[source.nameOutputObject];
-      mRatios[source.nameOutputObject] = nullptr;
```
knopers8, line 72: Same here, please delete pointers in `~RatioGeneratorTPC` instead.

mlesch: Similar comment here. In the destructor it gave me issues when i stopped the task.

PR #1698, opened 2023-03-29
Fix for trending rate and one more raw qc histo

noferini:


njacazio (approved):
Thanks a lot Francesco!

PR #1701, opened 2023-03-30
Do not check empty MOs, sanitize testCheck, compute aggregated Activity for QO

knopers8:
I added a protection to avoid checking MOs which are nullptr, which testCheck previously abused for no good reason.

Also, most of the tests in testCheck did not have any sense, they tested a dummy class that was written in testCheck. I replaced them with something more meaningful.

The main point of this commit is that now Check computes the aggregated activity, though it is not used yet to store objects, this will come later.

### DIFF
--- a/Framework/src/RootClassFactory.cxx
+++ b/Framework/src/RootClassFactory.cxx
@@ -31,7 +31,7 @@ void loadLibrary(const std::string& moduleName)
   ILOG(Info, Devel) << "Loading library " << library << ENDM;
   int libLoaded = gSystem->Load(library.c_str(), "", true);
   if (libLoaded < 0) {
-    BOOST_THROW_EXCEPTION(FatalException() << errinfo_details("Failed to load Detector Publisher Library"));
+    BOOST_THROW_EXCEPTION(FatalException() << errinfo_details("Failed to load the library " + library));
   }
 }
 
--- a/Framework/include/QualityControl/Check.h
+++ b/Framework/include/QualityControl/Check.h
@@ -73,9 +73,6 @@ class Check
   //TODO: Unique Input string
   static o2::header::DataDescription createCheckDataDescription(const std::string& checkName);
 
-  // For testing purpose
-  void setCheckInterface(CheckInterface* checkInterface) { mCheckInterface = checkInterface; };
-
   UpdatePolicyType getUpdatePolicyType() const;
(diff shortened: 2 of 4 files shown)


Framework/src/RootClassFactory.cxx
```diff
@@ -31,7 +31,7 @@ void loadLibrary(const std::string& moduleName)
   ILOG(Info, Devel) << "Loading library " << library << ENDM;
   int libLoaded = gSystem->Load(library.c_str(), "", true);
   if (libLoaded < 0) {
-    BOOST_THROW_EXCEPTION(FatalException() << errinfo_details("Failed to load Detector Publisher Library"));
+    BOOST_THROW_EXCEPTION(FatalException() << errinfo_details("Failed to load the library " + library));
```
Barthelemy, line 34: this is a message that dates back to the early days of AMORE...

Barthelemy (approved):
all good

Barthelemy:
@ktf we are getting and error in the full system test : 
```
--
Detected critical problem in logfile reco_NOGPU.log
reco_NOGPU.log-[17034:qc-task-EMC-CellTask]: Info in <TJAlienFile::Open>: Accessing file /alice/data/CCDB/EMC/Calib/TimeCalibParams/11/21022/9ec94200-6ed9-11ed-803c-0aa200331b9a in SE <ALICE::CERN::OCDB>
reco_NOGPU.log-[17034:qc-task-EMC-CellTask]: [07:28:24][INFO] ccdb reads http://alice-ccdb.cern.ch/EMC/Calib/TimeCalibParams for 1680506905593 (retrieve, agent_id: alibuild01.cern.ch-1680506035-98UCMx), 
reco_NOGPU.log:[ERROR] Workflow crashed - pid 16953 (gpu-reconstruction) was killed abnormally with exit code 128, could be out of memory killer, segfault, unhandled exception, SIGKILL, etc...
reco_NOGPU.log-[INFO] QUIT_REQUESTED
reco_NOGPU.log-[16954:its-stf-decoder]: [07:28:51][INFO] Received device shutdown request (signal 15).
```
any idea whom should be contacted ?

Framework/src/RootClassFactory.cxx
knopers8: indeed that was my feeling :D

PR #1704, opened 2023-03-31
TOF: Fix checker of hit multiplicity

ercolessi:
@njacazio


njacazio (approved):
Thanks Francesca, very good

PR #1706, opened 2023-04-03
askOnceAllButField changed to true in the GeomHelper

chiarazampolli:
@shahor02 , @davidrohr , @knopers8


knopers8 (approved):
Should it be the default in the framework?

iravasen:
thanks @chiarazampolli, related question: should be used = true at P2?

chiarazampolli:
> > thanks @chiarazampolli, related question: should be used = true at P2?
> 
> yes

For GLO at P2, I already put "true".
Chiara

PR #1709, opened 2023-04-04
TOF: Fix possible reason for crashing Trending Tasks

ercolessi:
@noferini


ercolessi:
@Barthelemy it would be nice to have this PR included in the next software upgrade at P2 (next week?), do you think it is feasible?

Barthelemy:
@ercolessi yes, no problem

PR #1710, opened 2023-04-04
MCH: various fixes to post-processing and checks

aferrero2707:
- fixed retrieval of QualityObjects from CCDB
- fixed display of QualityObjects in summary canvas when objects are not updated
- fixed bug in check of "good channels ratio" plot when no reference is available
- removed verbose infologger messages
- made plots and reductors optional in post-processing configuration
- removed PaveText panel with check result on individual plots

### DIFF
--- a/Modules/MUON/MCH/include/MCH/Helpers.h
+++ b/Modules/MUON/MCH/include/MCH/Helpers.h
@@ -137,6 +137,26 @@ struct CcdbObjectHelper {
 
 //_________________________________________________________________________________________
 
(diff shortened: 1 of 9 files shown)


aferrero2707:
@Barthelemy @knopers8 @aphecetche this is a first bugfix PR for the new MCH QC. It also removes some verbose Infologger output that was reported by Barth. Any chance to have it in the next FLP suite?

Barthelemy:
@aferrero2707 yes, I am releasing now

PR #1712, opened 2023-04-06
[EMCAL-831, EMCAL-524, EMCAL-530] Exclude LEDMON from QC

mfasDa:
- Cell Task: LEDMONs are currently excluded,
  dedicated histograms might be added for calib
  events
- Cluster Task: LEDMONs must be included entirely,
  they are arranged in strip modules, they are not
  clusterizable.


mfasDa:
Triggered by https://github.com/AliceO2Group/AliceO2/pull/11073

PR #1713, opened 2023-04-06
[EMCAL-530] Fixes in the Cluster task

mfasDa:
- Use new container setter to automatically
  initialize the lookup table and make sure the event
  is complete
- Iterate over event range instead of timeframe range
  in rejection of LEDMONs


mfasDa:
CI error unrelated

PR #1715, opened 2023-04-08
 Data volume per sector

vikasLstar:
This PR is made for adding an object named "data volume per sector" to the list of objects the task already has.


knopers8:
Please fix the formatting as indicated by the `Clang format` check

bazinski:
I suggest, lose the stats, and was there not other stuff along the top? I seem to remember some dots along the top for the  plot you are replicated.

And then of course as @knopers8 says, correct formatting.

knopers8 (approved):

PR #1716, opened 2023-04-10
FIT PostProc: bugfix for incorrect types in dynamic cast

afurs:


afurs:
@Barthelemy please add this PR into next QC version

afurs:
Unrelated CICD issue

Modules/FIT/FDD/src/PostProcTask.cxx
```diff
@@ -306,13 +306,13 @@ void PostProcTask::update(Trigger t, framework::ServiceRegistryRef)
   }
 
   auto mo3 = mDatabase->retrieveMO(mPathDigitQcTask, "AmpPerChannel", t.timestamp, t.activity);
-  auto hAmpPerChannel = mo3 ? dynamic_cast<TH2D*>(mo3->getObject()) : nullptr;
+  auto hAmpPerChannel = mo3 ? dynamic_cast<TH2F*>(mo3->getObject()) : nullptr;
```
knopers8, line 309: Why not `TH2`, so it works for both?

afurs: I think better to cast into exact type of this object.

knopers8 (approved):

Modules/FIT/FDD/src/PostProcTask.cxx
knopers8: I hope we do not get to a case where QC objects are produced with a QC version with `TH2D`, but the checks are applied with a newer QC version...

PR #1719, opened 2023-04-11
ITS Tracks online: new MOs resetting each cycle

IsakovAD:
Added two new MO resetting each cycle, which will be used in PP trends:
1) Number of tracks per event
2) Number of clusters used to reconstruct track

### DIFF
--- a/Modules/ITS/include/ITS/ITSTrackTask.h
+++ b/Modules/ITS/include/ITS/ITSTrackTask.h
@@ -62,6 +62,7 @@ class ITSTrackTask : public TaskInterface
 
   std::vector<TObject*> mPublishedObjects;
   TH1D* hNClusters;
+  TH1D* hNClustersReset;
   TH1D* hTrackEta;
   TH1D* hTrackPhi;
   TH1D* hVerticesRof;
@@ -72,6 +73,7 @@ class ITSTrackTask : public TaskInterface
   TH1D* hVertexContributors;
   TH1D* hAssociatedClusterFraction;
   TH1D* hNtracks;
+  TH1D* hNtracksReset;
   TH2D* hNClustersPerTrackEta;
   TH2D* hClusterVsBunchCrossing;
   TH2D* hNClusterVsChipITS;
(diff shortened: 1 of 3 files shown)


iravasen:
@IsakovAD thanks a lot to start this. I have some comments:

1. You are resetting the two histo also in the `reset()` method of the task. At P2 we run with `delta` mode which means that the 2 plots are reset automatically at every cycle (i.e. the `reset` method is called after every cycle). So in this case we have to keep the `Reset` of the histos only in 1 place (monitorData or reset). But there is another thing (see point 2). 
2. In `delta` mode, the merger node will continue to sum up all the objects received so the 2 histos in the end will have the entries for the whole run. To avoid this there is the concept of the moving window (https://github.com/AliceO2Group/QualityControl/blob/master/doc/Advanced.md#moving-window) but if we apply that, the merger will reset EVERY plot we have in the task (i.e. specified in the task `reset` method). 

So my proposal would be:
- run the task in `entire` mode where the plots are NOT reset. Here you have to keep the `histo->Reset` in monitorData to be sure to reset the 2 histos. Then the merger will consider only the histos coming from the task (which will be reset at every cycle in monitorData) and the post-processing will do its job on this object which by definition contains entries only from the last cycle (last n ROFs). So here you do not need to do anything except removing the Reset of the two histos in the reset method of the task. 

On top of this we need also:
- to add the two plots in the post-processing task of the tracks. 
- to add also the monitoring of the TOTAL number of clusters (i.e. attached + not attached to tracks). This can be done in the cluster task if you like (I don't have a strong opinion here).

iravasen:
to integrate, if you like:
- in `delta` mode the merger keeps a local copy of every plot and sum to it every new plot which come in the next cycles (this new entries are reset at every cycle in the single nodes).
- in `entire` mode, the task keeps "in memory" the full object and the merger merges only what received from the nodes every time. 

At the end things are similar but in some case (like ours) we have some differences.

IsakovAD:
@iravasen 

I have deleted 2 new plots from reset() function, so now we can use the module in ENTIRE mode

Regarding the PP of the Tracks, Maria is developing this update already

iravasen:
@IsakovAD thanks a lot! It's fine with me. Approving

iravasen (approved):

iravasen:
(remember to apply the clang)

IsakovAD:
Changes in RawTask are clang formating

PR #1721, opened 2023-04-12
[FOCAL-8] Use source ID to distinguish FOCAL-E subsystems

mfasDa:
Endpoint unreliable, was mainly a temporary solution
for the 2022 testbeam. Using the source ID is more
reliable as it is set by the firmwares and will also work
after recabling


mfasDa:
Same as https://github.com/AliceO2Group/AliceO2/pull/11132

knopers8 (approved):

PR #1730, opened 2023-04-18
[FOCAL-8] Use FEE ID to distinguish the subsystems

mfasDa:
- Pads: Only 1 FEE, 0xcafe
- Pixels: all other FEEs

### DIFF
--- a/Modules/FOCAL/src/TestbeamRawTask.cxx
+++ b/Modules/FOCAL/src/TestbeamRawTask.cxx
@@ -15,6 +15,7 @@
 ///
 
 #include <algorithm>
+#include <iomanip>
 #include <iostream>
 #include <unordered_set>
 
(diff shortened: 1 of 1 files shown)


mfasDa:
Replaces #1721

Same explanation as in https://github.com/AliceO2Group/AliceO2/pull/11185: The STF builder complained when running with the source IDs which were currently in use. Both systems must run with the DAQID of FOCAL (39) in order to be compatible with the STF builder, which then can no longer be used to distinguish pads and pixels. The identification of the subsystem must be fully based on the FEE ID. In order to be compatible for the moment we use to identify Pad data the FEE ID which is currently in use (0xcafe - only 1 FEE installed) and conclude that all other FEEs are pixel FEEs. As a long term solution we will implement a mapper that maps FEE IDs to subsystems and position (layer, module, index in module).

Barthelemy (approved):

Barthelemy:
failures unrelated, merging;.

PR #1733, opened 2023-04-19
FOCAL / fix for mac

Barthelemy:


Barthelemy:
for some reason there is an error on mac that seems to come from infologger. This works around the problem.

knopers8 (approved):

Barthelemy:
force merging because it breaks other PR's

PR #1734, opened 2023-04-19
TRD tracked matched pulseheight fixup

bazinski:
Pulseheight spectrum that is matched to tracks does not account for when the pileup info is not there, and time is 0.

Additional plots added, for expert for diagnostics.

### DIFF
--- a/Modules/TRD/include/TRD/PulseHeightTrackMatch.h
+++ b/Modules/TRD/include/TRD/PulseHeightTrackMatch.h
@@ -69,6 +69,9 @@ class PulseHeightTrackMatch final : public TaskInterface
   std::shared_ptr<TH1F> mParsingTimePerTF;
   std::shared_ptr<TH1F> mTracksPerEvent;
   std::shared_ptr<TH1F> mTrackletsPerMatchedTrack;
+  std::shared_ptr<TH1F> mDebugCounter;
+  std::shared_ptr<TH1F> mTriggerMatches;
+  std::shared_ptr<TH1F> mTriggerRecordsMatchesDigits;
   std::shared_ptr<TProfile> mPulseHeightpro = nullptr;
   std::shared_ptr<TProfile2D> mPulseHeightperchamber = nullptr;
   // information pulled from ccdb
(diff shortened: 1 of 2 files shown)


Modules/TRD/src/PulseHeightTrackMatch.cxx
```diff
@@ -108,20 +117,29 @@ void PulseHeightTrackMatch::monitorData(o2::framework::ProcessingContext& ctx)
 
   auto crustart = std::chrono::high_resolution_clock::now();
 
+  mDebugCounter->Fill(0);
+  mTriggerMatches->Fill(triggerrecords.size());
+  int loopcount = 0;
```
knopers8, line 122: unused variable

Modules/TRD/include/TRD/PulseHeightTrackMatch.h
```diff
@@ -69,6 +69,9 @@ class PulseHeightTrackMatch final : public TaskInterface
   std::shared_ptr<TH1F> mParsingTimePerTF;
   std::shared_ptr<TH1F> mTracksPerEvent;
   std::shared_ptr<TH1F> mTrackletsPerMatchedTrack;
+  std::shared_ptr<TH1F> mDebugCounter;
```
knopers8, line 72: That's unusual, but if it works, then sure

knopers8 (approved):

Modules/TRD/include/TRD/PulseHeightTrackMatch.h
bazinski: I just want to be able to see live when things go wrong, as with the data I have been looking at the updates to the tasks main histogram is incredibly slow. It seemed like the easiest way of seeing where it is rejecting the data. This will ultimately come out of course, later on.

Barthelemy:
@knopers8 What do you think of the error in build/QualityControl/o2-dataflow-cs8  ?

bazinski:
@Barthelemy that does not seem related to the code in this pr, is a rebase needed?

Barthelemy:
no, I don't think so but I wanted to opinion of Piotr.

knopers8:
Unrelated.

It seems as if the CI machine was overloaded.

PR #1739, opened 2023-04-21
ITS FEE: Summary lane status plots changed to fraction 

IsakovAD:
Lane status plots was changed from Number of Lanes -> Fraction of Lanes. 

Corresponding Check and Post-processing tasks were changed as well


iravasen:
@IsakovAD thanks a lot! Only one comment:
- what do you think if we show the fraction directly in %?

IsakovAD:
@iravasen 

It could be done, but since our other shifter MOs show information in fractions, so I decided to keep it constant across all plots.

iravasen (approved):

PR #1740, opened 2023-04-21
TPC: Update CheckOfSlices

mlesch:
Update:
-  Takes uncertainties of single data points into account for calculation of mean and checks
- Allows for Range and Zero Check
- Saving MetaData for QualityObserver

### DIFF
--- a/Modules/TPC/include/TPC/CheckOfSlices.h
+++ b/Modules/TPC/include/TPC/CheckOfSlices.h
@@ -45,7 +45,7 @@ class CheckOfSlices : public o2::quality_control::checker::CheckInterface
 
  private:
   ClassDefOverride(CheckOfSlices, 3);
-  std::string createMetaData(std::vector<std::string>);
+  std::string createMetaData(const std::vector<std::string>& pointMetaData);
   std::string mCheckChoice;
   double mExpectedPhysicsValue;
   double mNSigmaExpectedPhysicsValue;
--- a/Modules/TPC/include/TPC/CheckOfSlices.h
+++ b/Modules/TPC/include/TPC/CheckOfSlices.h
@@ -44,25 +44,29 @@ class CheckOfSlices : public o2::quality_control::checker::CheckInterface
   std::string getAcceptedType() override { return "TCanvas"; }
 
  private:
-  ClassDefOverride(CheckOfSlices, 2);
+  ClassDefOverride(CheckOfSlices, 3);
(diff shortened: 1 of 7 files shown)


Modules/TPC/include/TPC/CheckOfSlices.h
```diff
@@ -44,17 +44,33 @@ class CheckOfSlices : public o2::quality_control::checker::CheckInterface
   std::string getAcceptedType() override { return "TCanvas"; }
 
  private:
-  ClassDefOverride(CheckOfSlices, 2);
+  ClassDefOverride(CheckOfSlices, 3);
+  void calculateStatistics(const double* yValues, const double* yErrors, bool useErrors, const int firstPoint, const int lastPoint, double& mean, double& stddevOfMean);
```
wiechula, line 48: I think this is copy'n'paste of `CheckOfTrendings::calculateStatistics`
Please put it in one common helper file e.g. `include/TPC/Utility.h`

Modules/TPC/src/CheckOfSlices.cxx
```diff
@@ -110,184 +119,438 @@ Quality CheckOfSlices::check(std::map<std::string, std::shared_ptr<MonitorObject
...
+    checkMessage = mMediumString;
     msg->SetFillColor(kOrange);
   } else if (checkResult == Quality::Null) {
     h->SetFillColor(0);
+    msg->AddText("Quality::Null. Failed checks:");
+    checkMessage = mNullString;
+  }
+
+  // Split lines by hand as \n does not work with TPaveText
+  std::string delimiter = "\n";
```
wiechula, line 322:
```suggestion
  const std::string delimiter = "\n";
```

wiechula (changes requested):

Modules/TPC/include/TPC/CheckOfSlices.h
mlesch: I moved `calculateStatistics` to` include/TPC/Utility.h` and removed it from both `CheckOfTrendings` and `CheckOfSlices`

Modules/TPC/src/CheckOfSlices.cxx
mlesch: `Done, I also did a similar change now in CheckOfTrendings`

wiechula (changes requested):

wiechula (approved):

PR #1741, opened 2023-04-24
[FOCAL-8, FOCAL-13] Apply bad channel map for pads

mfasDa:
Apply bad channel map for pad channels (no
CMN or calib channels)


mfasDa:
Adding also filtering of 0xDEADBEEF messages to this pull request.

Barthelemy (approved):

PR #1744, opened 2023-04-25
[FOCAL-12] QC task for Pad pedestal calibration

mfasDa:
Monitor extracted pedestal values as function of
channel ID and position in stack.

### DIFF
--- a/Modules/FOCAL/src/PedestalCalibTask.cxx
+++ b/Modules/FOCAL/src/PedestalCalibTask.cxx
@@ -1,0 +1,135 @@
+// Copyright 2019-2020 CERN and copyright holders of ALICE O2.
+// See https://alice-o2.web.cern.ch/copyright for details of the copyright holders.
+// All rights not expressly granted are reserved.
+//
+// This software is distributed under the terms of the GNU General Public
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+///
+/// \file   PedestalCalibTask.cxx
+/// \author My Name
+///
+
+#include <TCanvas.h>
+#include <TH1.h>
+#include <TH2.h>
+
+#include "FOCALCalib/PadPedestal.h"
+
+#include "QualityControl/QcInfoLogger.h"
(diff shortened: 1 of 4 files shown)


Barthelemy:
@TimoWilken @ktf it seems that the macos build is stuck

Modules/FOCAL/src/PedestalCalibTask.cxx
```diff
@@ -0,0 +1,135 @@
...
+  ILOG(Debug, Support) << "Start monitoring data" << ENDM;
+  bool hasPedestalsCCDB = false;
+  for (auto&& input : o2::framework::InputRecordWalker(ctx.inputs())) {
+    // get message header
+    if (input.header != nullptr && input.payload != nullptr) {
+      auto payloadSize = o2::framework::DataRefUtils::getPayloadSize(input);
+      const auto* header = o2::framework::DataRefUtils::getHeader<header::DataHeader*>(input);
+      if (payloadSize) {
+        if ((strcmp(header->dataOrigin.str, "CLP") == 0) && (strcmp(header->dataDescription.str, "FOC_PADPEDESTALSCLP") == 0)) {
+          ILOG(Info, Support) << "Pedestals for Pads found" << ENDM;
```
Barthelemy, line 84: I am always fearful of logging within `monitorData`. It can easily create floods of logs as this method is called at high frequency. Perhaps you could move it to `Debug` or make it so that it does not log more than X per minute: 
```
  static AliceO2::InfoLogger::InfoLogger::AutoMuteToken msgLimit(LogInfoSupport, 1, 600); // send it only every 10 minutes
    ILOG_INST.log(msgLimit, "blabla");
```

Barthelemy (changes requested):

Barthelemy (approved):

Modules/FOCAL/src/PedestalCalibTask.cxx
mfasDa: I can move later the log level to debug but as this is a QC component only attached to a pedestal run, which monitors the object stored in the CCDB, the monitorData will be called only once, when the CCDB object is published. So there will be no message flood. For the time being while we are still developing the pedestal run I would like to keep the logging at level info. Once the development is done I will decrease the level to debug.

mfasDa:
@TimoWilken looks like the macOS builder is dead?

PR #1747, opened 2023-04-25
FT0 PostProc: turning off PL option for TGraph

afurs:


afurs:
Trivial, no need in CICD

PR #1752, opened 2023-05-01
Changing the NTracks_mean and NClusters_mean to show statictics per c…

selindra:
…ycle

### DIFF
--- a/Modules/ITS/itsQCTrendingTracks.json
+++ b/Modules/ITS/itsQCTrendingTracks.json
@@ -31,13 +31,13 @@
         "dataSources": [
           {
             "type": "repository",
-            "paths": ["ITS/MO/ITSTrackTask/NClusters",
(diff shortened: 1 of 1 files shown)


Barthelemy (approved):

IsakovAD:
@iravasen 

With this commit, PP for Track will trend only reset histograms. Changes were tested, so I propose merging.

I think that we don't need to keep old non-resetting trends in the PP task. 

One more thing, we need to update P2 consul accordingly. Unfortunately, I can't have access to it for a moment.

iravasen:
Thanks! I approve. I will take care of modifying consul at P2 once this PR will be there. Note that it might be in two weeks since P2 updates will not happen frequently. BUT we can test on staging since upgrade here will happen on a weekly base as usual.

iravasen (approved):

PR #1754, opened 2023-05-03
ZDC RAW: New data Format

cpuggion84:
Change decoder data for the new data format.

### DIFF
--- a/Modules/ZDC/src/ZDCRawDataTask.cxx
+++ b/Modules/ZDC/src/ZDCRawDataTask.cxx
@@ -81,15 +81,49 @@ void ZDCRawDataTask::startOfCycle()
 
 void ZDCRawDataTask::monitorData(o2::framework::ProcessingContext& ctx)
 {
+  o2::framework::DPLRawParser parser(ctx.inputs());
+  int PayloadPerGBTW = 10;
+  int dataFormat;
+  const uint32_t* gbtw;
+  uint64_t count = 0;
+  size_t payloadSize;
+  size_t offset;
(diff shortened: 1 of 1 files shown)


Modules/ZDC/src/ZDCRawDataTask.cxx
```diff
@@ -81,15 +81,49 @@ void ZDCRawDataTask::startOfCycle()
...
-    auto const* payload = it.data();
-    size_t payloadSize = it.size();
-    if (payload != nullptr) {
-      for (int32_t ip = 0; ip < (int32_t)payloadSize; ip += 16) {
-        processWord((const uint32_t*)&payload[ip]);
+    auto rdhPtr = reinterpret_cast<const o2::header::RDHAny*>(it.raw());
+    if (rdhPtr == nullptr || !o2::raw::RDHUtils::checkRDH(rdhPtr, true)) {
+      nErr[0]++;
+      if (nErr[0] < 5) {
+        LOG(warning) << "ZDCDataReaderDPLSpec::run - Missing RAWDataHeader on page " << count;
```
Barthelemy, line 98: Having logging in `monitorData` is very risky. This method is called many times per seconds and it could start flooding the logs. Could you put it debug or make sure it is not called more than X per minute ?

Barthelemy (changes requested):

Modules/ZDC/src/ZDCRawDataTask.cxx
cpuggion84: ok I removed the messages.

Barthelemy (approved):

PR #1765, opened 2023-05-12
[EMCAL-524] Drop loading calibration objects if on-the-fly calibration is not requested

mfasDa:
- Don't load calibration objects if on-the-fly calibration
  is switched off
- Include channel status statistics histograms in group
  of recalibration histograms


mfasDa:
For async QC: QC subscribing only to calibrated cells - no need to fill the calibration histograms. In this case we can also drop loading of the calibration objects in order to reduce the amount of grid file handles, as requested by the grid experts.

Barthelemy:
@mfasDa could you check the error in the full system check ? I am not sure if it is serious or not.

PR #1770, opened 2023-05-16
Use 1D representation of MID strip occupancy

dstocco:
The strip occupancy map provided directly a 2D representation which was meant to give a quick view of the actual position of the strips.
This representation is good for shifters, but it makes it difficult to pinpoint the ID of the problematic strip.
This PR introduces a 1D representation that can be more easily related to the strip ID for further analysis.
The 2D representation is kept for shifters, but, the conversion from 1D to 2D representation is entrusted to a separate class that can be used by different tasks which are handling MID digits.
The commits are meant to be atomic: please do not squash.


dstocco:
Notifying @ValerieRamillien of the changes.

Barthelemy (approved):

PR #1774, opened 2023-05-16
OufOfBcCheck for FEE modules

DawidSkora:
This PR adds new check - OutOfBunchCollFeeModulesCheck which will allow user to compare ratio of collisions out of the bunch to all collisions per FEE module. 
This check shall be run on new histogram created by PostProcessingTask - OutOfBunchColl_BCvsFeeModules


Modules/FIT/FV0/src/OutOfBunchCollFeeModulesCheck.cxx
```diff
@@ -0,0 +1,156 @@
...
+      std::vector<float> allCollPerFeeModule(mo->getMetadataMap().size() + 1, 0);
+      for (auto metainfo : mo->getMetadataMap()) {
+        int bin = 0;
+        float value = 0;
+        try {
+          bin = std::stoi(metainfo.first);
+          value = std::stof(metainfo.second);
+        } catch (const std::invalid_argument& e) {
+          continue;
+        }
```
afurs, line 88: I guess exception will break the run. Maybe better to send error message into InfoLogger?

DawidSkora: Exception shouldn't brake the run, only pass actual 'for' run. However there was a bug related to wrong metadata key. It was fixed by latest commit.

Barthelemy (approved):

PR #1776, opened 2023-05-17
Failure to retrieve MO and QO should produce a warning

knopers8:
Seeing how retrieveMO and retrieveQO are used by postprocessing tasks, it is sometimes fine if we do not manage to retrieve an object, thus we make it a warning.

cc @martenole


Barthelemy (approved):

martenole:
Thanks!

PR #1780, opened 2023-05-18
ZDC RAW: Alignment plot

cpuggion84:


Modules/ZDC/src/ZDCRawDataTask.cxx
```diff
@@ -453,6 +459,29 @@ void ZDCRawDataTask::initHisto()
   } else
     setBinHisto1D(26, -0.5, 25.5);
   addNewHisto("SUMMARYRATE", "hrateSummary", "Rate Summary (KHz)", "NONE", "LBC");
+
+  if (auto param = mCustomParameters.find("SUMMARY_ALIGN"); param != mCustomParameters.end()) {
+    ILOG(Debug, Devel) << "Custom parameter - SUMMARY_ALIGN: " << param->second << ENDM;
+    tokenString = tokenLine(param->second, ";");
+    setBinHisto2D(atoi(tokenString.at(0).c_str()), atof(tokenString.at(1).c_str()), atof(tokenString.at(2).c_str()), atoi(tokenString.at(3).c_str()), atof(tokenString.at(4).c_str()), atof(tokenString.at(5).c_str()));
+  } else
```
Barthelemy, line 467: we strongly recommend to use curvy braces even for single line instructions.

Modules/ZDC/include/ZDC/ZDCRawDataTask.h
```diff
@@ -56,6 +56,26 @@ class ZDCRawDataTask final : public TaskInterface
     TH2* histo;
     std::vector<std::string> condHisto;
```
Barthelemy, line 57: This class has become huge. I don't think that it is manageable, nor that it allows to really understand what is going on. I encourage you to think about some refactoring, in the future.

Barthelemy (changes requested):

Barthelemy:
The only change I request to merge it is to have curly braces around single line `if` `else`.

Modules/ZDC/include/ZDC/ZDCRawDataTask.h
cpuggion84: you're right, now I have to finish adding some plots for VDM commissioning. Then I'll try to see how to better organize the code. Thanks for the advice

Modules/ZDC/src/ZDCRawDataTask.cxx
cpuggion84: ok I change it

knopers8 (approved):

PR #1784, opened 2023-05-21
[QC-974] Introduce mergeable histogram ratios

aferrero2707:
This PR introduces two generic classes that allow to create and publish histogram ratios in QC tasks.
The classes derive from MergeInterface, such that histogram rations from multiple sources can be properly combined on the merger nodes.
One class introduces ratios of TH1D histograms, and the other of TH2F histograms.

The classes can be conveniently used to normalize plots to the number of processed TFs or orbits, or to compute efficiencies.

The `uniformScaling` constructor parameter controls whether the denominator is a single constant (practically implemented as a 1-bin histogram) or if each bin has its own denominator value (in which case the numerator and denominator histograms have the same number of bins).
The first option allows to save memory and is appropriate for global histogram normalization, while the second is more flexible and appropriate for efficiency-like histograms.

### DIFF
--- a/Modules/Common/include/Common/TH1Ratio.h
+++ b/Modules/Common/include/Common/TH1Ratio.h
@@ -62,8 +62,6 @@ class TH1Ratio : public T, public o2::mergers::MergeInterface
   Bool_t Add(const TH1* h1, const TH1* h2, Double_t c1 = 1, Double_t c2 = 1) override;
   Bool_t Add(const TH1* h1, Double_t c1 = 1) override;
   void SetBins(Int_t nx, Double_t xmin, Double_t xmax) override;
-  void SetBins(Int_t nx, Double_t xmin, Double_t xmax, Int_t ny, Double_t ymin, Double_t ymax) override;
-  void SetBins(Int_t nx, Double_t xmin, Double_t xmax, Int_t ny, Double_t ymin, Double_t ymax, Int_t nz, Double_t zmin, Double_t zmax) override;
 
  private:
   T* mHistoNum{ nullptr };
--- a/Modules/Common/include/Common/TH1Ratio.h
+++ b/Modules/Common/include/Common/TH1Ratio.h
@@ -1,0 +1,84 @@
+// Copyright 2019-2020 CERN and copyright holders of ALICE O2.
+// See https://alice-o2.web.cern.ch/copyright for details of the copyright holders.
+// All rights not expressly granted are reserved.
+//
+// This software is distributed under the terms of the GNU General Public
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+/// \file TH1Ratio.h
+/// \brief A generic TH1Ratio inheriting MergeInterface
+///
+/// \author Piotr Konopka, piotr.jan.konopka@cern.ch, Andrea Ferrero
+
+#ifndef QUALITYCONTROL_TH1RATIO_H
+#define QUALITYCONTROL_TH1RATIO_H
+
+#include "Mergers/MergeInterface.h"
+#include <TH1F.h>
+#include <TH1D.h>
+
+namespace o2::quality_control_modules::common
+{
+
+template <class T>
+class TH1Ratio : public T, public o2::mergers::MergeInterface
+{
+ public:
+  TH1Ratio() = default;
+  TH1Ratio(TH1Ratio const& copymerge);
+  TH1Ratio(const char* name, const char* title, int nbinsx, double xmin, double xmax, bool uniformScaling = false);
+  TH1Ratio(const char* name, const char* title, bool uniformScaling = false);
+
+  ~TH1Ratio();
+
+  void init();
+
+  void merge(MergeInterface* const other) override;
+
+  T* getNum() const
+  {
+    return mHistoNum;
+  }
+
+  T* getDen() const
+  {
+    return mHistoDen;
+  }
+
+  bool hasUniformScaling() const
+  {
+    return mUniformScaling;
+  }
+
+  void update();
+
+  // functions inherited from TH1x
+  void Reset(Option_t* option = "") override;
+  void Copy(TObject& obj) const override;
+  Bool_t Add(const TH1* h1, const TH1* h2, Double_t c1 = 1, Double_t c2 = 1) override;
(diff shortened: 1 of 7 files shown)


aferrero2707:
@knopers8 @Barthelemy this is a draft implementation of some generic histogram normalization/ratio classes we discussed with Piotr. I still have to write some documentation. Could you please have a look and let me know if the codeis ok for you?

The code is partly inspired from ROOT's [TEfficiency class](https://root.cern/root/html606/TEfficiency_8cxx_source.html).

Modules/Common/include/Common/TH1Ratio.h
```diff
@@ -0,0 +1,82 @@
...
+/// \file TH1Ratio.h
+/// \brief An example of a custom TH1Quotient inheriting MergeInterface
+///
+/// \author Piotr Konopka, piotr.jan.konopka@cern.ch, Sebastien Perrin, Andrea Ferrero
+
+#ifndef QUALITYCONTROL_TH1RATIO_H
+#define QUALITYCONTROL_TH1RATIO_H
+
+#include <sstream>
+#include <iostream>
```
knopers8, line 21:
```suggestion
```
These two do not seem to be needed in the header.

Modules/Common/include/Common/TH1Ratio.h
```diff
@@ -0,0 +1,82 @@
...
+#include <TObject.h>
+#include <TH1.h>
+#include <TList.h>
```
knopers8, line 24: This one neither
```suggestion
```

Modules/Common/include/Common/TH1Ratio.h
```diff
@@ -0,0 +1,82 @@
...
+#include "Mergers/MergeInterface.h"
+
+using namespace std;
```
knopers8, line 27:
```suggestion
```

Modules/Common/src/TH1Ratio.cxx
```diff
@@ -0,0 +1,225 @@
...
+{
+  ILOG(Debug, Devel) << "SetBins(Int_t nx, Double_t xmin, Double_t xmax, Int_t ny, Double_t ymin, Double_t ymax) not valid for MergeableTH2Ratio" << ENDM;
+}
+
+void TH1Ratio::SetBins(Int_t nx, Double_t xmin, Double_t xmax,
+                       Int_t ny, Double_t ymin, Double_t ymax,
+                       Int_t nz, Double_t zmin, Double_t zmax)
+{
+  ILOG(Debug, Devel) << "SetBins(Int_t nx, Double_t xmin, Double_t xmax, Int_t ny, Double_t ymin, Double_t ymax, Int_t nz, Double_t zmin, Double_t zmax) not valid for MergeableTH2Ratio" << ENDM;
+}
```
knopers8, line 223: I assume this comes from the ROOT's class. As far as I understand, we do not need these placeholder methods in our case, so perhaps you could remove them?

Modules/Common/src/TH2Ratio.cxx
```diff
@@ -0,0 +1,228 @@
...
+  getDen()->SetBins(nx, xmin, xmax, ny, ymin, ymax);
+  TH2F::SetBins(nx, xmin, xmax, ny, ymin, ymax);
+}
+
+void TH2Ratio::SetBins(Int_t nx, Double_t xmin, Double_t xmax,
+                       Int_t ny, Double_t ymin, Double_t ymax,
+                       Int_t nz, Double_t zmin, Double_t zmax)
+{
+  ILOG(Debug, Devel) << "SetBins(Int_t nx, Double_t xmin, Double_t xmax, Int_t ny, Double_t ymin, Double_t ymax, Int_t nz, Double_t zmin, Double_t zmax) not valid for TH2Ratio" << ENDM;
+}
```
knopers8, line 226:
```suggestion

void TH2Ratio::SetBins(Int_t nx, Double_t xmin, Double_t xmax,
                       Int_t ny, Double_t ymin, Double_t ymax)
{
  getNum()->SetBins(nx, xmin, xmax, ny, ymin, ymax);
  getDen()->SetBins(nx, xmin, xmax, ny, ymin, ymax);
  TH2F::SetBins(nx, xmin, xmax, ny, ymin, ymax);
}
```

same here (and header files)

Modules/Common/src/TH2Ratio.cxx
```diff
@@ -0,0 +1,228 @@
...
+/// \brief An example of a custom TH2Quotient inheriting MergeInterface
+///
+/// \author Piotr Konopka, piotr.jan.konopka@cern.ch, Sebastien Perrin, Andrea Ferrero
+
+#include "Common/TH2Ratio.h"
+#include "QualityControl/QcInfoLogger.h"
+#include <TList.h>
+
+#include <fmt/format.h>
+#include <limits>
```
knopers8, line 22:
```suggestion
```
I don't see them used, could you please remove them to reduce the compilation time?

Modules/Common/src/TH1Ratio.cxx
```diff
@@ -0,0 +1,225 @@
...
+/// \file TH1Ratio.cxx
+/// \brief An example of a custom TH2Quotient inheriting MergeInterface
+///
+/// \author Piotr Konopka, piotr.jan.konopka@cern.ch, Sebastien Perrin, Andrea Ferrero
+
+#include "Common/TH1Ratio.h"
+#include "QualityControl/QcInfoLogger.h"
+// #include <fmt/format.h>
+
+using namespace std;
```
knopers8, line 21:
```suggestion
```
Probably not needed.

knopers8 (changes requested):
Thank you! The code looks reasonable, I have only "editorial" comments.

Since this class seems easy to test, would you consider adding a test in `Common/test`?

Modules/Common/include/Common/TH1Ratio.h
```diff
@@ -0,0 +1,82 @@
...
+namespace o2::quality_control_modules::common
+{
+
+class TH1Ratio : public TH1D, public o2::mergers::MergeInterface
```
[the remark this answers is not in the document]
aferrero2707: I was also thinking about using a template, but was not sure you would like the idea. Since we seem to agree on that, I will give it a try.

aferrero2707:
@knopers8 I re-implemented the two classes as templates, and added some unit tests. For the templates, I have decided to put the method implementation into a separate file with respect to the class declarations, for better readability. Let me know if that's ok for you.

knopers8 (approved):
Great, thank you!

PR #1787, opened 2023-05-22
[QC-972] Notify if aggregator name is more than 16 characters. 

Barthelemy:
Raise an error if two devices share the 16 first characters as it will clash. Provide the env var QUALITYCONTROL_NO_FATAL_16_CHARS to carry on despite the error.

### DIFF
--- a/Framework/src/InfrastructureSpecReader.cxx
+++ b/Framework/src/InfrastructureSpecReader.cxx
@@ -91,6 +91,7 @@ TaskSpec InfrastructureSpecReader::readSpecEntry<TaskSpec>(const std::string& ta
   TaskSpec ts;
 
   ts.taskName = taskTree.get<std::string>("taskName", taskID);
+  validateName(ts.taskName); // bail if the 16 first characters of the name match another device
   ts.className = taskTree.get<std::string>("className");
   ts.moduleName = taskTree.get<std::string>("moduleName");
   ts.detectorName = taskTree.get<std::string>("detectorName");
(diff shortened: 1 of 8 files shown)


Framework/src/InfrastructureSpecReader.cxx
```diff
@@ -91,6 +92,7 @@ TaskSpec InfrastructureSpecReader::readSpecEntry<TaskSpec>(const std::string& ta
   TaskSpec ts;
 
   ts.taskName = taskTree.get<std::string>("taskName", taskID);
+  validateName(ts.taskName); // bail if the 16 first characters of the name match another device
```
knopers8, line 95: But you are using `taskName`, `checkName`, etc instead of deviceName, but a check called `Tracks` will not be in a device named the same as a QC task (e.g. `qc-check-TST-Tracks` vs `qc-task-TST-Tracks`).

You are also preventing us from having tasks with the same name from two different detectors.

What would make more sense is to check if there are any duplicate output specs in the workflow.

Framework/src/InfrastructureSpecReader.cxx
```diff
@@ -402,4 +407,19 @@ T InfrastructureSpecReader::readSpecEntry(const std::string&, const boost::prope
   throw std::runtime_error("Unknown entry type: " + std::string(typeid(T).name()));
 }
 
+void InfrastructureSpecReader::validateName(const std::string& name)
+{
+  std::string const trunc = name.substr(0, 16);
+  auto test = sDeviceNamesCache.emplace(trunc);
+  if (!test.second) { // the second of the pair indicates if there was an insertion, i.e. it was not already there
+    std::string errorMessage = std::string("Multiple devices share the same first 16 characters of their names: ") + trunc + "[" + name.substr(16) + "]. Probable cause: name was longer than 16 characters and was truncated. Disable one of the faulty tasks/aggregator/postprocessing or call the detector expert to change the name.";
```
knopers8, line 415: Again, devices do not have the 16-character limitation, `DataDescription` has. This is very misleading.

knopers8 (changes requested):

PR #1796, opened 2023-05-25
[QC-429] NewObject trigger recognizes new objects whose validity ends before current timestamp

knopers8:


knopers8:
@tklemenz this should solve your case as well.

tklemenz:
I cannot fully test because no new IDCs are being uploaded since yesterday morning and I cannot upload data to the test-ccdb manually for some reason. But when I run the workflow the warning about the object that the trigger is looking for doesn't show up anymore. So apparently it can find an object in the ccdb even though it is not valid at the given time. 
It seems to me that the problem is fixed. But I'll let you know once we know for sure. Thanks a lot!
Will this be in the next release?

knopers8:
>  Will this be in the next release?

Hopefully yes.

Barthelemy (approved):
all good and clear !

PR #1798, opened 2023-05-25
Adding t0-peak check and trend with other quality plots

deependra170598:


Modules/TRD/src/PulsePositionCheck.cxx
```diff
@@ -0,0 +1,155 @@
...
+
+      auto* h = dynamic_cast<TH1F*>(mo->getObject());
+	    
+      //check max bin is in the spike on left.
+      auto maxbin = h->GetMaximumBin();
+
+      if (maxbin < mPulseHeightPeakRegion.first || maxbin > mPulseHeightPeakRegion.second) {
+        // is the peak in the peak region.
+        result = Quality::Bad;
+        result.addReason(FlagReasonFactory::Invalid(),
```
knopers8, line 84: Here is the list of flags you can use:
https://github.com/AliceO2Group/AliceO2/blob/dev/DataFormats/QualityControl/etc/flagReasons.csv
These should normally explain what is the effect on data. If you do not know what is the effect on data, use `Unknown`, but feel free to propose new flags if you think something is missing.

`Invalid` is a special flag and  should not be used by detector code.

m-c-danisch:
Hi, I'm afraid that at the moment asking for
`if (maxbin < mPulseHeightPeakRegion.first || maxbin > mPulseHeightPeakRegion.second)`
will lead to BAD quality very often when the quality is acutally GOOD. (Local maximum in the correct region, but not global maximum)
Let's hope that with a better tuned pile-up rejection, developed in https://github.com/AliceO2Group/QualityControl/pull/1802, the way the automatic check is implemented here, will work

bazinski:
> Hi, I'm afraid that at the moment asking for `if (maxbin < mPulseHeightPeakRegion.first || maxbin > mPulseHeightPeakRegion.second)` will lead to BAD quality very often when the quality is acutally GOOD. (Local maximum in the correct region, but not global maximum) Let's hope that with a better tuned pile-up rejection, developed in #1802, the way the automatic check is implemented here, will work

@m-c-danisch Well now would be the time to change it to 

virtual Int_t | GetMaximumBin (Int_t &locmax, Int_t &locmay, Int_t &locmaz) const

with in the constraints already sitting there.

bazinski:
> Hi, I'm afraid that at the moment asking for `if (maxbin < mPulseHeightPeakRegion.first || maxbin > mPulseHeightPeakRegion.second)` will lead to BAD quality very often when the quality is acutally GOOD. (Local maximum in the correct region, but not global maximum) Let's hope that with a better tuned pile-up rejection, developed in #1802, the way the automatic check is implemented here, will work

Instead of using the Max of the histogram, walk through the bins from mPulseHeightPeakRegion.first to .second and figure out the local max ?

bazinski (approved):

knopers8:
please change the status of PR to "open" instead of "draft" when you consider it ready.

PR #1805, opened 2023-05-29
[EMCAL-530] Fix cell double counting in on-the-fly recalibration

mfasDa:
- Forward start marker in calibrated cell container
- Add taskParameter for artificial sleep in order to provide some time needed to attach the VSCode debugger.

### DIFF
--- a/Modules/EMCAL/src/ClusterTask.cxx
+++ b/Modules/EMCAL/src/ClusterTask.cxx
@@ -118,6 +118,13 @@ void ClusterTask::initialize(o2::framework::InitContext& /*ctx*/)
     return input == "true";
   };
 
+  if (hasConfigValue("debuggerDelay")) {
+    if (get_bool("debuggerDelay")) {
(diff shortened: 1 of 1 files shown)


mfasDa:
The actual bugfix is adding the line

```c++
currentlast = calibratedCells.size();
```

Without this line the event start marker is at 0, which means that events at the beginning of a timeframe are counted several times, since the clusterizer includes cells always from the start marker.

mfasDa:
I would like to keep these lines in:

``` c++
  if (hasConfigValue("debuggerDelay")) {
    if (get_bool("debuggerDelay")) {
      // set delay in order to allow for attaching the debugger
      sleep(20);
    }
  }
```

They don't harm as the default case is false, but in case requested adds an artificial delay, which leaves sufficient time to attach the VSCode debugger (gdb/lldb in the backend) to the process. This can be quite useful.

mfasDa:
CI error unrelated

Barthelemy (approved):

PR #1811, opened 2023-05-30
Fix checks of MID calibration objects

dstocco:
The last PR: https://github.com/AliceO2Group/QualityControl/pull/1770 introduced a bug in the check of calibration objects, leading to a crash at P2. I temporarily disabled the QC check at P2.
This PR should solve the issue.


dstocco:
@Barthelemy @knopers8 : if you spot no issue with this PR can this be merged? I'd like it to be in the next flp-suite release since it fixes an issue that is currently preventing the checks on the shifter plots at P2 during calibration runs.
My apologies for the pressure. Thanks!

knopers8 (approved):

PR #1814, opened 2023-05-31
[WIP] ITS Deconding Errors: new error check for GBTLinkErrors

IsakovAD:
- added separate quality checks for each ErrorID
- Improved info messages, now QC shows info about error id
- Added "medium" quality level
- included QualityReason for QualitySummary

### DIFF
--- a/Modules/ITS/include/ITS/ITSDecodingErrorCheck.h
+++ b/Modules/ITS/include/ITS/ITSDecodingErrorCheck.h
@@ -40,6 +40,34 @@ class ITSDecodingErrorCheck : public o2::quality_control::checker::CheckInterfac
   // Override interface
   Quality check(std::map<std::string, std::shared_ptr<MonitorObject>>* moMap) override;
   void beautify(std::shared_ptr<MonitorObject> mo, Quality checkResult = Quality::Null) override;
+  std::vector<int> vDecErrorLimits, vListErrorIdBad, vListErrorIdMedium;
+  bool doFlatCheck = false;
+
+  TString sErrorDesc[23] = {
+    "Page data does not start with expected RDH",                        // ErrNoRDHAtStart
+    "RDH is stopped, but the time is not matching the stop packet",      // ErrPageNotStopped
+    "Page with RDH.stop does not contain diagnostic word only",          // ErrStopPageNotEmpty
+    "RDH page counters for the same RU/trigger are not continuous",      // ErrPageCounterDiscontinuity
+    "RDH and GBT header page counters are not consistent",               // ErrRDHvsGBTHPageCnt
+    "GBT trigger word was expected but not found",                       // ErrMissingGBTTrigger
+    "GBT payload header was expected but not found",                     // ErrMissingGBTHeader
+    "GBT payload trailer was expected but not found",                    // ErrMissingGBTTrailer
+    "All lanes were stopped but the page counter in not 0",              // ErrNonZeroPageAfterStop
+    "End of FEE data reached while not all lanes received stop",         // ErrUnstoppedLanes
+    "Data was received for stopped lane",                                // ErrDataForStoppedLane
+    "No data was seen for lane (which was not in timeout)",              // ErrNoDataForActiveLane
+    "ChipID (on module) was different from the lane ID on the IB stave", // ErrIBChipLaneMismatch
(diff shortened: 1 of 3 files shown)


IsakovAD:
@iravasen with this PR, we need to modiffy parameters of the DecodingError task in the consul's json

Modules/ITS/include/ITS/ITSDecodingErrorCheck.h
```diff
@@ -40,6 +40,34 @@ class ITSDecodingErrorCheck : public o2::quality_control::checker::CheckInterfac
   // Override interface
   Quality check(std::map<std::string, std::shared_ptr<MonitorObject>>* moMap) override;
   void beautify(std::shared_ptr<MonitorObject> mo, Quality checkResult = Quality::Null) override;
+  std::vector<int> vDecErrorLimits, vListErrorIdBad, vListErrorIdMedium;
+  bool doFlatCheck = false;
+
+  TString sErrorDesc[23] = {
```
iravasen, line 46: instead of defining the local list of names you can do:
```
o2::itsmft::GBTLinkDecodingStat statistics;
statistics.ErrNames[<some_number>]; //to get the string you have defined locally
```
In this way if there is an update in O2, we have it for free also in our QC

Modules/ITS/src/ITSDecodingErrorCheck.cxx
```diff
@@ -29,15 +30,41 @@ namespace o2::quality_control_modules::its
...
 
   for (auto& [moName, mo] : *moMap) {
     (void)moName;
     if ((mo->getName() == "General/LinkErrorPlots") || (mo->getName() == "General/ChipErrorPlots")) {
       result = Quality::Good;
       auto* h = dynamic_cast<TH1D*>(mo->getObject());
-      if (h->GetMaximum() > 200) {
-        result.set(Quality::Bad);
+
+      if (doFlatCheck || mo->getName() == "General/ChipErrorPlots") {
```
iravasen, line 47: shouldn't you put `if( doFlatCheck && mo->getName() == "General/ChipErrorPlots")`?

Modules/ITS/src/ITSDecodingErrorCheck.cxx
```diff
@@ -29,15 +30,41 @@ namespace o2::quality_control_modules::its
...
+
+          if (vDecErrorLimits[iBin - 1] <= h->GetBinContent(iBin)) {
+            vListErrorIdBad.push_back(iBin - 1);
+            result.set(Quality::Bad);
+            result.addReason(o2::quality_control::FlagReasonFactory::Unknown(), Form("BAD: ID = %d, %s", iBin - 1, sErrorDesc[iBin - 1].Data()));
+          } else if (vDecErrorLimits[iBin - 1] / 2 < h->GetBinContent(iBin)) {
+            vListErrorIdMedium.push_back(iBin - 1);
+            if (result != Quality::Bad)
+              result.addReason(o2::quality_control::FlagReasonFactory::Unknown(), Form("Medium: ID = %d, %s", iBin - 1, sErrorDesc[iBin - 1].Data()));
+            result.set(Quality::Medium);
```
iravasen, line 65: I would include this in the `if (result != Quality::Bad)` just above so the set quality = medium only if the quality is not already bad.

Modules/ITS/include/ITS/ITSDecodingErrorCheck.h
IsakovAD: fixed

Modules/ITS/src/ITSDecodingErrorCheck.cxx
IsakovAD: Basically, I wrote it this way before, because I expect to always apply flat threshold for ChipErrorPlots, since we do not use it for shifters. 

I have fixed the code, so it should be more clear to read.

Modules/ITS/src/ITSDecodingErrorCheck.cxx
IsakovAD: Thanks! Indeed, it was a bug. Fixed

IsakovAD:
@iravasen Thank you for reviewing the code. I have applied your comments in the new commit.

iravasen:
@IsakovAD thanks a lot, approving.

PR #1821, opened 2023-06-07
[QC-429] Introduce Timekeeper to track validity of objects

knopers8:
We do not propagate its results to QC objects yet, but log them instead. Once we see the logged values make sense in different setups, we can switch.

Until we can access DataTakingContext during initialization, we are using a naive workaround for determining the deployment mode. We are also assuming that tf has 32 orbits, which does not apply for old data.

### DIFF
--- a/Framework/include/QualityControl/Timekeeper.h
+++ b/Framework/include/QualityControl/Timekeeper.h
@@ -1,0 +1,79 @@
+// Copyright 2019-2020 CERN and copyright holders of ALICE O2.
+// See https://alice-o2.web.cern.ch/copyright for details of the copyright holders.
+// All rights not expressly granted are reserved.
+//
+// This software is distributed under the terms of the GNU General Public
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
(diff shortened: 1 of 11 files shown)


Framework/include/QualityControl/Timekeeper.h
```diff
@@ -0,0 +1,79 @@
...
+#ifndef QUALITYCONTROL_TIMEKEEPER_H
+#define QUALITYCONTROL_TIMEKEEPER_H
+
+#include "QualityControl/ValidityInterval.h"
+
+namespace o2::framework
+{
+struct ProcessingContext;
+struct TimingInfo;
+} // namespace o2::framework
```
knopers8, line 26: i should remove that

Framework/include/QualityControl/Timekeeper.h
```diff
@@ -0,0 +1,79 @@
...
+namespace o2::quality_control::core
+{
+
+// could be moved to a separate file
+using TimeframeIdRange = o2::math_utils::detail::Bracket<uint32_t>;
+const static TimeframeIdRange gInvalidTimeframeIdRange{
+  std::numeric_limits<uint32_t>::max(), std::numeric_limits<uint32_t>::min()
+};
+
+class Timekeeper
```
Barthelemy, line 37: you could perhaps add a word about this class

Barthelemy (approved):
very nice, thank you

PR #1824, opened 2023-06-08
ITS FHR: small bug fix

iravasen:
With Fill() one has to use the real numbers and not numbers + 1


IsakovAD (approved):

IsakovAD:
@knopers8 Could you please merge the PR? We have errors in `error: dependency cycle: stage/lib64/G__O2SimulationDataFormat_rdict.pcm -> stage/lib64/libO2DataFormatsCTP.so -> DataFormats/Detectors/CTP/CMakeFiles/O2lib-DataFormatsCTP.dir/src/Digits.cxx.o -> cmake_object_order_depends_target_O2lib-DataFormatsCTP` which is not related to our changes

PR #1825, opened 2023-06-09
fix log about computing partition id

knopers8:


knopers8:
no need for builds to run to test a typo, merging

PR #1826, opened 2023-06-09
Ctptest1

mbombara:
Task for reading CTP readout - publishing 3 histograms (BCID, Inputs and Classes).

### DIFF
--- a/Modules/CTP/src/RawDataQcTask.cxx
+++ b/Modules/CTP/src/RawDataQcTask.cxx
@@ -65,13 +65,13 @@ void CTPRawDataReaderTask::startOfCycle()
 
 void CTPRawDataReaderTask::monitorData(o2::framework::ProcessingContext& ctx)
 {
-  //LOG(info) << "============  Starting monitoring ================== ";
-  // get the input
+  // LOG(info) << "============  Starting monitoring ================== ";
+  //  get the input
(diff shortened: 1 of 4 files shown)


Modules/CTP/src/RawDataQcTask.cxx
```diff
@@ -46,6 +46,8 @@ void CTPRawDataReaderTask::initialize(o2::framework::InitContext& /*ctx*/)
   getObjectsManager()->startPublishing(mHistoBC);
   getObjectsManager()->startPublishing(mHistoInputs);
   getObjectsManager()->startPublishing(mHistoClasses);
+  mDecoder.setDoLumi(1);
```
Barthelemy, line 49: These two methods take a boolean, please don't use `0` and `1` but use a boolean.

Barthelemy (approved):
A very minor comment otherwise it is good.

Modules/CTP/src/RawDataQcTask.cxx
mbombara: Ok, thanks! Should I change it now, or in the next release is ok? :-)

Barthelemy: next release is ok

PR #1830, opened 2023-06-12
Set Activity in Check Runner again

knopers8:
this rolls back the changes of the commit below without having to revert a lot of intertwined commits. https://github.com/AliceO2Group/QualityControl/commit/82be5ff2700eed311483ace20d29dd284e4e4b3e

we need to think how aggregators should behave in the new scenario for a proper fix


knopers8:
tests pass on my laptop, merging for a patch release

PR #1834, opened 2023-06-13
[QC-984] Fix pp finalize no init

Barthelemy:


Barthelemy:
@knopers8 It is not clear to me whether we should fix it somewhere else. It is not clear either why we never get to `initialize`

Framework/src/PostProcessingRunner.cxx
```diff
@@ -237,6 +237,9 @@ void PostProcessingRunner::doUpdate(const Trigger& trigger)
 
 void PostProcessingRunner::doFinalize(const Trigger& trigger)
 {
+  if(mTaskState != TaskState::Running) {
+    ILOG(Warning, Support) << "Attempt at finalizing the user task although it was not initialized. Skipping the finalization." << ENDM;
```
knopers8, line 241: this does not skip it ;)

Barthelemy: damn

PR #1835, opened 2023-06-14
(important) ITS FEE: removed TCanvas

IsakovAD:
- Removed TCanvas with Fraction of bad lanes seemingly, it was creating crash of the FEE task on merger
- TLine and TLatex were moved to the ListOfFunctions


iravasen (approved):

IsakovAD:
Hello @Barthelemy, 

Could you, please, deploy this PR to the FLP patch as soon as possible? 

It fixes an issue crashing FEE task, which is currently is switched off at P2.

Barthelemy (approved):

PR #1836, opened 2023-06-14
FV0: Use FPGA logic in trigger calculation

andreasmolander:


andreasmolander:
We compare trigger signals from hardware, with triggers signals calculated with the same conditions in QC software, to verify that the data is not corrupted. Previously we saw some discrepancy, but when using this logic in QC, which is exactly the same as in the hardware, the triggers match perfectly.

knopers8 (approved):

PR #1845, opened 2023-06-20
[QC-429] Correct validity information in NewObject and Periodic Triggers

knopers8:
This will allow TrendingTask and similar ones to use the validity end as timestamps on trends. If they were to use validFrom of objects with correct validity, all the points would land in the same timestamp. Using validUntil instead makes sense, as it basically indicates the last update of an object.

There will be a separate PR changing the code in Trending-like tasks, since that one may take longer to pass reviews of all module owners.


Barthelemy (approved):

Barthelemy:
errors unrelated, merging.

PR #1846, opened 2023-06-20
[QC-429] Enable new object validity rules

knopers8:
Not ready to merge, but it shows the last parts of the code needed to switch to the new validity rules.


Barthelemy:
could you put it in WIP if it is not ready ?

Barthelemy (approved):
Good, and thanks to O2_QC_OLD_VALIDITY we are safe.

PR #1848, opened 2023-06-20
Add separate histograms BcVsFeeModules and OutOfBunchColl for each trigger

DawidSkora:
Add separate histograms BcVsFeeModules and OutOfBunchColl for each trigger. Align OutOfBunchCollFeeModulesCheck to with new histograms.


Barthelemy (approved):

Barthelemy:
Error unrelated

PR #1849, opened 2023-06-22
[QC-429] Trending tasks should use End Of Validity for point timestamps

knopers8:
If trended objects follow the new validity rules, they should use the end of validity as the timestamp of a data point, since the start will be usually the same (SOR). I did not touch tasks which use creation time or current timestamp instead, since these will not be affected by the change. FIT tasks have a way to select what kind of timestamps should be used, so I added a new option for "validUntil".

### DIFF
--- a/Framework/src/ActivityHelpers.cxx
+++ b/Framework/src/ActivityHelpers.cxx
@@ -102,7 +102,9 @@ std::function<validity_time_t(void)> getCcdbSorTimeAccessor(uint64_t runNumber)
 
 std::function<validity_time_t(void)> getCcdbEorTimeAccessor(uint64_t runNumber)
 {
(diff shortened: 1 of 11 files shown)


knopers8:
Needs https://github.com/AliceO2Group/QualityControl/pull/1845, but can be already reviewed.

Barthelemy (approved):

Barthelemy:
Do we need more reviews or can I merge ?

knopers8:
Yes, I would appreciate a cross-check of the task owners ( @wiechula , @njacazio , @aferrero2707 , @sbysiak , @bazinski ).

bazinski (approved):

wiechula (approved):

aferrero2707 (approved):

Barthelemy:
@knopers8 There seems to be a legit issue now with the checks.

knopers8:
Yes, bad merge.

PR #1850, opened 2023-06-22
[QC-429] Use CCDB RCT as the first source for SOR/EOR in async QC

knopers8:


Barthelemy (approved):
understood. good, thank you

PR #1853, opened 2023-06-26
ITS General: correct merging of averages with THRatio

IsakovAD:
Averaged objects in Track and Cluster tasks were updated to THRatios

### DIFF
--- a/Modules/ITS/include/ITS/ITSTrackTask.h
+++ b/Modules/ITS/include/ITS/ITSTrackTask.h
@@ -57,6 +57,14 @@ class ITSTrackTask : public TaskInterface
   void formatAxes(TH1* h, const char* xTitle, const char* yTitle, float xOffset = 1., float yOffset = 1.);
   void addObject(TObject* aObject);
   void createAllHistos();
+  template <class T>
+  void formatAxes(T* obj, const char* xTitle, const char* yTitle, float xOffset, float yOffset)
+  {
+    obj->GetXaxis()->SetTitle(xTitle);
+    obj->GetYaxis()->SetTitle(yTitle);
+    obj->GetXaxis()->SetTitleOffset(xOffset);
+    obj->GetYaxis()->SetTitleOffset(yOffset);
+  }
 
   static constexpr int NLayer = 7;
   static constexpr int NLayerIB = 3;
@@ -85,7 +93,7 @@ class ITSTrackTask : public TaskInterface
   std::unique_ptr<TH2DRatio> hHitFirstLayerPhi6cls;
   std::unique_ptr<TH2DRatio> hHitFirstLayerPhi7cls;
   TH2D* hClusterVsBunchCrossing;
-  std::unique_ptr<TH2DRatio> hNClusterVsChipITS;
+  TH2D* hNClusterVsChipITS;
 
   float mVertexXYsize = 0.5;
   float mVertexZsize = 15.;
--- a/Modules/ITS/include/ITS/ITSTrackTask.h
+++ b/Modules/ITS/include/ITS/ITSTrackTask.h
@@ -25,6 +25,8 @@
 #include <Framework/TimingInfo.h>
 #include <TTree.h>
 #include <TLine.h>
+#include "Common/TH1Ratio.h"
+#include "Common/TH2Ratio.h"
 
 class TH1D;
 class TH2D;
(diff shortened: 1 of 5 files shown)


Modules/ITS/include/ITS/ITSTrackTask.h
```diff
@@ -63,24 +74,24 @@ class ITSTrackTask : public TaskInterface
...
-  TH2D* hHitFirstLayerPhi7cls;
+  std::unique_ptr<TH2DRatio> hNClustersPerTrackEta;
+  std::unique_ptr<TH2DRatio> hNClustersPerTrackPhi;
+  std::unique_ptr<TH2DRatio> hHitFirstLayerPhiAll;
+  std::unique_ptr<TH2DRatio> hHitFirstLayerPhi4cls;
+  std::unique_ptr<TH2DRatio> hHitFirstLayerPhi5cls;
+  std::unique_ptr<TH2DRatio> hHitFirstLayerPhi6cls;
+  std::unique_ptr<TH2DRatio> hHitFirstLayerPhi7cls;
   TH2D* hClusterVsBunchCrossing;
   TH2D* hNClusterVsChipITS;
```
iravasen, line 96: I think also `hNClusterVsChipITS` needs the same treatment. In general all the plots with the `kIsAverage` bit set. Maybe give another rapid look at the two tasks to be sure you did not miss any plot, thank you!

Modules/ITS/src/ITSClusterTask.cxx
```diff
@@ -595,14 +582,14 @@ void ITSClusterTask::addObject(TObject* aObject)
   } else
     mPublishedObjects.push_back(aObject);
 }
-
+/*
```
iravasen, line 585: let's remove this function if you don't need it anymore

iravasen:
On top of the comments I'm not fully sure about / I would like to mention:
- usage of `update` at the end of cycle. Is it really needed? According to https://github.com/AliceO2Group/QualityControl/blob/master/Modules/Common/include/Common/TH2Ratio.inl it seems to me the relevant function is `merge` which has a call to `update()` inside and which should (to my knowledge) be the function which is called on the merger. 
- In the cluster task, in the `reset()` method we need to reset also the counters `mClusterSize` and `nClusters` for example (I might miss some counters here). Otherwise on the merger side we overcount. 
- We have to check carefully the errors associated to num and den to be sure that the poissonian error works in all cases. 
- In the object shipped to CCDB do we still have the info about num and den or only the ratio is stored? I'm wondering about async QC and the merging done there among different chunks.

iravasen:
let's have maybe a discussion tomorrow

Modules/ITS/include/ITS/ITSTrackTask.h
IsakovAD: Yes, indeed, I didn't implement TH2Ratio for `hNClusterVsChipITS `. The problem is that this object has non-uniform binning defined by an array, and it is impossible to construct the THRatio object in such a way. We need to change the mother class to have  `hNClusterVsChipITS ` scaled correctly. Otherwise, we can keep this object as it is.

IsakovAD:
@iravasen Thank you for reviewing the code. There are my replies:

In general, the main idea of this THRatio object that it includes a denominator, a nominator and a TH with ratio of both. So, you can work with it as follows:

`object->getDen()->GetBinContent(1,1)` -> to access bin 1,1 of Denominator 
or  
`object->getNum()->GetBinContent(1,1)` -> to access bin 1,1 of Numerator 
or
`object->GetBinContent(1,1)` -> to access ratio of Numerator bin 1,1 /  Denominator bin 1,1 

This ratio is calculated in `update()` function

1) I would keep this at the end of cycle, since we need it to calculate the ratio. I was using the following instruction written by Andrea for THRatio objects: https://codimd.web.cern.ch/6bCknFwPQ6WUJSYKyEsPZg. 
2) Indeed, it is need to be changed, thanks. 
3) In the `init` function of THRatio we set Sumw2() for den, num and T object. So I think that we are covered by this in terms of errors. See, https://github.com/AliceO2Group/QualityControl/blob/3f5102af6a692d5b2ccbbe87217d638ccd5ef086/Modules/Common/include/Common/TH1Ratio.inl#L89-L99
4) See the first lines of explanation

Modules/ITS/include/ITS/ITSTrackTask.h
iravasen: ok good point, we could keep the object as it is. In the end we are not interested in the absolute value of the z axis.

Modules/ITS/src/ITSClusterTask.cxx
IsakovAD: Done

IsakovAD:
@iravasen New commit:

1) Deleted `formatAxes`
2) Introduced reset of nCLusters and ClusterSize counter each cycle
3) fixed a bug in  updated function of ClusterTask, which would make a crash for `PublishDetailedSummary=0` option

iravasen:
Thanks a lot! All good for me!

iravasen (approved):

PR #1854, opened 2023-06-26
[OCTRL-791] Get Run info from ECS

Barthelemy:


Framework/include/QualityControl/CheckRunner.h
```diff
@@ -219,7 +219,7 @@ class CheckRunner : public framework::Task
   std::string mDeviceName;
   std::map<std::string, Check> mChecks;
   std::string mDetectorName;
-  std::shared_ptr<Activity> mActivity;
+  std::shared_ptr<Activity> mActivity; // why a shared_ptr ?
```
knopers8, line 222: I think it was to share it with Checks.

Framework/include/QualityControl/runnerUtils.h
```diff
@@ -74,87 +74,44 @@ inline bool hasChecks(std::string configSource)
...
-  ILOG(Debug, Devel) << "Pass Name returned by computePassName : " << fallbackPassName << ENDM;
-  return fallbackPassName;
-}
-
-inline std::string computeProvenance(const std::string& fallbackProvenance = "")
-{
-  ILOG(Debug, Devel) << "Provenance returned by computeProvenance : " << fallbackProvenance << ENDM;
-  return fallbackProvenance;
+  result = result > 0 /* found it in service */ ? result : fallbackNumber;
+  ILOG(Info, Devel) << name << " returned by computeRunNumber (default) : " << result << ENDM;
```
knopers8, line 90: this log could be updated if you use this template not only for run number anymore.

Framework/include/QualityControl/runnerUtils.h
```diff
@@ -74,87 +74,44 @@ inline bool hasChecks(std::string configSource)
...
-    computePassName(fallbackActivity.mPassName),
-    computeProvenance(fallbackActivity.mProvenance),
-    fallbackActivity.mValidity
-  };
-  // if available we overwrite with the info from logbook.
-  Bookkeeping::getInstance().populateActivity(activity, runNumber);
+    runType,
+    fallbackActivity.mPeriodName,
+    fallbackActivity.mPassName,
+    fallbackActivity.mProvenance,
```
knopers8, line 110: why fallback activity instead of taking the computed values if they are available?

Framework/include/QualityControl/Activity.h
```diff
@@ -31,6 +31,8 @@ namespace o2::quality_control::core
 /// \author   Barthelemy von Haller
 class Activity
 {
+  // TODO make it a struct ?
```
knopers8, line 34: no harm in that

Framework/include/QualityControl/CheckRunner.h
Barthelemy: ok, thank you, I will modify the comment

Framework/include/QualityControl/runnerUtils.h
Barthelemy: correct, I'll fix that

Framework/include/QualityControl/Activity.h
Barthelemy: I am just wondering if there is an impact on the Serialization with ROOT

PR #1855, opened 2023-06-28
[QC-429] Force 10y validity for the time being

knopers8:
In this corner case, we were downloading SOR and EOR of an existing run (42) from BK, but overwriting SOR in CheckRunner to keep the old behaviour. However, we were not overwriting EOR, thus using the time from the past, making validity invalid. This forces CcdbDatabase to apply 10y validity until we move to the new rules.


Barthelemy (approved):

Barthelemy:
Thank you

PR #1864, opened 2023-07-01
GLO: add trending to MeanVertex post-processing

aferrero2707:
This is an alternative method to trend the position of the mean vertex retrieved from the `MeanVertex` calibration objects.
Instead of filling intermediate histograms and using the generic `TrendingTask` to plot the graphs, the graphs are directly implemented in the `MeanVertexPostProcessing` task.

This has few advantages:
- the values are directly retrieved from the calibration objects, hence they are not averaged over the run duration as in the current implementation
- the graphs are beautified by setting the vertical axis range and by adding horizontal reference lines to see when the values go outside the acceptable range. The limits are configurable via task parameters.

I have also changed a bit the way the CCDB objects are retrieved, using the time stamp from the post-processing trigger instead of the current time, as this seems more correct.

### DIFF
--- a/Modules/GLO/include/GLO/MeanVertexPostProcessing.h
+++ b/Modules/GLO/include/GLO/MeanVertexPostProcessing.h
@@ -53,6 +53,8 @@ class MeanVertexPostProcessing final : public quality_control::postprocessing::P
  public:
   /// \brief Constructor
   MeanVertexPostProcessing() = default;
+  /// \brief Destructor
+  ~MeanVertexPostProcessing() override;
 
   /// \brief Configuration of a post-processing task.
   /// Configuration of a post-processing task. Can be overridden if user wants to retrieve the configuration of the task.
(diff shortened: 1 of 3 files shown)


aferrero2707:
@knopers8 @chiarazampolli @shahor02 can you have a look and give me your opinion? Thanks!

Modules/GLO/include/GLO/MeanVertexPostProcessing.h
```diff
@@ -20,20 +20,37 @@
...
+#include <TCanvas.h>
 
 namespace o2::quality_control_modules::glo
 {
 
+class TrendGraph : public TCanvas
+{
+ public:
+  TrendGraph(std::string name, std::string title, std::string label, float rangeMin, float rangeMax);
+
```
shahor02, line 35: Checking superficially, should work. But the TrendGraph should have a virtual destructor.

aferrero2707: @shahor02 thanks for checking! The virtual destructor is needed even if it would be empty?

Barthelemy (approved):

PR #1865, opened 2023-07-03
[QC-977] Select automatically the correct params 

Barthelemy:
based on the current activity.

### DIFF
--- a/Framework/src/CustomParameters.cxx
+++ b/Framework/src/CustomParameters.cxx
@@ -72,7 +72,7 @@ std::optional<std::string> CustomParameters::atOptional(const std::string& key, 
   const int runType = activity.mType; // get the type for this run
   // convert it to a string (via a string_view as this is what we get from O2)
   const std::string_view runTypeStringView = o2::parameters::GRPECS::RunTypeNames[runType];
-  const std::string runTypeString = { runTypeStringView.begin(), runTypeStringView.end() };
+  const std::string runTypeString{ runTypeStringView };
   // get the param
   return atOptional(key, runTypeString, activity.mBeamType);
 }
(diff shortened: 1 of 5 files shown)


Barthelemy:
doc and test coming

Barthelemy:
ready

Framework/src/CustomParameters.cxx
```diff
@@ -10,7 +10,7 @@
 // or submit itself to any jurisdiction.
 
 #include "QualityControl/CustomParameters.h"
-
+#include "DataFormatsParameters/ECSDataAdapters.h"
```
knopers8, line 13:
```suggestion
#include <DataFormatsParameters/ECSDataAdapters.h>
```

Framework/src/CustomParameters.cxx
```diff
@@ -66,6 +66,17 @@ std::optional<std::string> CustomParameters::atOptional(const std::string& key,
   }
 }
 
+std::optional<std::string> CustomParameters::atOptional(const std::string& key, const Activity& activity) const
+{
+  // Get the proper parameter for the given activity
+  const int runType = activity.mType; // get the type for this run
+  // convert it to a string (via a string_view as this is what we get from O2)
+  const std::string_view runTypeStringView = o2::parameters::GRPECS::RunTypeNames[runType];
+  const std::string runTypeString = { runTypeStringView.begin(), runTypeStringView.end() };
```
knopers8, line 75:
```suggestion
  const std::string runTypeString{ runTypeStringView };
```
doesn't this work? a bit simpler.

knopers8 (approved):
i have two small correction proposals, but it looks good as it is

Barthelemy:
thank you for the suggestions, I committed them.

PR #1869, opened 2023-07-05
[QC-992] Optimize port selection procedure in Service Discovery

Barthelemy:
Try to start a server immediately on a selected port and retry a different one in case of failure.


Barthelemy:
@awegrzyn @knopers8 Tried to do it... not 100% sure but it works.

knopers8 (approved):

awegrzyn (approved):

PR #1876, opened 2023-07-10
FT0 DigitQcTask: hotfix for chID per BC plot

afurs:
Trivial, needed at P2


afurs:
Hi @Barthelemy , this is trivial hotfix, we would to have it at P2(next FLP release) ASAP. Could you please skip CICD?

Barthelemy:
Hi @afurs , I can skip the CI but that does mean that it will end up faster at P2 in the FLP Suite. Do you need me to do a patch version with it ?

Barthelemy (approved):

afurs:
> I can skip the CI but that does mean that it will end up faster at P2 in the FLP Suite

Yes, please patch it. Thanks!

afurs:
@Barthelemy for now we have this task only on FLP, but soon it may be migrated into EPN part. At QC node we cannot deploy such task.

Barthelemy:
@afurs ok then we need new rpms. I'll ask @vascobarroso

PR #1887, opened 2023-07-17
Fix histogram ratio copy functions

aferrero2707:
The update fixes a segfault in the `Copy()` functions whenever the numerator or denominator are not properly initialized.


aferrero2707:
@knopers8 @Barthelemy this is a quite urgent PR, as the fix is needed to properly run the ASYNC RelVal (see discussion [here](https://alice.its.cern.ch/jira/browse/QC-993)).

I will do further tests and work on a better fix next week.

knopers8 (approved):
thank you!

PR #1890, opened 2023-07-18
ITS: Added trigger ramp bit to RDH plot + updated checker Payload size

iravasen:
@IsakovAD @nicolovalle


IsakovAD (approved):

iravasen:
@IsakovAD in agreement with @nicolovalle I've also updated the checker for the payload size plot. This is important for the technical runs where now the max size goes beyond 1000 bytes reaching 1400 bytes. This is achieved now with a param which can be added to the json (default is indeed 1400).

knopers8 (approved):

PR #1893, opened 2023-07-26
ITS FEE: added RDH plot cumulative since SOX

iravasen:
This is to have a copy of the already-existing RDH plot which DOES NOT reset at every cycle. This is useful to monitor the amount of trigger ramp bits after SOX.


iravasen:
@IsakovAD @nicolovalle can you quickly revise this? In this way we merge it quickly so to have it at the next release of the SW.

IsakovAD (approved):

nicolovalle:
> @IsakovAD @nicolovalle can you quickly revise this? In this way we merge it quickly so to have it at the next release of the SW.

Thanks a lot for adding this. All good for what I can see.

PR #1895, opened 2023-07-27
[QC-869] Moving windows of selected plots in sync processing

knopers8:
Also brings "--full-chain" flag to o2-qc, which allows to run a QC chain with Mergers between a single instance of a QC task and a CheckRunner. Useful for testing that an object is mergeable and for testing moving windows generated by Mergers.

Needs https://github.com/AliceO2Group/AliceO2/pull/11705


Barthelemy (approved):
very nice, thank you

PR #1899, opened 2023-07-28
TPC: Update Quality Aggregator

mlesch:
New way of handling MetaData "Comments" in the TPC Quality-Aggregator: Only the comments of the QOs with the worst quality are aggregated and stored


wiechula (approved):

wiechula:
@Barthelemy , the error seem unrelated. Any suggestion?

PR #1900, opened 2023-07-30
MFT: fixed displaying of dead chips in digit maps 

tomas-herman:
- displaying of excluded (dead or masked) chips in the digit chip maps using the beautifier
- removal of redundant mNOfDigitsTime histogram that was causing messages in FLP infologgers (`Object mNOfDigitsTime bigger than maximum allowed size` and `Object mNOfDigitsTime will not be uploaded`)

### DIFF
--- a/Modules/MFT/src/QcMFTDigitCheck.cxx
+++ b/Modules/MFT/src/QcMFTDigitCheck.cxx
@@ -28,7 +28,6 @@
 #include <TPaveText.h>
 // O2
 #include <DataFormatsITSMFT/NoiseMap.h>
-#include "CCDB/CcdbApi.h"
 #include <ITSMFTReconstruction/ChipMappingMFT.h>
 
 // Quality Control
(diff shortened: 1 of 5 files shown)


knopers8 (approved):
BTW, perhaps using this method would also work for you? This uses the CCDB specified in `qc.config.conditionDB.url`:

https://github.com/AliceO2Group/QualityControl/blob/d459f9741c5f43586c477eb2fa5e7bebc1c277a2/Framework/include/QualityControl/UserCodeInterface.h#L59

Modules/MFT/src/QcMFTDigitCheck.cxx
```diff
@@ -105,8 +119,73 @@ Quality QcMFTDigitCheck::check(std::map<std::string, std::shared_ptr<MonitorObje
 
 std::string QcMFTDigitCheck::getAcceptedType() { return "TH1"; }
 
+void QcMFTDigitCheck::readMaskedChips()
+{
+  o2::ccdb::CcdbApi api;
+  api.init(mDeadMapCcdbAddress.data());
+  long timestamp = -1;
```
knopers8, line 126: If you are going to use this check in async processing, this might request a wrong object, not matching the time of data.

tomas-herman: Is there a standardised way of accessing CCDBB objects with the validity corresponding to the data?

knopers8: Not well exercised, but... `UserCodeInterface::retrieveConditionAny` would be the method to use and the timestamp could be `mo->getValidity().getMin()`, which now is still the time of creating an object, but soon it will be SOR in case of integrated plots and the cycle start in case of moving windows.

tomas-herman: Thanks a lot for pointing us to this. I modified the code using the methods you suggested.

PR #1903, opened 2023-07-31
ITS decoding error check

zz951:
Fix the bug for ErrID10


Modules/ITS/include/ITS/ITSDecodingErrorCheck.h
```diff
@@ -41,7 +41,7 @@ class ITSDecodingErrorCheck : public o2::quality_control::checker::CheckInterfac
   // Override interface
   Quality check(std::map<std::string, std::shared_ptr<MonitorObject>>* moMap) override;
   void beautify(std::shared_ptr<MonitorObject> mo, Quality checkResult = Quality::Null) override;
-  std::vector<int> vDecErrorLimits, vListErrorIdBad, vListErrorIdMedium;
+  std::vector<int> vDecErrorLimits, vDecErrorLimitsRatio, vDecErrorType, vListErrorIdBad, vListErrorIdMedium;
```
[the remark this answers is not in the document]
zz951: ok, I got it!

Modules/ITS/src/ITSDecodingErrorCheck.cxx
```diff
@@ -30,14 +30,30 @@ namespace o2::quality_control_modules::its
 
 Quality ITSDecodingErrorCheck::check(std::map<std::string, std::shared_ptr<MonitorObject>>* moMap)
 {
+  // set timer
+  std::chrono::time_point<std::chrono::high_resolution_clock> start;
+  std::chrono::time_point<std::chrono::high_resolution_clock> end;
+  int TIME;
+  start = std::chrono::high_resolution_clock::now();
```
IsakovAD, line 37: I would create start 'start' timer in the initialization of the check ( `init ()` function). 

Idea is that we need to now the total time of the run, not just a time you spend in one `check` function

Summarizing: 

init() : create object `std::chrono::time_point<std::chrono::high_resolution_clock> start;`, start timer

 check(): measure how much time we have after initialization (it should roughly correspond to start of the run)

Modules/ITS/src/ITSDecodingErrorCheck.cxx
```diff
@@ -30,14 +30,30 @@ namespace o2::quality_control_modules::its
...
+  std::vector<int> vDecErrorType = convertToArray<int>(o2::quality_control_modules::common::getFromConfig<string>(mCustomParameters, "DecLinkErrorType", ""));
+  if (vDecErrorType.size() != o2::itsmft::GBTLinkDecodingStat::NErrorsDefined) {
+    LOG(error) << "Incorrect vector with DecodingError Type, check .json" << ENDM;
+    doFlatCheck = true;
+  }
 
   Quality result = Quality::Null;
-
+  end = std::chrono::high_resolution_clock::now();
+  TIME = std::chrono::duration_cast<std::chrono::microseconds>(end - start).count();
```
IsakovAD, line 56: microseconds are too much for us) let`s use just a seconds?

IsakovAD:
@zz951 nice work! Could you please have a look on my comments?

zz951:
> @zz951 nice work! Could you please have a look on my comments?

Thank you! I will modify it soon!

Modules/ITS/src/ITSDecodingErrorCheck.cxx
```diff
@@ -31,10 +31,10 @@ namespace o2::quality_control_modules::its
 Quality ITSDecodingErrorCheck::check(std::map<std::string, std::shared_ptr<MonitorObject>>* moMap)
 {
   // set timer
-  std::chrono::time_point<std::chrono::high_resolution_clock> start;
-  std::chrono::time_point<std::chrono::high_resolution_clock> end;
-  int TIME;
-  start = std::chrono::high_resolution_clock::now();
+  if (mTFCount == 0) {
```
IsakovAD, line 34: In reality, QCCheck is called each QC Cycle (30 sec to 2 min, depending on task), not TF, so I would rename the variable to `nCycle` to avoid confusion

The situation is different for QC Task, which is indeed called each new TF

Modules/ITS/src/ITSDecodingErrorCheck.cxx
```diff
@@ -106,6 +104,8 @@ Quality ITSDecodingErrorCheck::check(std::map<std::string, std::shared_ptr<Monit
       }
     }
   }
+  end = std::chrono::high_resolution_clock::now();
+  TIME = std::chrono::duration_cast<std::chrono::seconds>(end - start).count();
```
IsakovAD, line 108: You updated `TIME` variable at the end of `check()` function, as result, you use the old `TIME` value from the previous cycle. 

To solve that, can you try to modify the "if"case from the start of the function to something like that: 

```
if (mTFCount == 0) {
    start = std::chrono::high_resolution_clock::now();
    mTFCount++;
} else{
   end = std::chrono::high_resolution_clock::now();
  TIME = std::chrono::duration_cast<std::chrono::seconds>(end - start).count();
}
```

IsakovAD:
@zz951 we are almost here. Just a couple of minor comment

Modules/ITS/src/ITSDecodingErrorCheck.cxx
zz951: Ok, done!

Modules/ITS/src/ITSDecodingErrorCheck.cxx
```diff
@@ -30,14 +30,30 @@ namespace o2::quality_control_modules::its
...
+  std::chrono::time_point<std::chrono::high_resolution_clock> start;
+  std::chrono::time_point<std::chrono::high_resolution_clock> end;
+  int TIME;
+  start = std::chrono::high_resolution_clock::now();
   std::vector<int> vDecErrorLimits = convertToArray<int>(o2::quality_control_modules::common::getFromConfig<string>(mCustomParameters, "DecLinkErrorLimits", ""));
   if (vDecErrorLimits.size() != o2::itsmft::GBTLinkDecodingStat::NErrorsDefined) {
     LOG(error) << "Incorrect vector with DecodingError limits, check .json" << ENDM;
     doFlatCheck = true;
   }
+  std::vector<int> vDecErrorLimitsRatio = convertToArray<int>(o2::quality_control_modules::common::getFromConfig<string>(mCustomParameters, "DecLinkErrorLimitsRatio", ""));
```
IsakovAD, line 46: you are casting to `int`, while you need to have `float` here

IsakovAD: Actually, if you try to run `convertToArray<float>`, it will crash, because the function doesn't have a case for float. 

Could you please modify it like: 

```
  std::vector<T> convertToArray(std::string input)
  {

    std::istringstream ss{ input };

    std::vector<T> result;
    std::string token;

    while (std::getline(ss, token, ',')) {

      if constexpr (std::is_same_v<T, int>) {
        result.push_back(std::stoi(token));
      } if constexpr (std::is_same_v<T, float>) {
        result.push_back(std::stof(token));
      } else if constexpr (std::is_same_v<T, std::string>) {
        result.push_back(token);
      }
    }
    return result;
  }
 ```  
 
Here I added new `constexp` condition also for float

File: Modules/ITS/include/ITS/ITSDecodingErrorCheck.h [hunk 7]
```diff
@@ -41,7 +41,8 @@ class ITSDecodingErrorCheck : public o2::quality_control::checker::CheckInterfac
   // Override interface
   Quality check(std::map<std::string, std::shared_ptr<MonitorObject>>* moMap) override;
   void beautify(std::shared_ptr<MonitorObject> mo, Quality checkResult = Quality::Null) override;
-  std::vector<int> vDecErrorLimits, vListErrorIdBad, vListErrorIdMedium;
+  std::vector<int> vDecErrorLimits, vDecErrorType, vListErrorIdBad, vListErrorIdMedium;
```
IsakovAD, line 44: In principle, we don't need some of these vectors to be defined here, since we create `vDecErrorLimits`, `vDecErrorLimitsRatio`, `vDecErrorType` in the .cxx file

IsakovAD:
@iravasen @nicolovalle Could you please review this PR? For me all is fine

Modules/ITS/src/ITSDecodingErrorCheck.cxx
```diff
@@ -30,14 +30,31 @@ namespace o2::quality_control_modules::its
...
+    TIME = std::chrono::duration_cast<std::chrono::seconds>(end - start).count();
+  }
   std::vector<int> vDecErrorLimits = convertToArray<int>(o2::quality_control_modules::common::getFromConfig<string>(mCustomParameters, "DecLinkErrorLimits", ""));
   if (vDecErrorLimits.size() != o2::itsmft::GBTLinkDecodingStat::NErrorsDefined) {
     LOG(error) << "Incorrect vector with DecodingError limits, check .json" << ENDM;
     doFlatCheck = true;
   }
+  std::vector<float> vDecErrorLimitsRatio = convertToArray<float>(o2::quality_control_modules::common::getFromConfig<string>(mCustomParameters, "DecLinkErrorLimitsRatio", ""));
+  if (vDecErrorLimitsRatio.size() != o2::itsmft::GBTLinkDecodingStat::NErrorsDefined) {
+    LOG(error) << "Incorrect vector with DecodingError limits Ratio, check .json" << ENDM;
```
knopers8, line 48: I will merge it like this before the upcoming QC release, but here you are mixing fairlogger with infologger interface. FairLogger does not need `ENDM` at the end, as it recognizes the end of line. `ENDM` shall be used with infologger (`ILOG(Error)`).

zz951: OK，thanks a lot! Then, should I modify it and commit again right now or next time?

knopers8: As you prefer. If you are not waiting for Nicolo's review before merging this one, I would propose that I merge this one and you do these changes in another PR, so we are not blocked by failing CI by any chance.

zz951: ok, thanks a lot! I will do these changes next time.

PR #1906, opened 2023-07-31
Improvements to the THRatio initialization and interface

aferrero2707:
Various improvements to the THRatio copying and cloning functions:
- numerator and denominator plots are initialized with dummy parameters in the default constructor, instead of assigning `nullptr` to the corresponding pointers
- added override functions to update the histo name and title, and propagate the changes to the internal numerator and denominator
- Copy() function also updates the `mUniformScaling` parameter of the destination object
- added unit tests for Copy() and Clone() functions

### DIFF
--- a/Modules/Common/include/Common/TH1Ratio.inl
+++ b/Modules/Common/include/Common/TH1Ratio.inl
@@ -25,6 +25,8 @@ TH1Ratio<T>::TH1Ratio()
     o2::mergers::MergeInterface(),
     mUniformScaling(true)
 {
+  // do not add created histograms to gDirectory
+  // see https://root.cern.ch/doc/master/TEfficiency_8cxx.html
   {
     TDirectory::TContext ctx(nullptr);
     mHistoNum = new T("num", "num", 10, 0, 10);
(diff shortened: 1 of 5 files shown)


Modules/Common/include/Common/TH2Ratio.inl
```diff
@@ -19,19 +19,38 @@
...
+  {
+    TString name_num = T::GetName() + TString("_num");
+    TString name_den = T::GetName() + TString("_den");
+    TString title_num = T::GetTitle() + TString(" num");
+    TString title_den = T::GetTitle() + TString(" den");
+    TDirectory::TContext ctx(nullptr);
+    mHistoNum = new T(name_num, title_num, 10, 0, 10, 10, 0, 10);
+    mHistoDen = new T(name_den, title_den, 1, -1, 1, 1, -1, 1);
+  }
+  copymerge.Copy(*this);
```
knopers8, line 53: why allocate these two histograms if, as I assume, they are overwritten by the `Copy` soon after?

knopers8 (commented):
thank you! could please have a look at my minor comments?

Modules/Common/include/Common/TH1Ratio.inl
```diff
@@ -19,18 +19,38 @@
...
 {
 
+template<class T>
+TH1Ratio<T>::TH1Ratio()
+  : T(),
+    o2::mergers::MergeInterface(),
+    mUniformScaling(true)
+{
+  {
+    TDirectory::TContext ctx(nullptr);
```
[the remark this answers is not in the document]
aferrero2707: This is "inspired" from [TEfficiency](https://root.cern.ch/doc/master/TEfficiency_8cxx_source.html).
I guess this is done because the numerator and denominator histograms are saved together with the container object, and therefore should not be also written independently.
I can write a line of comment with the link.

Modules/Common/include/Common/TH2Ratio.inl
aferrero2707: Because the `Copy()` function assumes that the histograms are already initialized. There is maybe a little overhead in creating dummy histograms and then re-initializing them properly in the `Copy()` function, but on the other hand this minimizes the code duplication between the copy-constructor and the real `Copy()` function.

knopers8: I see, thanks for explaining!

Modules/Common/include/Common/TH1Ratio.inl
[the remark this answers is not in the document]
knopers8: Thanks!

PR #1910, opened 2023-08-02
Moving some inline helpers to source files

knopers8:
...for faster compilation, with negligible performance hit.

-j 32:
with PR
real    3m33.082s
user    94m38.529s
sys     6m34.454s
reference
real    3m33.666s
user    94m55.119s
sys     6m33.528s

### DIFF
--- a/Framework/CMakeLists.txt
+++ b/Framework/CMakeLists.txt
@@ -55,6 +55,8 @@ add_library(O2QualityControl
   src/TaskInterface.cxx
   src/UserCodeInterface.cxx
   src/RepositoryBenchmark.cxx
+  src/RepoPathUtils.cxx
+  src/stringUtils.cxx
   src/InfrastructureGenerator.cxx
   src/InfrastructureSpecReader.cxx
   src/Check.cxx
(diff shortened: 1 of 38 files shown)


knopers8:
Hi @TimoWilken, I see multiple QC PR checks failing with a circular dependency in O2:
```
## sw/BUILD/O2-latest/log
--
+ grep stage
+ cut -f1 -d:
+ cmake --build . -- -j32 install
ninja: error: dependency cycle: stage/lib64/G__O2DetectorsVertexing_rdict.pcm -> stage/lib64/libO2StrangenessTracking.so -> Detectors/StrangenessTracking/tracking/CMakeFiles/O2lib-StrangenessTracking.dir/src/StrangenessTracker.cxx.o -> cmake_object_order_depends_target_O2lib-StrangenessTracking -> Detectors/StrangenessTracking/tracking/G__O2StrangenessTracking.cxx -> stage/lib64/G__O2DetectorsVertexing_rdict.pcm
--
```
Are you aware of this? Would you know whom to contact asking for a fix? Thanks!

PR #1916, opened 2023-08-08
[QC-999] Set ccdb url also for Checks and Aggregators

Barthelemy:


knopers8 (changes requested):
I don't quite understand the choices here. Since you need `Aggregator` and `Check` to be aware of `conditionUrl`, why not add it to their corresponding config structures instead of making them aware of the config structures of the Runners? I would propose to add `conditionUrl` to `CheckConfig` and `AggregatorConfig` instead.

Barthelemy:
There was a reason but I have forgotten about it meanwhile.

knopers8 (approved):
thanks!

PR #1920, opened 2023-08-10
ITS FEE task rearrangement and new misconfiguration plot

nicolovalle:


nicolovalle:
@IsakovAD @iravasen 

In this PR:

- TrailerCount plot added: intended for shifters' layout, it counts number of closing packet TDTs (ROF/orbit in physics)
    - The payload is parsed and the identifier only of each GBT word is decoded. Doen't produce overhead in the local tests, but the feature can be disabled via json by setting "EnablePayloadParsing": "0"
    - The corresponding checker have been added. The expected number of ROF/orbit is set on json: "expectedROFperOrbit": "18"
- ITS Header Word is now read and the number of active lanes per feeid filled in the new ActiveLanes plot
- mIndexCheck and mIdCheck objects have been deleted. A single mDecodingCheck objest is used to spot decoding errors while reading DataFormat, DDW or IHW
- STFinfo histogram protected against overflow
- ProcessingTime2 plot introduced to be used when plots from different machines are merged. ProcessingTime has been keept for offline data replay but it is inconsistent when used at P2
- laneStatusOverviewFlagFAULT has been deleted. laneStatusOverviewFlagERROR now contains both Error and Faulty lanes: better logic for shifter's layout
- The Detector Words in RDH is now read only at the first and last page of each orbit. This reduces computation time and allows a precise estimate of the counts in time units
- The list of trigger types and the list of bits in the detector word is initialized in a separate structure and the histograms are automatically filled accordingly

IsakovAD (approved):

PR #1921, opened 2023-08-10
[QC-996] policy multiple_per_run can delete first and last

Barthelemy:


Framework/script/RepoCleaner/qcrepocleaner/rules/multiple_per_run.py
```diff
@@ -85,7 +97,29 @@ def process(ccdb: Ccdb, object_path: str, delay: int,  from_timestamp: int, to_t
             logger.debug(f"     not in the allowed period, skip this bucket")
             preservation_list.extend(run_versions)
         else:
-            logger.debug(f"     not in the grace period")
+            logger.debug(f"    not in the grace period")
+
+            if delete_first_last:
```
knopers8, line 102: I don't quite understand why we only need the flag for the 2nd object since SOR, but not the second from the last. Won't we delete the last object each time we run the policy again?

Barthelemy: I don't think so because we work on the run as a whole. If we delete the first then we delete the last, if we don't delete the first we don't delete the last. 
The mark on the second is a mark we use for the whole run.

Framework/script/RepoCleaner/qcrepocleaner/rules/multiple_per_run.py
```diff
@@ -85,7 +97,29 @@ def process(ccdb: Ccdb, object_path: str, delay: int,  from_timestamp: int, to_t
...
+                run_versions.sort(key=lambda x: x.createdAt)
+                # Get flag cleaner_2nd from first object (if there)
+                cleaner_2nd = "cleaner_2nd" in run_versions[0].metadata
+                if cleaner_2nd or len(run_versions) < 4:
+                    logger.debug(f"        first version has flag cleaner_2nd or there are less than 4 version, "
+                                 f"we continue to next bucket")
+                    preservation_list.extend(run_versions)
+                    continue
+                # flag second with `cleaner_2nd`
+                ccdb.updateValidity(run_versions[1], run_versions[1].validFrom, run_versions[1].validTo,
```
Barthelemy, line 113: @knopers8 I think that here we should not actually update the validity right ?

knopers8: now we probably never should update validity

Framework/script/RepoCleaner/qcrepocleaner/rules/multiple_per_run.py
knopers8: ah, ok, clear

Framework/script/RepoCleaner/qcrepocleaner/rules/multiple_per_run.py
Barthelemy: ah wait, my bad, this is actually not touching the validity but adds the flag.

Barthelemy: @knopers8 is it ok ? it is not touching the validity, it is just using this method to add the flag.

knopers8: yes, but perhaps we should have another method for doing this.

Barthelemy: I have created a dedicated method. Behind the scene we still have to specify the validity due to the way the CCDB rest api works.

knopers8: thank you

knopers8 (approved):

PR #1922, opened 2023-08-11
Adjust to new TRD raw stats

martenole:
Will require https://github.com/AliceO2Group/AliceO2/pull/11754 to be merged first

### DIFF
--- a/Modules/TRD/include/TRD/RawData.h
+++ b/Modules/TRD/include/TRD/RawData.h
@@ -56,7 +56,6 @@ class RawData final : public TaskInterface
   TH1F* mDataAcceptance = nullptr;
   TH2F* mDataVolumePerHalfChamber = nullptr;
   TH2F* mDataVolumePerSector = nullptr;
-  TH2F* mDataVolumePerHalfSectorCru = nullptr;
   TH1F* mTimeFrameTime = nullptr;
   TH1F* mTrackletParsingTime = nullptr;
   TH1F* mDigitParsingTime = nullptr;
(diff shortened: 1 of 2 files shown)


martenole:
Change of plans. Hopefully this can be merged first before the changes in O2

martenole:
Hi @Barthelemy or @knopers8 could you please check and merge, if OK for you? Not sure why the macOS-arm check is not starting since Friday. But I need to merge some changes in O2 after and then again in QC and would like to have this before the next sw update.
The goal is for the TRD raw task to subscribe only to some statistics instead of all TRD data to allow for disabling the sampling. But for that the statistics class needs to change and this requires these subsequent updates.
Thanks!

knopers8:
@TimoWilken, @ktf  apparently the macOS check is stuck since Friday. Could you please have a look?

martenole:
The FST error by the way is unrelated. The check was green before

knopers8:
> I think there is an issue with the mac builder in the DAQ lab which keeps rebooting. Can you check if it's running fine?

I don't have access to these, I'm trying to find anyone who has and isn't on holidays :/

ktf:
Indeed the fullSystem test is now green. Merging this.

knopers8:
I will bother you personally if this PR was failing on mac ;)

martenole:
Personally I prefer to do it this way instead of not accepting any changes to QC until someone fixes the builders, since that seems to take some time. There will be a follow up PR and a PR in O2 that I would like to have merged before the tag for the next release is done.

knopers8:
I agree and I just did the same with other PRs.

ktf:
You know where to find me. ;-)

PR #1923, opened 2023-08-14
Rework of GenericHistogramCheck and addition of MetaData.

makor:


Modules/TPC/src/GenericHistogramCheck.cxx
```diff
@@ -195,36 +248,43 @@ void GenericHistogramCheck::beautify(std::shared_ptr<MonitorObject> mo, Quality
...
   double yText2 = 0;
-  auto h = dynamic_cast<TH1D*>((mo->getObject()));
+
   if (mHistDimension == 1) {
-    TLine* lineX = new TLine(mMeanX, h->GetMinimum() * 1.1, mMeanX, h->GetMaximum() * 1.1);
-    TLine* lineXEV = new TLine(mExpectedValueX, h->GetMinimum() * 1.1, mExpectedValueX, h->GetMaximum() * 1.1);
+
+    auto h1 = dynamic_cast<TH1F*>((mo->getObject()));
+    if (!h1) {
+      ILOG(Warning, Support) << "h1 not found in D= 1" << ENDM;
```
wiechula, line 256: I guess one should return here, otherwise it will crash accessing nullptr.
```suggestion
      ILOG(Warning, Support) << "h1 not found in D= 1" << ENDM;
      return;
```

Modules/TPC/src/GenericHistogramCheck.cxx
```diff
@@ -195,36 +248,43 @@ void GenericHistogramCheck::beautify(std::shared_ptr<MonitorObject> mo, Quality
...
-    xText = h->GetXaxis()->GetXmin() + std::abs(h->GetXaxis()->GetXmax() - h->GetXaxis()->GetXmin()) * 0.01;
-    yText = h->GetYaxis()->GetXmax() * 0.9;
-    // if we need a second line, move it 5% down
-    yText2 = h->GetYaxis()->GetXmax() * 0.9 - std::abs(h->GetYaxis()->GetXmax() - h->GetYaxis()->GetXmin()) * 0.05;
-    if (!h) {
-
-      ILOG(Warning, Support) << "h not found in D= 2" << ENDM;
+    auto h2 = dynamic_cast<TH2F*>((mo->getObject()));
+    if (!h2) {
+      ILOG(Warning, Support) << "h2 not found in D= 2" << ENDM;
```
wiechula, line 278: Same here:
```suggestion
      ILOG(Warning, Support) << "h2 not found in D= 2" << ENDM;
      return;
```

wiechula (changes requested):

makor:
The changes are now included, thanks for spotting them!

knopers8:
I was told that @wiechula  is on holidays, merging so it gets into the release before the software freeze, as it seems that @makor addressed the first round of Jens' comments.

PR #1926, opened 2023-08-15
[EMCAL-524, EMCAL-530] Adjustable multiplicity ranges

mfasDa:
Mutliplicity ranges can now be configured via the task
parameters for the cell task and the cluster task.
Needed in order to optimise histogram binning for pp
and PbPb runs which have very different ranges.


mfasDa:
Ideally the beam type would be available in initialise, both in the synchronous and the asynchronous QC ...

PR #1927, opened 2023-08-15
variables consistency

maroozm:
- Removing unused vars and naming vars as per json

### DIFF
--- a/Modules/TRD/TRDQC.json
+++ b/Modules/TRD/TRDQC.json
@@ -44,12 +44,6 @@
           "type": "direct",
           "query": "digits:TRD/DIGITS;tracklets:TRD/TRACKLETS;triggers:TRD/TRKTRGRD;rawstats:TRD/RAWSTATS"
         },
-        "taskParameters": {
-          "peakregionstart": "7.0",
-          "peakregionend": "20.0",
-          "pulseheightpeaklower": "1.0",
-          "pulseheightpeakupper": "5.0"
-        },
         "location": "remote",
         "saveObjectsToFile": "qcrootobjects.root"
       },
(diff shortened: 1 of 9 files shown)


Modules/TRD/TRDQC.json
```diff
@@ -128,10 +128,10 @@
...
         "taskParameters": {
           "trackType": "0xf",
-          "peakregionstart": "7.0",
-          "peakregionend": "20.0",
-          "pulseheightpeaklower": "1.0",
-          "pulseheightpeakupper": "5.0"
+          "driftRegionStart": "7.0",
+          "driftRegionEnd": "20.0",
+          "peakRegionStart": "1.0",
+          "peakRegionEnd": "5.0"
```
martenole, line 134: could you please also remove the obsolete `taskParameters` from all other tasks please? Looks like they were copy-pasted everywhere without need

Modules/TRD/src/PulseHeightTrackMatch.cxx
```diff
@@ -28,23 +30,13 @@
...
 {
 }
 
-void PulseHeightTrackMatch::retrieveCCDBSettings()
-{
-  if (auto param = mCustomParameters.find("ccdbtimestamp"); param != mCustomParameters.end()) {
-    mTimestamp = std::stol(mCustomParameters["ccdbtimestamp"]);
-    ILOG(Debug, Support) << "configure() : using ccdbtimestamp = " << mTimestamp << ENDM;
-  } else {
-    mTimestamp = o2::ccdb::getCurrentTimestamp();
```
martenole, line 43: please remove `mTimestamp` also from the header

martenole (changes requested):

Modules/TRD/TRDQC.json
maroozm: thanks. you mean from the TrackletsTask.cxx, DigitsTask.cxx, RawData.cxx?? or in the json only??

martenole: I think its only in the json, right? I'd like the TRDQC.json to be the single file which we can take as reference to configure the QC for production. But please check if there are spurious parameters also in the other tasks which are not used at all. For the DigitsTask for example maybe the regions are also specified for the PH plot based on digits only, I don't know

maroozm: thanks. yes, thats what i was thinking, because there are some params in tasks not used. I will try to make all of them consistent ( with the json as well).

maroozm: hi. removed some params from other tasks as well

martenole (approved):
Looks good now, thanks!

martenole:
Could this be merged please?

PR #1931, opened 2023-08-18
Fix retrieving lhc period on EPNs

knopers8:


knopers8:
needed for a patch release, compiles on my machine, merging.

PR #1937, opened 2023-08-22
Delete a prehistoric config file

knopers8:


Barthelemy (approved):
Should go to the museum

Barthelemy:
@knopers8 you should remove it also from the CMakeLists

PR #1939, opened 2023-08-22
[EMCAL-1037] Task for EMCAL BC monitoring

mfasDa:
Monitoring BC distributions for
- EMCAL readout (physics triggers only)
- CTP digits for all EMCAL triggers (including MB readout trigger) Needs CTP readout in run

### DIFF
--- a/Modules/EMCAL/src/BCTask.cxx
+++ b/Modules/EMCAL/src/BCTask.cxx
@@ -1,0 +1,197 @@
+// Copyright 2019-2020 CERN and copyright holders of ALICE O2.
+// See https://alice-o2.web.cern.ch/copyright for details of the copyright holders.
+// All rights not expressly granted are reserved.
+//
+// This software is distributed under the terms of the GNU General Public
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
+// In applying this license CERN does not waive the privileges and immunities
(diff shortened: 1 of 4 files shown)


mfasDa:
Must be a dedicated task for two reasons:
- Requires CTP readout, which is not guaranteed in all runs (in particular not in stand-alone runs)
- Should not be downsampled, processing of CTP digits and trigger records is fast

mfasDa:
Ready to merge

mfasDa:
CI errors unrelated

mfasDa:
CI errors unrelated

Modules/EMCAL/src/BCTask.cxx
```diff
@@ -0,0 +1,197 @@
...
+#include <TCanvas.h>
+#include <TH1.h>
+
+#include <boost/algorithm/string/case_conv.hpp>
+
+#include "CommonConstants/LHCConstants.h"
+#include "CommonConstants/Triggers.h"
+#include "DataFormatsCTP/Configuration.h"
+#include "DataFormatsCTP/Digits.h"
+#include "DataFormatsEMCAL/TriggerRecord.h"
```
knopers8, line 26: since these come from O2, `<>` brackets could be used instead.

knopers8 (approved):

PR #1945, opened 2023-08-24
[QC-1006] Add method to interface to custom parameters

Barthelemy:
@aferrero2707


knopers8 (approved):
this call will be ambiguous, but it's probably OK, the compiler should complain:
```
atOrDefaultValue("key", "default", {});
```

Barthelemy:
ah indeed

PR #1946, opened 2023-08-24
ITS FEETask payload parsing optimization

nicolovalle:
Now the parsing of payload to count the GBT word identifiers can be triggered with a custom Orbit and TF periodicity. 
In addition, the ITS Header Word decoding can be disabled.
@IsakovAD @iravasen


IsakovAD (approved):
All is fine!

PR #1952, opened 2023-08-28
Give the possibility to save the debug messages to the discardFile.

Barthelemy:


Framework/src/TaskRunner.cxx
```diff
@@ -455,7 +455,7 @@ void TaskRunner::startOfActivity()
 
   // register ourselves to the BK
   if (gSystem->Getenv("O2_QC_REGISTER_IN_BK")) { // until we are sure it works, we have to turn it on
-    ILOG(Debug, Devel) << "Registering taskRunner to BookKeeping" << ENDM;
+    ILOG(Info, Devel) << "Registering taskRunner to BookKeeping" << ENDM;
```
knopers8, line 458: i recommend using `git add -p` to individually accept each change you want bring to the commit ;)

knopers8 (approved):

Framework/src/TaskRunner.cxx
Barthelemy: thank you for your suggestion :-D

PR #1953, opened 2023-08-28
[QC-1010] Make activity available in aggregator

Barthelemy:
And add doc how to access it from other places.

### DIFF
--- a/Framework/src/Aggregator.cxx
+++ b/Framework/src/Aggregator.cxx
@@ -224,4 +224,13 @@ AggregatorConfig Aggregator::extractConfig(const core::CommonSpec& commonSpec, c
   };
 }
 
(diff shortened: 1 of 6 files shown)


Barthelemy:
@aferrero2707

Framework/src/Aggregator.cxx
```diff
@@ -224,4 +224,13 @@ AggregatorConfig Aggregator::extractConfig(const core::CommonSpec& commonSpec, c
   };
 }
 
+void Aggregator::setActivity(std::shared_ptr<core::Activity> activity)
```
knopers8, line 227: You are not using this method anywhere, this PR will not make any difference as far as i can tell

knopers8 (changes requested):

PR #1955, opened 2023-08-29
Fix some warnings in TRD

Barthelemy:


martenole (approved):
Thanks for the fix. Please @deependra170598 check also when you are compiling your QC changes that there are not any new warnings introduced by your changes

knopers8 (approved):

PR #1963, opened 2023-08-31
Include what you use: ptree.h

ktf:
Framework will soon reduce exposing ptree.h to the bare minimum, due to interference with ROOT.


knopers8 (approved):

Barthelemy:
we will merge it as soon as macOS-arm passes ;-)

Barthelemy (approved):

PR #1964, opened 2023-09-01
reduce Nobjects in TOF digit task

noferini:
Hi @Barthelemy , @ercolessi @njacazio ,
I reduced the number of objects in the digit task.
I grouped 72xTProfile in 1xTProfile2D


njacazio (approved):
Thanks Francesco

noferini:
Hi @Barthelemy , if you agree I would not wait for the macOS-arm. It is pending from several days and I don't think we need it for this PR. My feeling is that this is check is very slow.

Barthelemy (approved):

Barthelemy:
totally agree, I force merge and it goes to the new release. Thanks!

PR #1965, opened 2023-09-01
[QC-1012] Allow TrendingTaskITSFhr to trend slightly desynchronized objects

knopers8:
The FHR Task generates objects specific to particular detector layers. Since detector layers are assigned to different subsets of FLPs, the merged objects do not have exactly the same validity, as FLPs do not have QC cycles ideally in sync. Thus we have look for the objects a bit behind and forward from the trigger timestamp as well.

This introduces an configuration parameter `maxObjectTimeShiftMs` to allow for certain shift of QC objects coming from the trended task. Normally it should correspond to the cycle length of the latter.


knopers8:
tested with some improvised config, seems to work as expected.

Barthelemy (approved):

knopers8:
@iravasen please shout if you do not agree with this solution.

PR #1966, opened 2023-09-01
TPC ClusterVisualizer should use the provided trigger timestamp

knopers8:
to retrieve ClusterData. Otherwise it matches a wrong object by using the current time.


knopers8:
currently breaking Clusters and RawDigits plots in production, merging to have a quick patch release.

PR #1970, opened 2023-09-04
correction to check function

deependra170598:
Correcting check function but still investigating trend problem

### DIFF
--- a/Modules/TRD/src/TrackletPerTriggerCheck.cxx
+++ b/Modules/TRD/src/TrackletPerTriggerCheck.cxx
@@ -18,6 +18,7 @@
 #include "QualityControl/MonitorObject.h"
 #include "QualityControl/Quality.h"
 #include "QualityControl/QcInfoLogger.h"
+#include "Common/Utils.h"
 // ROOT
 #include <TH1.h>
 #include <TString.h>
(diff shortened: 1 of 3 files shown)


deependra170598:
Hi @knopers8 and @martenole, May you review this PR?

Modules/TRD/src/TrackletPerTriggerCheck.cxx
```diff
@@ -102,10 +102,14 @@ Quality TrackletPerTriggerCheck::check(std::map<std::string, std::shared_ptr<Mon
 
       // applying check
       float MeanTracletPertrigger = h->GetMean();
-      if (MeanTracletPertrigger > mDesiredMeanRegion.second && MeanTracletPertrigger < mDesiredMeanRegion.first) {
+      if (MeanTracletPertrigger < mDesiredMeanRegion.second && MeanTracletPertrigger > mDesiredMeanRegion.first) {
```
martenole, line 105: to avoid these kind of typos wouldn't it be better to use two variables in your checker class, e.g. `mMeanRegionLow` and `mMeanRegionHigh` or something like that? And then for reading the parameters from json file I'd prefer you use the common `getFromConfig` as it is done for example in https://github.com/AliceO2Group/QualityControl/blob/a0dc3f4234734b36d0b3dc9ce54cb3a3d305e591/Modules/TRD/src/DigitsTask.cxx#L408-L415. You'll need to include `Common/Utils.h` for this

deependra170598: I updated this PR with the above suggestions.

Modules/TRD/src/TrackletPerTriggerCheck.cxx
```diff
@@ -28,48 +29,29 @@
...
-  if (auto param = mCustomParameters.find("ccdbtimestamp"); param != mCustomParameters.end()) {
-    mTimeStamp = std::stol(mCustomParameters["ccdbtimestamp"]);
-    ILOG(Debug, Support) << "configure() : using ccdbtimestamp = " << mTimeStamp << ENDM;
-  } else {
-    mTimeStamp = o2::ccdb::getCurrentTimestamp();
-    ILOG(Debug, Support) << "configure() : using default timestam of now = " << mTimeStamp << ENDM;
-  }
+  auto mTimeStamp = getFromConfig<long>(mCustomParameters, "ccdbtimestamp", o2::ccdb::getCurrentTimestamp());
+  ILOG(Debug, Support) << "using timestamp = " << mTimeStamp << ENDM;
   auto& mgr = o2::ccdb::BasicCCDBManager::instance();
```
martenole, line 42: what are you using the BasicCCDBManager for?

deependra170598: I try to explain how I understood the concept of BasicCCDBManager. Although I do not fully understand.
1. instance() method of class BasicCCDBManager : public CCDBManagerInstance initialise the class with ccdb server. 
2. Then setTimestamp() method is used for setting timestamp manually if required to be given from config file otherwise current time stamp is used. It set timestamp cache for all queries (line no. 75 of BasicCCDBManager.h).
3. I see that BasicCCDBManager class itself by default set current timestamp (line no. 189 of BasicCCDBManager.h), so if we remove it from checker class it run without any problem. 
4. In lack of knowledge, I kept these line because I did not know its consequences.
 I will do how you suggest.

deependra170598: I had in my mind that in background of workflow, It set the timestamp of QO object generated by checker class. And if a specific object is needed by any postprocessing in future, It can use that timestamp to query that QO object. Although this thought needs validation.

martenole: Hi Deependra, I would simply remove the CCDB manager completely from your task, because to me it does not look like you are using it. In some cases people are requesting CCDB objects in their tasks and use the CCDB manager for it, but I don't think you need any calibration objects for your check so I would get rid of it to keep the code as clean as possible. Later we will need some CCDB objects for more advanced checks, but also there not the CCDB manager should be used anymore, but the more modern DPL CCDB fetcher. But we can talk about this when we get there :)

deependra170598: Hi Ole, I removed that part.

martenole (approved):

deependra170598:
Hi @knopers8 , May you look at it if any review is required from your side?

knopers8:
macOS CI errors seem legit:
```
/Volumes/build/alice-ci-workdir/qualitycontrol-o2/sw/SOURCES/QualityControl/1970/0/Modules/TRD/src/TrackletPerTriggerCheck.cxx:69:119: error: format specifies type 'long' but the argument has type 'int' [-Werror,-Wformat]
        msg1->AddText(TString::Format("Hist Can't be ignored. Stat is enough. Entries: %d > Threshold: %ld", Entries, mStatThreshold));
                                                                                                       ~~~            ^~~~~~~~~~~~~~
                                                                                                       %d
--
        msg1->AddText(TString::Format("Hist Can't be ignored. Stat is enough. Entries: %d > Threshold: %ld", Entries, mStatThreshold));
                                                                                                       ~~~            ^~~~~~~~~~~~~~
                                                                                                       %d
/Volumes/build/alice-ci-workdir/qualitycontrol-o2/sw/SOURCES/QualityControl/1970/0/Modules/TRD/src/TrackletPerTriggerCheck.cxx:72:114: error: format specifies type 'long' but the argument has type 'int' [-Werror,-Wformat]
        msg1->AddText(TString::Format("Hist Can be ignored. Stat is low. Entries: %d < Threshold: %ld", Entries, mStatThreshold));
                                                                                                  ~~~            
...
```

martenole:
Thanks, should be good to merge now

knopers8 (approved):

PR #1974, opened 2023-09-06
[QC-1023] Legend customizations in SlicedTrendingTask

mfasDa:
Customizable settings:
- Number of columns
- Text size

### DIFF
--- a/Framework/src/SliceTrendingTask.cxx
+++ b/Framework/src/SliceTrendingTask.cxx
@@ -179,7 +179,8 @@ void SliceTrendingTask::generatePlots()
     c->SetName(plot.name.c_str());
     c->SetTitle(plot.title.c_str());
 
-    drawCanvasMO(c, plot.varexp, plot.name, plot.option, plot.graphErrors, mAxisDivision[varName]);
+    TitleSettings titlesettings{ plot.legendObservableX, plot.legendObservableY, plot.legendUnitX, plot.legendUnitY, plot.legendCentmodeX, plot.legendCentmodeY };
+    drawCanvasMO(c, plot.varexp, plot.name, plot.option, plot.graphErrors, mAxisDivision[varName], titlesettings);
 
     int NumberPlots = 1;
     if (plot.varexp.find(":time") != std::string::npos || plot.varexp.find(":run") != std::string::npos) { // we plot vs time, multiple plots on canvas possible
@@ -241,7 +242,7 @@ void SliceTrendingTask::generatePlots()
 } // void SliceTrendingTask::generatePlots()
 
 void SliceTrendingTask::drawCanvasMO(TCanvas* thisCanvas, const std::string& var,
-                                     const std::string& name, const std::string& opt, const std::string& err, const std::vector<std::vector<float>>& axis)
+                                     const std::string& name, const std::string& opt, const std::string& err, const std::vector<std::vector<float>>& axis, const TitleSettings& titlesettings)
 {
   // Determine the order of the plot (1 - histo, 2 - graph, ...)
   const size_t plotOrder = std::count(var.begin(), var.end(), ':') + 1;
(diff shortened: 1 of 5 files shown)


mfasDa:
@mlesch please review!

With the hard coded legend settings the legend can become unreadable - the text is much too small. At least the number of columns and the text size must be configurable.

Barthelemy (approved):

mfasDa:
Without customisation the legend looks like
 
CellTrending_withoutCustomLegend

where it is even impossible to read anything, with the customisation at least the text is readable but not yet fully meaningful:

CellTrending_withCustomLegend

I would like to add more customisations so that the title can be expressed in a more meaningful way.  What it should say in the example is:

0.5 GeV < E_{cell} < 5 GeV
5.0 GeV < E_{cell} < 10 GeV
10.0 GeV < E_{cell} < 50 GeV

I want to do this with optional parameters legendObservableX, legend observableY, legendUnitX and legendUnitY. Still to be checked is a case where we have integer-like binning where the range reflects only a single value (use case supermodule ID in our case).

mfasDa:
I added some options to manipulate the histogram titles used in the legend in order to make the legend entries more user-friendly. Instead of `RangeX [..., ....]` the title can appear as `... <= X < ...`. For this the following variables are added to the configuration:
- legendObservableX
- legendObservableY
- legendUnitX
- legendUnitY
Units are optional. The ranges are parsed from the ranges in the selection string. In case the observable string in either direction is `None` the range string is not shown. In case the observable string is empty the default selection string is shown (old behaviour). Using these customizations the following plot is created:

Screenshot 2023-09-07 at 16 23 16

which is a lot more readable than with the default ranges. A leftover problem are indexed observables. i.e. for module/sector indices we don't want to used `0.5 < moduleID < 1.5` but rather `Module 1`. In order to achieve this the settings 
- legendCentmodeX
- legendCentmodeY
are introduced, in which the mean of the range is used. With this setting the following plot is created:

Screenshot 2023-09-07 at 16 26 35

which is again better readable than the version above.

Framework/src/SliceTrendingTask.cxx
```diff
@@ -543,3 +546,87 @@ void SliceTrendingTask::beautifyGraph(T& graph, const SliceTrendingTaskConfig::P
...
   }
 }
+
+void SliceTrendingTask::beautifyLegend(TLegend* leg, const SliceTrendingTaskConfig::Plot& plotconfig, TCanvas* canv)
+{
+  int ncolums = 2;
+  try {
+    ncolums = std::stoi(plotconfig.legendNColums);
+  } catch (...) {
+    LOG(error) << "key legNColums must be integer" << ENDM;
```
Barthelemy, line 556: in the framework we use the infologger, could you switch to `ILOG` ?

Framework/src/SliceTrendingTask.cxx
```diff
@@ -543,3 +546,87 @@ void SliceTrendingTask::beautifyGraph(T& graph, const SliceTrendingTaskConfig::P
...
+  }
+  leg->SetNColumns(ncolums);
+
+  double textsize = 2.0;
+  try {
+    textsize = std::stod(plotconfig.legendTextSize);
+  } catch (...) {
+    LOG(error) << "key legendTextSize must be double" << ENDM;
```
Barthelemy, line 564: same

Framework/src/SliceTrendingTask.cxx
mfasDa: Done

Framework/src/SliceTrendingTask.cxx
mfasDa: Done

Barthelemy:
@mlesch could you also review this PR ? thank you

Barthelemy:
@mlesch I will merge tomorrow

mlesch:
@mfasDa ,

Apologies for the long answer time, I was absent for the past couple of weeks. I was aware of the shortcomings of the legend but I never managed to find the time to improve this, this is really great work! Thanks a lot for this improvement! 

I went through your PR and did not find anything suspicious. I also saw that it is backwards compatible so it should not disturb any current json setups that use the SliceTrendingTask. 

That being said, @Barthelemy you can merge. Thanks!

PR #1977, opened 2023-09-08
[HMP] Raw checker

gvolpe79:


Barthelemy (changes requested):
just a few comments that you could consider.

Barthelemy (approved):

PR #1978, opened 2023-09-08
ZDC: bug fixes for ZDCRecDataCheck

aferrero2707:
* corrected name of `h_summmary_ADC` in `check()` method
* added reset of TDC parameters vector similar to the ADC one. Without this reset, the check performed [here](https://github.com/AliceO2Group/QualityControl/blob/3584e9b886b8ef16166336f11f298c976ba4df45/Modules/ZDC/src/ZDCRecDataCheck.cxx#L94) always fails starting from the second cycle, leading to a `Null` quality output.


aferrero2707:
@knopers8 @cpuggion84 the changes in this PR are minimal but are essential in order to get proper results from `ZDCRecDataCheck`.

One thing which is still not clear to me is why the `ZDCRecDataCheck::init()` method is called at each cycle in `ZDCRecDataCheck::check()` instead of only once in `ZDCRecDataCheck::configure()`.

Barthelemy (approved):

PR #1980, opened 2023-09-12
Don't rely on O2 for std pollution

Barthelemy:


vkucera:
Thanks @Barthelemy !

Barthelemy:
@mfasDa O2 had a `using namespace std` leaking to us and it was hiding the problem. @vkucera fixed it but we have to fix it here first.

Barthelemy:
@vkucera the release is here: https://github.com/alisw/alidist/pull/5133

vkucera:
Perfect! Thanks a lot @Barthelemy !

PR #1981, opened 2023-09-12
Update TRD tracking task histograms

martenole:
Hi @Barthelemy if possible these minor improvements would be nice to have also for HI. It adds a title to a plot shown to the QC shifter and one new plot which I would also like to add to the shifters layout (average number of TRD tracklets per track as function of eta-phi).


Barthelemy:
@martenole Please contact Filippo Costa if you want something installed at P2.

Barthelemy (approved):

PR #1982, opened 2023-09-12
[QC-1028] Optionally disable the last, incomplete, cycle

Barthelemy:


Barthelemy:
@aferrero2707 @knopers8 Tested on my machine successfully.

PR #1987, opened 2023-09-15
[O2-4122, EMCAL-1037] Require CCDB manager to not fatal when CTP configuration is not available

mfasDa:
Handling of missing CTP configuration should be
dedicated to the task, which has a mode for
missing CTP configuration (BCs from EMCAL only).
The BasicCCDBManager must not fatal when the
CTP configuration is not available.

### DIFF
--- a/Modules/EMCAL/src/BCTask.cxx
+++ b/Modules/EMCAL/src/BCTask.cxx
@@ -19,6 +19,7 @@
 
 #include <boost/algorithm/string/case_conv.hpp>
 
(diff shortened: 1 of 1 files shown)


mfasDa:
It is understood that this is only a temporary solution to prevent the task from crashing until we can reliably access the CTP configuration also in replay runs. The line will be removed until O2-4122 is fully resolved.

mfasDa:
Will be replaced by DPL CCDB backend

PR #1988, opened 2023-09-15
[WIP][O2-4122, EMCAL-1037] Move to DPL CCDB Backend

mfasDa:
Use DPL CCDB Backend for accessing the trigger
configuration.

### DIFF
--- a/Modules/EMCAL/include/EMCAL/BCTask.h
+++ b/Modules/EMCAL/include/EMCAL/BCTask.h
@@ -25,6 +25,11 @@
 
 class TH1F;
 
+namespace o2::ctp
+{
+class CTPConfiguration;
+}
+
 using namespace o2::quality_control::core;
 
 namespace o2::quality_control_modules::emcal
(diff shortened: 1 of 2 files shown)


mfasDa:
Set to WIP because query of `ccdb-run-dependent` currently ends in fatal as all attributes in DataDescriptorQueryBuilder are parsed as string while bool is expected, leading to a type mismatch fatal.

mfasDa:
Do not merge yet! In the current state the feature is unusable for us. Run-dependent queries are essential, the task will produce nonsense otherwise.

mfasDa:
Needs https://github.com/AliceO2Group/AliceO2/pull/11917

Barthelemy:
@mfasDa What should we do with this PR?

mfasDa:
@Barthelemy I hope at some point fix the underlying problem. But we cannot merge it yet as the task doesn't get the CCDB objects in case a data sampling policy is applied, which we have to do. The CCDB backend seems to work only for direct data sources.

PR #1993, opened 2023-09-18
[O2-4122, EMCAL-1037] Require CCDB manager to not fatal when CTP configuration is not available

mfasDa:
Handling of missing CTP configuration should be
dedicated to the task, which has a mode for
missing CTP configuration (BCs from EMCAL only).
The BasicCCDBManager must not fatal when the
CTP configuration is not available.

### DIFF
--- a/Modules/EMCAL/src/BCTask.cxx
+++ b/Modules/EMCAL/src/BCTask.cxx
@@ -19,6 +19,7 @@
 
 #include <boost/algorithm/string/case_conv.hpp>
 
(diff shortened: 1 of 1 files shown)


mfasDa:
I re-open https://github.com/AliceO2Group/QualityControl/pull/1987 as temporary solution until the issues blocking https://github.com/AliceO2Group/QualityControl/pull/1988 are fixed.

knopers8 (approved):

PR #1995, opened 2023-09-21
[QC-1034] finaliseCCDB accessible from TaskInterface

knopers8:


knopers8:
@mfasDa regarding a helper to associate a binding with originally sampled inputs, it is more complicated than I initially expected and I did not find yet an elegant way (or even not terrible) to provide you with this information, since it breaks a few abstraction layers in QC. I have to think a bit more...

Barthelemy (approved):

PR #2000, opened 2023-09-27
Make some output debug instead of info

Barthelemy:
Because it creates quite some logs in the infologger. 

@aferrero2707 what do you think ?


knopers8 (approved):

aferrero2707:
@Barthelemy absolutely! We will remove those messages in a future PR. Thanks!

PR #2002, opened 2023-09-28
[QC-1015] Do not trend slices if some objects are unavailable

knopers8:
We saw several crashes at P2 when one of the requested objects was not available. In such case, `dataRetrieveVector->at(p)` calls would throw, since the vector was not filled all requested data points. I think the most sane thing to do in such case, is to avoid adding a new row in the TTree, otherwise we would have to fill it partially with 0, which might appear as very misleading in some of the produced plots.

### DIFF
--- a/Framework/src/SliceTrendingTask.cxx
+++ b/Framework/src/SliceTrendingTask.cxx
@@ -140,6 +140,9 @@ void SliceTrendingTask::trendValues(const Trigger& t,
       if (obj) {
         mReductors[dataSource.name]->update(obj, *mSources[dataSource.name],
                                             dataSource.axisDivision, mNumberPads[dataSource.name]);
+      } else {
+        ILOG(Error, Support) << "Some objects could not be retrieved, will skip this trending cycle" << ENDM;
+        return;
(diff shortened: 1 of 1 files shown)


knopers8:
What do you think @mlesch?

Barthelemy (approved):

mlesch:
@knopers8 I think it's fine. The only thing I'm wondering about is the following case: What happens if an object is available, then for one or two cycle there's an issue, and then it's available again. This one object then has less entries than the other objects in the TTree. Could this give any issues?

knopers8:
If one of the objects is not available we do not add a new row in the tree at all, so we do not risk getting out of sync between the branches.

An alternative to this would be to use zeroes for objects failed to retrieve, but this could be very misleading in some plots. I don't know if TTree has any way to have optional values...

I think this is anyway an improvement, because previously we would just crash in such cases.

mlesch:
Yes, it's an improvement for sure, there I fully agree. I would say that we can leave it like this and if I have some time I can also think about some further improvement in the future.

PR #2003, opened 2023-09-28
[QC-1017] Circular dependency if Aggregator and Check have the same n…

Barthelemy:
…ames

Took the opportunity to rename and clean up a few things.

### DIFF
--- a/Framework/include/QualityControl/InfrastructureGenerator.h
+++ b/Framework/include/QualityControl/InfrastructureGenerator.h
@@ -179,10 +179,10 @@ class InfrastructureGenerator
   static void printVersion();
 
   /**
-   * If an aggregator has the same name as a check, we throw a runtime_error.
+   * If any aggregator has the same name as one of the checks, we throw a runtime_error.
    * @param infrastructureSpec
    */
-  static void throwIfNameClash(const InfrastructureSpec& infrastructureSpec);
+  static void throwIfAggNamesClashCheckNames(const InfrastructureSpec& infrastructureSpec);
 
  private:
   // Dedicated methods for creating each QC component to hide implementation details.
--- a/Framework/include/QualityControl/InfrastructureGenerator.h
+++ b/Framework/include/QualityControl/InfrastructureGenerator.h
@@ -178,6 +178,12 @@ class InfrastructureGenerator
 
   static void printVersion();
 
+  /**
+   * If an aggregator has the same name as a check, we throw a runtime_error.
+   * @param infrastructureSpec
+   */
+  static void throwIfNameClash(const InfrastructureSpec& infrastructureSpec);
+
  private:
   // Dedicated methods for creating each QC component to hide implementation details.
 
--- a/Framework/src/InfrastructureGenerator.cxx
+++ b/Framework/src/InfrastructureGenerator.cxx
@@ -731,6 +731,7 @@ void InfrastructureGenerator::generateCheckRunners(framework::WorkflowSpec& work
 
 void InfrastructureGenerator::throwIfNameClash(const InfrastructureSpec& infrastructureSpec)
 {
+  // TODO simplify with ranges when we'll start using c++20
   std::vector<std::string> allNames;
   allNames.reserve(infrastructureSpec.checks.size());
   std::transform(infrastructureSpec.checks.begin(), infrastructureSpec.checks.end(), std::back_inserter(allNames),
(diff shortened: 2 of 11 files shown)


Framework/include/QualityControl/InfrastructureGenerator.h
```diff
@@ -178,6 +178,12 @@ class InfrastructureGenerator
 
   static void printVersion();
 
+  /**
+   * If an aggregator has the same name as a check, we throw a runtime_error.
+   * @param infrastructureSpec
+   */
+  static void throwIfNameClash(const InfrastructureSpec& infrastructureSpec);
```
knopers8, line 185: The name and placement suggests that it checks for any name clashes. Could you please make it more specific or document so that it indeed is supposed to check any name clashes?

Framework/src/InfrastructureGenerator.cxx
```diff
@@ -729,18 +729,30 @@ void InfrastructureGenerator::generateCheckRunners(framework::WorkflowSpec& work
...
-  std::vector<framework::OutputSpec> checkRunnerOutputs;
-  for (const auto& checkSpec : infrastructureSpec.checks) {
-    checkRunnerOutputs.emplace_back(Check::createOutputSpec(checkSpec.checkName));
+  std::vector<std::string> allNames;
+  allNames.reserve(infrastructureSpec.checks.size());
+  std::transform(infrastructureSpec.checks.begin(), infrastructureSpec.checks.end(), std::back_inserter(allNames),
+                 [](const auto& check) { return check.checkName; } );
+  for(const auto& aggregator : infrastructureSpec.aggregators) {
+    if(std::count(allNames.begin(), allNames.end(), aggregator.aggregatorName)) {
+      ILOG(Error, Ops) << "The aggregator \"" << aggregator.aggregatorName << "\" has the same name as one of the Check. This is forbidden." << ENDM;
```
knopers8, line 740:
```suggestion
      ILOG(Fatal, Ops) << "The aggregator \"" << aggregator.aggregatorName << "\" has the same name as one of the Check. This is forbidden." << ENDM;
```
sounds like a fatal to me

Framework/src/InfrastructureGenerator.cxx
```diff
@@ -729,18 +729,30 @@ void InfrastructureGenerator::generateCheckRunners(framework::WorkflowSpec& work
   ILOG(Debug, Devel) << ENDM;
 }
 
-void InfrastructureGenerator::generateAggregator(WorkflowSpec& workflow, const InfrastructureSpec& infrastructureSpec)
+void InfrastructureGenerator::throwIfNameClash(const InfrastructureSpec& infrastructureSpec)
 {
-  std::vector<framework::OutputSpec> checkRunnerOutputs;
-  for (const auto& checkSpec : infrastructureSpec.checks) {
-    checkRunnerOutputs.emplace_back(Check::createOutputSpec(checkSpec.checkName));
+  std::vector<std::string> allNames;
```
knopers8, line 734:
```suggestion
  // TODO simplify with ranges when we'll start using c++20
  std::vector<std::string> allNames;
```


we can do sth like this:
```
std::vector<std::string> allNames = infrastructureSpec.checks 
                                   | std::views::transform([](const auto& check) { return check.checkName; });

for(const auto& aggregator : infrastructureSpec.aggregators) {
    if(std::ranges::count(allNames, aggregator.aggregatorName)) {
        ILOG(Error, Ops) << "The aggregator \"" << aggregator.aggregatorName << "\" has the same name as one of the Check. This is forbidden." << ENDM;
        throw std::runtime_error(string("aggregator has the same name as a check: ") + aggregator.aggregatorName);
    }
}
```

Framework/test/testThrowNameClash.json
```diff
@@ -0,0 +1,71 @@
...
+        "active": "true",
+        "className": "o2::quality_control_modules::skeleton::SkeletonTask",
+        "moduleName": "QcSkeleton",
+        "detectorName": "TST",
+        "cycleDurationSeconds": "10",
+        "maxNumberCycles": "-1",
+        "dataSource": {
+          "type": "dataSamplingPolicy",
+          "name": "tst-raw"
+        },
```
knopers8, line 23: just to minimize the config file and avoid rather unrelated issues break this test, can we get rid of sampling here and use direct data source instead?

Framework/include/QualityControl/InfrastructureGenerator.h
Barthelemy: yes, indeed. Done.

knopers8 (approved):

PR #2004, opened 2023-09-28
[EMCAL-1045] Skip Raw pages with header or trailer corruption

mfasDa:
Corruptions lead to infity loops as the position of the next payload
cannot be determined


Barthelemy:
the build error is legit

mfasDa:
Header included expilcitly

mfasDa:
CI error unrelated

PR #2006, opened 2023-09-29
[QC-1036] Fix merging invalid validity

knopers8:
As discovered, when a plot is empty it correctly gets invalid validity range (uint_max, 0). However, when we merge it with a correct non-empty object, the invalid range becomes the inverted new range of the valid object (0, uint_max), which is wrong. This PR makes invalid validity intervals be ignored when merging objects.


knopers8:
failure in o2-dataflow unrelated, others are passing, merging to create a patch release for async QC

PR #2009, opened 2023-10-11
[EMCAL-526] Protection against unexpected tower ID

mfasDa:
Unexpected tower IDs might be a result of a data
corruption, must be caught and discarded.


mfasDa:
Fixing https://ali-bookkeeping.cern.ch?page=log-detail&id=72630 - very rare data corruption

Barthelemy (approved):

PR #2011, opened 2023-10-12
TRD: Check on Trigger and timeframe without tracklets

deependra170598:
In this PR, BinContent of trackletsperevent and trackletspertimeframe at zeroth bin is observed and a warning as a text over the histogram is issue if there is any entry in the zeroth bin.

Also a check is applied on mean of trackletspertimeframe.


deependra170598:
Hi @martenole  , May you  review this PR?

deependra170598:
Hi @martenole , May you review new check: Modules/TRD/src/TrackletCountCheck.cxx ? If it looks fine I will remove older checks on Tracklet per trigger and timeframe.

Modules/TRD/include/TRD/TrackletCountCheck.h
```diff
@@ -0,0 +1,51 @@
...
+// This software is distributed under the terms of the GNU General Public
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+///
+/// \file   TrackletCountCheck.h
+/// \author My Name
```
martenole, line 14: remove or put meaningful info

Modules/TRD/include/TRD/TrackletCountCheck.h
```diff
@@ -0,0 +1,51 @@
...
+#ifndef QC_MODULE_TRD_TRDTRACKLETCOUNTCHECK_H
+#define QC_MODULE_TRD_TRDTRACKLETCOUNTCHECK_H
+
+#include "QualityControl/CheckInterface.h"
+
+namespace o2::quality_control_modules::trd
+{
+
+/// \brief  Example QC Check
+/// \author My Name
```
martenole, line 26: as above

Modules/TRD/src/TrackletCountCheck.cxx
```diff
@@ -0,0 +1,190 @@
...
+// This software is distributed under the terms of the GNU General Public
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+///
+/// \file   TrackletCountCheck.cxx
+/// \author My Name
```
martenole, line 14: as above

Modules/TRD/src/TrackletCountCheck.cxx
```diff
@@ -0,0 +1,190 @@
...
+  mStatThresholdPerTrigger = getFromConfig<int>(mCustomParameters, "StatThresholdPerTrigger", 1000);
+  ILOG(Debug, Support) << "using StatThreshold Per Trigger= " << mStatThresholdPerTrigger << ENDM;
+}
+
+Quality TrackletCountCheck::check(std::map<std::string, std::shared_ptr<MonitorObject>>* moMap)
+{
+  mResultPertrigger = Quality::Null;
+  mResultPerTimeFrame = Quality::Null;
+
+  for (auto& [moName, mo] : *moMap) {
```
martenole, line 60: is this the proposed way to iterate through the input? Out of curiosity what is `moName`?

Modules/TRD/src/TrackletCountCheck.cxx
```diff
@@ -0,0 +1,190 @@
...
+      // Warning about triggers without any tracklets
+      int UnderFlowTrackletsPerTrigger = h->GetBinContent(0);
+      if (UnderFlowTrackletsPerTrigger > 0.) {
+        msg1->AddText(TString::Format("Triggers without Tracklets: %d", UnderFlowTrackletsPerTrigger));
+      }
+
+      // applying check
+      float MeanTrackletPertrigger = h->GetMean();
+      if (MeanTrackletPertrigger > mThresholdMeanLowPerTrigger && MeanTrackletPertrigger < mThresholdMeanHighPerTrigger) {
+        TText* Checkmsg = msg1->AddText("Mean Per Trigger is found in bound region: ok");
```
martenole, line 94: maybe add to the text the mean (it its not anyhow visible from the statistics box of the histogram) and the boundaries?

Modules/TRD/src/TrackletCountCheck.cxx
```diff
@@ -0,0 +1,190 @@
...
+
+    (void)moName;
+    if (mo->getName() == "trackletsperevent") {
+      auto* h = dynamic_cast<TH1F*>(mo->getObject());
+      if (h == nullptr) {
+        // ILOG(Debug, Support) << "Requested Histogram type does not match with the Histogram in source" << ENDM;
+        continue;
+      }
+      // warning about statistics available
+      TPaveText* msg1 = new TPaveText(0.3, 0.7, 0.7, 0.9, "NDC"); // check option "br","NDC"
```
martenole, line 70: isn't this leaking memory?

martenole (commented):
Ok, looks in general good I would say. Could you please remove the two obsolete classes in this PR?

Modules/TRD/src/TrackletCountCheck.cxx
deependra170598: Script: o2-qc-module-configurator.sh generates the format of check or task class. In the generated check class, iteration over the monitor objects is done this way. moName is the relative path of the monitor object w.r.t. the Task directory in QCG. For example: moName =TrackletsTask/trackletsperevent

Modules/TRD/src/TrackletCountCheck.cxx
deependra170598: I don't know how to check memory leaks. If you help me how to check it would be good. I saw this way of writing text over hist in some other modules' class.

Modules/TRD/src/TrackletCountCheck.cxx
martenole: Ok, thanks for the explanation

Modules/TRD/src/TrackletCountCheck.cxx
martenole: well here one does not need to check, you are creating an object with `new` every time the `check` function is called and never deleting it. You could simply have a member variable plain pointer which you initialize once with a new object and that's it. Or you use a smart pointer.

Modules/TRD/include/TRD/TrackletCountCheck.h
```diff
@@ -10,39 +10,43 @@
...
   std::string getAcceptedType() override;
-  float mThresholdMeanLow, mThresholdMeanHigh;
-  int mStatThreshold;
+  float mThresholdMeanLowPerTrigger, mThresholdMeanHighPerTrigger, mThresholdMeanLowPerTimeFrame, mThresholdMeanHighPerTimeFrame;
+  int mStatThresholdPerTrigger;
+  Quality mResultPertrigger;
+  Quality mResultPerTimeFrame;
+  Quality mFinalResult;
+
+  std::shared_ptr<TPaveText> msg1;
```
martenole, line 44: now please follow the coding guidelines and name it e.g. `mTrackletPerTriggerMessage` or so

Modules/TRD/src/TrackletCountCheck.cxx
```diff
@@ -0,0 +1,189 @@
...
+    ILOG(Debug, Support) << "moName =" << moName << ENDM;
+    (void)moName;
+    if (mo->getName() == "trackletsperevent") {
+      auto* h = dynamic_cast<TH1F*>(mo->getObject());
+      if (h == nullptr) {
+        // ILOG(Debug, Support) << "Requested Histogram type does not match with the Histogram in source" << ENDM;
+        continue;
+      }
+      // warning about statistics available
+      msg1 = std::make_shared<TPaveText>(0.3, 0.7, 0.7, 0.9, "NDC");
```
martenole, line 70:
```suggestion
if (!msg1) {
      msg1 = std::make_shared<TPaveText>(0.3, 0.7, 0.7, 0.9, "NDC");
      msg1->SetTextSize(10);
}
```

deependra170598: In this way, it is inserting each time a new line into the Message pad. To use your suggestion I have to use ```msg1.reset()``` before allocating memory so that the text box does not overflow. Am I correct now?

martenole:
Hi @deependra170598 I don't understand why you need to recreate a new TPaveText everytime. Couldn't you just call [clear()](https://root.cern.ch/doc/master/classTPaveText.html#a9bcacce2648659ce907b43c5325da389) on the text box and then add new text in case needed? Anyways, if you have tested it with your current implementation and it worked this is also fine with me. The CI failures don't seem to be related to your changes.

martenole:
After discussing with @deependra170598 it would be fine for me to merge it as is, if @Barthelemy and @knopers8 agree. Further improvements can be added later. In local tests done by Dependra the checks work as expected.

Barthelemy (approved):

Barthelemy:
I agree, merged !

PR #2016, opened 2023-10-15
ITS - Fixes and improvement on FEE Task checker

nicolovalle:
@iravasen @IsakovAD , this is on the checks for the TriggerVsFeeID plot

- Fix: Before the checker took into account only the number of FEEs excluded from the checks, not their IDs
- Feature: New check allowing to spot if a FEEId stays behind with the number of shipped TFs (e.g. module going off along the run)

### DIFF
--- a/Modules/ITS/src/ITSFeeCheck.cxx
+++ b/Modules/ITS/src/ITSFeeCheck.cxx
@@ -135,37 +135,74 @@ Quality ITSFeeCheck::check(std::map<std::string, std::shared_ptr<MonitorObject>>
     if (mo->getName() == "TriggerVsFeeid") {
       result.set(Quality::Good);
       auto* h = dynamic_cast<TH2I*>(mo->getObject());
-      int counttrgflags[NTrg] = { 0 };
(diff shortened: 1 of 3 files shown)


Modules/ITS/src/ITSFeeCheck.cxx
```diff
@@ -135,37 +135,74 @@ Quality ITSFeeCheck::check(std::map<std::string, std::shared_ptr<MonitorObject>>
...
+        for (int ifee = 1; ifee <= h->GetNbinsX(); ifee++) {
+
+          if (std::find(skipfeeid.begin(), skipfeeid.end(), ifee) != skipfeeid.end())
+            continue;
+
+          int bincontent = (int)(h->GetBinContent(ifee, itrg + 1));
+
+          // checking trigger flags supposed to have at least one entry
+          if (TrgAtLeastOne.Contains(trgname)) {
+            if (bincontent < 1) {
```
IsakovAD, line 171: Is it expected that BAD flag will be triggered also for `bincontent = 1` ?

IsakovAD (approved):

PR #2018, opened 2023-10-18
[WIP] Using C++20 in QC

knopers8:
Now mostly as a demonstrator, since we need O2 & ROOT to move first.

I can confirm though that the QC library compiles fine and passes the tests in o2-qc-test-core

### DIFF
--- a/Framework/include/QualityControl/ActivityHelpers.h
+++ b/Framework/include/QualityControl/ActivityHelpers.h
@@ -36,13 +36,11 @@ concept RangeOfActivities = requires(Range range) {
 namespace implementation
 {
 
-template <RangeOfActivities RangeType>
-Activity commonActivityFields(const RangeType& activities);
+Activity commonActivityFields(const RangeOfActivities auto& activities);
 }
 
 /// Produces the most constrained Activity which will match all the provided
-template <RangeOfActivities RangeType>
-Activity strictestMatchingActivity(const RangeType& activities)
+Activity strictestMatchingActivity(const RangeOfActivities auto& activities)
 {
   if (std::ranges::empty(activities)) {
     return {};
@@ -63,8 +61,7 @@ Activity strictestMatchingActivity(const RangeType& activities)
 
 /// Produces an Activity which matches all the provided, but the validity is a union.
 /// Be sure to check if the result validity is valid, it might not if the argument validities do not overlap
-template <RangeOfActivities RangeType>
-Activity overlappingActivity(const RangeType& activities)
+Activity overlappingActivity(const RangeOfActivities auto& activities)
 {
   if (std::ranges::empty(activities)) {
     return {};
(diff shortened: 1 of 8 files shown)


knopers8:
Rebasing an old PR with some attempts at C++20...

Framework/include/QualityControl/ActivityHelpers.h
```diff
@@ -22,64 +22,59 @@
...
-
+namespace implementation
+{
+Activity commonActivityFields(const RangeOfActivities auto& activities);
 }
 
 /// Produces the most constrained Activity which will match all the provided
-template <typename Iter, typename Accessor = const Activity& (*)(const Activity&)>
-Activity strictestMatchingActivity(
-  Iter begin, Iter end, Accessor accessor = [](const Activity& a) -> const Activity& { return a; })
```
justonedev1, line 41: why was accessor removed?

Framework/include/QualityControl/DataSourceSpec.h
```diff
@@ -39,9 +39,9 @@ enum class DataSourceType {
 // this should allow us to represent all data sources which come from DPL (and maybe CCDB).
 struct DataSourceSpec {
   explicit DataSourceSpec(DataSourceType type = DataSourceType::Invalid);
-  // todo: use c++20 concepts when available
-  template <class... Args, class Enable = std::enable_if_t<(... && std::is_convertible_v<Args, DataSourceType>)>>
-  bool isOneOf(Args... dataSourceType) const
+
+  template <class... DataSourceType>
```
justonedev1, line 43: shouldn't there be any concept similarly to `computeNumericalActivityField` ?

justonedev1 (changes requested):

Framework/include/QualityControl/ActivityHelpers.h
knopers8: The accessor was added in the first place because we did not have a way to easily pass some kind of iterator of all the activities if they are members bigger structures. `std::ranges` allow for that, as it is demostrated in the test, so we can remove this and expect a `RangeOfActivities` instead.

Framework/include/QualityControl/DataSourceSpec.h
knopers8: Why? this method is a member of `DataSourceSpec` and expects `DataSourceType` only, but it can accept one or more of those.

Do you mean just saying something like `requires T == DataSourceType`)?

knopers8: the "todo" comment was probably my lack of awareness, that it could be done simpler.

justonedev1: well, there was a structure that checks whether the template args can be converted to the enum, so I thought that you want to keep the functionality

knopers8: no, not really. i think i had thought that one needs `is_convertible` to make it work also with `const DataSourceType` or `const DataSourceType&`, which apparently is not needed in the end.

justonedev1 (approved):

Barthelemy (approved):

PR #2020, opened 2023-10-19
[QC-1037] IncreasingEntries fault tolerance

Barthelemy:


Barthelemy:
tested

Modules/Common/src/IncreasingEntries.cxx
```diff
@@ -68,19 +66,28 @@ Quality IncreasingEntries::check(std::map<std::string, std::shared_ptr<MonitorOb
...
+    const double currentNumberEntries = histo->GetEntries();
+    size_t faultCount = mMoFaultCount.count(moName) > 0 ? mMoFaultCount.at(moName) : 0;
 
-    if (mMustIncrease && previousNumberEntries == currentNumberEntries) {
-      result = Quality::Bad;
-      result.addReason(FlagReasonFactory::NoDetectorData(), "Number of entries stopped increasing.");
-      mFaultyObjectsNames.push_back(mo->getName());
-    } else if (!mMustIncrease && previousNumberEntries != currentNumberEntries) {
+    if (mMustIncrease && previousNumberEntries == currentNumberEntries ||
+        !mMustIncrease && previousNumberEntries != currentNumberEntries) {
```
knopers8, line 74:
```suggestion
    if (mMustIncrease == (previousNumberEntries == currentNumberEntries)) {
) {
```
shorter, but perhaps less clear... just an idea

doc/Advanced.md
```diff
@@ -1486,6 +1486,18 @@ The behaviour of the check can be inverted by setting the customparameter "mustI
         }
 ```
 
+The number of cycles during which we tolerate increasing (or not respectively) the number of entries can be set with the custom parameter `nbBadCyclesLimit`:
+```
+        "extendedCheckParameters": {
+          "default": {
+            "default": {
+              "nbBadCyclesLimit": "3",
```
knopers8, line 1494:
```suggestion
              "nBadCyclesLimit": "3",
```
I would propose to use just `n` for "number", which is seems more common in our codebase.

knopers8:
needs formatting fix as well

doc/Advanced.md
Barthelemy: agreed, I will commit the change myself because there is a second place.

PR #2023, opened 2023-10-26
ITS Tracks: more plots for legnth of Tracks

IsakovAD:
Added several plots for more detailed monitoring of Track lenght:

1) TrackLegth Vs Eta
2) TrackLegth Vs Phi
3) TrackLegth Vs Pt

4) PtVsEta
5) PtVsPhi

All object are not normalized by nVertex

### DIFF
--- a/Modules/ITS/include/ITS/ITSTrackTask.h
+++ b/Modules/ITS/include/ITS/ITSTrackTask.h
@@ -87,6 +87,7 @@ class ITSTrackTask : public TaskInterface
   TH1D* hNtracksReset;
   std::unique_ptr<TH2DRatio> hNClustersPerTrackEta;
   std::unique_ptr<TH2DRatio> hNClustersPerTrackPhi;
+  std::unique_ptr<TH2DRatio> hNClustersPerTrackPt;
   std::unique_ptr<TH2DRatio> hHitFirstLayerPhiAll;
   std::unique_ptr<TH2DRatio> hHitFirstLayerPhi4cls;
   std::unique_ptr<TH2DRatio> hHitFirstLayerPhi5cls;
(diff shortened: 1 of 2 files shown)


IsakovAD:
@nicolovalle @iravasen Could you have a look at this PR? I added more plots 
after Mateo's request

IsakovAD:
You can see the new plots starting from 4th row of this layout:

https://qcg-test.cern.ch/?page=layoutShow&layoutId=61714cdbc27c751356c9f44c

iravasen:
Hi @IsakovAD, very nice work! I have only one comment: 
- the eta/phi distributions for every set of tracks (4,5,6,7 clusters) are already available with these 2 histograms https://github.com/AliceO2Group/QualityControl/blob/c5184971c72730942335524bc70404fa8c1d49d0/Modules/ITS/src/ITSTrackTask.cxx#L225-L226. I would say that I tend to prefer your new TH1 in this case because you see immediately what's needed but I let you decide what's best. Note also that from GUI you can make projections of bins from a TH2 so a TH2 is not that bad in the end.

iravasen:
I also propose to ask Rik to run this PR on the run we are using to study the hole in phi = 1.4 (sync vs async) just to see whether these new histograms are enough to understand the problem or whether a new set is needed.

iravasen:
btw, if we stay with TH2, you can do the same with the 4 pt distributions for each set of tracks.

IsakovAD:
@iravasen 

I've applied your comments; Now we have 2D plots: TrackLength VS. Eta, TrackLength VS. Phi, TrackLength VS. Pt. 

Everything is normalized by nVertex/nTracks 

Could you see on this layout? 

https://qcg-test.cern.ch/?page=layoutShow&layoutId=61714cdbc27c751356c9f44c

iravasen:
Hi @IsakovAD thanks a lot! One more thing I noted:
- can you please make smaller pt bins? I would say a bin width of 0.2-0.3 GeV/c should be ok. If you have time you can do a width of 0.2 GeV/c up to 1 GeV/c and then 0.5 GeV up to 100 GeV/c (you could also stop at 20-30 GeV/c in case).

IsakovAD:
@iravasen 

Could you approve the PR? You can find the fixed plots by the same link

iravasen (approved):

PR #2024, opened 2023-11-01
[QC-1038] fix Race condition in Service Discovery

Barthelemy:
I used a mutex and a condition variable with a flag to know that the port has been assigned. Perhaps there is a better way ?

### DIFF
--- a/Framework/src/ServiceDiscovery.cxx
+++ b/Framework/src/ServiceDiscovery.cxx
@@ -100,7 +100,7 @@ bool ServiceDiscovery::_register(const std::string& objects)
   // Wait until port is set by the thread
   {
     std::unique_lock<std::mutex> lk(mHealthPortMutex);
-    mHealthPortCV.wait(lk, [this]{return mHealthPortAssigned == true;});
+    mHealthPortCV.wait(lk, [this] { return mHealthPortAssigned == true; });
   }
   check.put("TCP", mHealthUrl + ":" + std::to_string(mHealthPort));
   checks.push_back(std::make_pair("", check));
--- a/Framework/src/ServiceDiscovery.cxx
+++ b/Framework/src/ServiceDiscovery.cxx
@@ -38,6 +38,7 @@ ServiceDiscovery::ServiceDiscovery(const std::string& url, const std::string& na
   : curlHandle(initCurl(), &ServiceDiscovery::deleteCurl), mConsulUrl(url), mName(name), mId(id), mHealthUrl(healthEndUrl)
 {
   mHealthUrl = mHealthUrl.empty() ? boost::asio::ip::host_name() : mHealthUrl;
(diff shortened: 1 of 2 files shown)


Barthelemy:
@awegrzyn could you review too ?

awegrzyn:
Seems ok, but what about simply running `_register` in new thread right after selecting the port? In such case no mutexes are needed and registering to Consul is done in separate thread (no impact on QC main thread).

Barthelemy:
@awegrzyn I thought about it but we call `_register` from many places in the code to update the list of objects we publish. Thus we might still end up calling it before we have a port.

awegrzyn:
Ah okay, understood, but I have a impression that we're getting back to original problem => QC main thread stuck because of waiting for port selection. @knopers8 Could you confirm?

knopers8:
The original problem was different. The port selection routine was selecting a port by binding a port, then releasing it and letting the new thread bind it again. In the time gap between releasing the port and binding it again, sometimes another process was binding it.

I will review the PR tomorrow

awegrzyn:
Ah ok, so if you don't mind to wait in main thread for port selection that it's fine.

Barthelemy:
If I understand correctly what I did, we wait in the main thread for the port selection (that should not take really long) and if none is found it will release the main thread any ways. I did some tests and it seemed to work. Thank you Piotr for reviewing it. Ideally I would like to have it in today's release but if there is a logical flaw we will wait.

Framework/src/ServiceDiscovery.cxx
```diff
@@ -96,6 +97,11 @@ bool ServiceDiscovery::_register(const std::string& objects)
   check.put("Name", "Health check " + mId);
   check.put("Interval", "5s");
   check.put("DeregisterCriticalServiceAfter", "1m");
+  // Wait until port is set by the thread
+  {
+    std::unique_lock<std::mutex> lk(mHealthPortMutex);
+    mHealthPortCV.wait(lk, [this] { return mHealthPortAssigned == true; });
```
knopers8, line 103: Maybe I misunderstood how `condition_variable` works, but won't it block the second time you call `_register`? I am not sure if it requires the other thread to call `notify_one()` for each `wait`.

knopers8: OK, I wrote a little reproducer and it seems it my concern is not valid, it seems that `notify_one()` unblocks the main thread forever, as long as the predicate is provided in `wait`.

knopers8 (approved):
it seems fine, thanks!

Framework/src/ServiceDiscovery.cxx
Barthelemy: as you noted, as long as the predicate is true, it will let the main thread continue.

PR #2028, opened 2023-11-08
[QC-858] Single CheckRunner per workflow

Barthelemy:


Framework/test/testInfrastructureGenerator.cxx
```diff
@@ -367,15 +367,15 @@ TEST_CASE("qc_infrastructure_local_batch_test")
     CHECK(workflow[4].outputs.size() == 0);
   }
 }
-
+using namespace std;
```
knopers8, line 370:
```suggestion
```

knopers8 (approved):
Thanks, it seems correct.

I am slightly worried if it is going to work well with async workflows, where we have one big workflow with all of the tasks/checks. If you have access to alien, perhaps you could run it with this input file and `--remote-batch` option:
```
/alice/data/2023/LHC23zp/539421/apass1/0940/QC/001/QC.root`
```
this config file (but ccdb-test)
```
/alice/data/2023/LHC23zp/539421/apass1/0940/QC/001/QC_production.json
```

Barthelemy:
Good point about async. Not sure how to run what you propose but it is an opportunity to learn. I'll let you know the result.

Barthelemy:
I tested with run 544124. I found 1 bug that I fixed (disabled checks were included). It worked.

PR #2031, opened 2023-11-09
Adapt to fmt 10.1.1

ktf:
Adapt to fmt 10.1.1


ktf:
Can I have a new tag with this?

Barthelemy:
yes, but do you know why `build/QualityControl/o2-dataflow` is not completing ?

Barthelemy:
timo removed it from the list ( it was the slc7)

Barthelemy (approved):

PR #2033, opened 2023-11-14
[QC-1020] Fall back to default values if beamtype or runtype cannot b…

Barthelemy:
…e found.

### DIFF
--- a/doc/Advanced.md
+++ b/doc/Advanced.md
@@ -1086,8 +1086,12 @@ The values can be accessed in various ways.
 
 ### Access optional values with or without activity
 
+The value for the key, runType and beamType is returned if found, or an empty value otherwise. 
+However, before returning an empty value we try to substitute the runType and the beamType with "default". 
+
 ```c++
-// returns an Optional<string> if it finds the key `myOwnKey` for the runType and beamType of the provided activity. 
+// returns an Optional<string> if it finds the key `myOwnKey` for the runType and beamType of the provided activity, 
+// or if it can find the key with the runType or beamType substituted with "default". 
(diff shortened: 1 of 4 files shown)


Barthelemy:
@aferrero2707 This is what you asked for the CustomParameters: if a value is not defined for the requested beam/run type, we use the one defined for "default" (if any). 

To me it seems a bit confusing but I see the benefit as well.

aferrero2707:
@Barthelemy is there a specific reason why there is no implementation of this version of `at()`?
```
std::string CustomParameters::at(const std::string& key, const Activity& activity)
```

doc/Advanced.md
```diff
@@ -1086,8 +1086,12 @@ The values can be accessed in various ways.
 
 ### Access optional values with or without activity
 
+The value for the key, runType and beamType is returned if found, or an empty value otherwise. 
+However, before returning an empty value we try to substitute the runType and the beamType with "default". 
+
 ```c++
-// returns an Optional<string> if it finds the key `myOwnKey` for the runType and beamType of the provided activity. 
+// returns an Optional<string> if it finds the key `myOwnKey` for the runType and beamType of the provided activity,
```
aferrero2707, line 1093: In the current example I think there is some mixed use of `"myOwnKey"` and `"myOwnKey1"`, my impression is that is there should be `"myOwnKey1"` everywhere.

Barthelemy:
1. I added the missing version of `at`
2. loop: yes
3. doc: yes

Thanks !

Barthelemy:
I pushed my changes to my branch but for some reason they don't appear here.

PR #2036, opened 2023-11-16
Enable TrkPID QC

njacazio:
depends on https://github.com/AliceO2Group/AliceO2/pull/12258

### DIFF
--- a/Modules/GLO/src/ITSTPCMatchingTask.cxx
+++ b/Modules/GLO/src/ITSTPCMatchingTask.cxx
@@ -21,6 +21,7 @@
 #include "GLO/ITSTPCMatchingTask.h"
 #include <Framework/InputRecord.h>
 #include <Framework/InputRecordWalker.h>
+#include <QualityControl/stringUtils.h>
 
 using matchType = o2::globaltracking::MatchITSTPCQC::matchType;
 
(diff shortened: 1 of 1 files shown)


Modules/GLO/src/ITSTPCMatchingTask.cxx
```diff
@@ -99,72 +106,7 @@ void ITSTPCMatchingTask::initialize(o2::framework::InitContext& /*ctx*/)
 
   mMatchITSTPCQC.initDataRequest();
   mMatchITSTPCQC.init();
-
```
njacazio, line 102: @chiarazampolli I changed the way we publish the histograms so that we don't need to duplicate the PRs. I can avoid it if you think it's unecessary

Modules/GLO/src/ITSTPCMatchingTask.cxx
```diff
@@ -42,6 +42,13 @@ void ITSTPCMatchingTask::initialize(o2::framework::InitContext& /*ctx*/)
     }
   }
 
+  if (auto param = mCustomParameters.find("useTrkPID"); param != mCustomParameters.end()) {
+    ILOG(Debug, Devel) << "Custom parameter - useTrkPID (= add plots for tracking PID): " << param->second << ENDM;
+    if (param->second == "true" || param->second == "True" || param->second == "TRUE") {
```
Barthelemy, line 47: `bool stringUtils::decodeBool(const std::string& value)` does the same as this condition.

Modules/GLO/src/ITSTPCMatchingTask.cxx
```diff
@@ -99,72 +106,7 @@ void ITSTPCMatchingTask::initialize(o2::framework::InitContext& /*ctx*/)
...
-  getObjectsManager()->startPublishing(mMatchITSTPCQC.getHistoResidualPt());
-  getObjectsManager()->startPublishing(mMatchITSTPCQC.getHistoResidualPhi());
-  getObjectsManager()->startPublishing(mMatchITSTPCQC.getHistoResidualEta());
-
-  getObjectsManager()->startPublishing(mMatchITSTPCQC.getHistoChi2Matching());
-  getObjectsManager()->startPublishing(mMatchITSTPCQC.getHistoChi2Refit());
-  getObjectsManager()->startPublishing(mMatchITSTPCQC.getHistoTimeResVsPt());
-
-  getObjectsManager()->startPublishing(mMatchITSTPCQC.getHistoDCAr());
+  mMatchITSTPCQC.publishHistograms(getObjectsManager());
```
Barthelemy, line 109: Is `publishHistograms` used somewhere else ? or just in QC ?

njacazio: I think just in QC, it is not the only way to access the histogram and publish them, however like this one does not need to have 2 PRs in order to publish all the histograms and one in the O2 is enough.

Modules/GLO/src/ITSTPCMatchingTask.cxx
njacazio: thanks, I implemented it! Also, while looking for `decodeBool` in the QualityControl I found several implementations, maybe this can be unified

Modules/GLO/src/ITSTPCMatchingTask.cxx
Barthelemy: ok understood

PR #2043, opened 2023-11-22
Introduce an env var to send several times the registration to BK

Barthelemy:
This is needed to test with Martin


knopers8 (approved):

Barthelemy:
errors unrelated, merging

PR #2037, opened 2023-11-17
TRD cleanup work

martenole:
Requires changes in O2DPG (update of json file) and O2 (some additional helper functions, PR will be opened soon).

Number of objects produced by digits task by default reduced to 50.

Quite some bugs fixed + major refactoring of the monitoring function.

PR #2049, opened 2023-11-28
TRD digit QC protection against division by zero

martenole:
Fix for FPE observed in alibi nightly MC tests.
Ping @knopers8

### DIFF
--- a/Modules/TRD/src/DigitsTask.cxx
+++ b/Modules/TRD/src/DigitsTask.cxx
@@ -306,7 +306,9 @@ void DigitsTask::monitorData(o2::framework::ProcessingContext& ctx)
       continue;
     }
     mDigitsPerEvent->Fill(nDigits);
-    mDigitsSizevsTrackletSize->Fill(nDigits / trigger.getNumberOfTracklets());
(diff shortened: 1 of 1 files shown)


knopers8 (approved):

chiarazampolli:
Hello @martenole , @knopers8 , 

We will need a new tag from QC for PbPb with this fix. Can this be done? We would need it as soon as possible - as always :-) 

Cheers,
Chiara

chiarazampolli:
Actually, at the moment we are using v1.123.0, so the bug is not there. But in general, we should understand how to handle data and MC when we use dev/master of O2/O2DPG, but an "old" tag from QC...

martenole:
Hi @chiarazampolli the bug has been there for a very long time. Also in v1.123.0 it is present. It is just very unlikely to have zero tracklets for a whole TF that's why it appeared only once now. I doubt it will ever happen in PbPb, but still I think it would be better to have the protection in..

martenole:
> Hello @martenole , Was it not introduced here?

Yes, you are right, sorry. Indeed I introduced the bug in https://github.com/AliceO2Group/QualityControl/pull/2037 and checked wrongly for an older version

PR #2057, opened 2023-11-30
[EMCAL-530] Use GRPGeomHelper for access of the geometry

mfasDa:


mfasDa:
Blocks https://github.com/AliceO2Group/O2DPG/pull/1356

Modules/EMCAL/src/ClusterTask.cxx
```diff
@@ -328,6 +328,10 @@ void ClusterTask::startOfCycle()
 
 void ClusterTask::monitorData(o2::framework::ProcessingContext& ctx)
 {
+  if (!o2::base::GeometryManager::isGeometryLoaded()) {
```
Barthelemy, line 331: by having a non-debug log in `monitorData` you will probably flood the infologger as soon as the geometry has not been loaded correctly...

mfasDa: Done. Using static variable to display only once.

Barthelemy (approved):

PR #2060, opened 2023-12-01
[EMCAL-1037] Support for multiple classes per trigger type

mfasDa:
- Support for multiple classes per trigger type
- Determine trigger cluster fromt the cluster in
  the CTP configuration instead of the class name
  (relevant for unanchored MC and 2022 data)
- Require beam presence only when encoded in string


mfasDa:
Fixes 2 problems:
a) In PbPb there were several MB trigger classes, we want an and of them
b) Trigger class names had a different layout then the standard layout, for which parsing (adpated to 2023) failed. The new version is more general

mfasDa:
CI errors unrelated.

Barthelemy (approved):

PR #2062, opened 2023-12-04
[QC-1041] Reset checkrunner

Barthelemy:


knopers8 (approved):
Thanks, all good, I would rename the PR/commit to "Reset Checks in CheckRunner" since the latter is already reset. Please also add some explanation what to reset in `CheckInterface::reset` in form of a comment, so the users know what to do there when implementing a new Check.

PR #2065, opened 2023-12-05
Fix ERROR at the end of short runs

lauraser:
Fix ERROR at the end of short runs, when no trigger is present for Quality Observer


Barthelemy (approved):

wiechula (approved):

wiechula:
@Barthelemy , not sure why the test fail. This does not seem to relate to the changes here.

PR #2070, opened 2023-12-08
TRD further code cleanup

martenole:
Hi @deependra170598 and @maroozm and @nasirmehdimalik 
please take a look if I have removed anything in this PR that might eventually be used. I think not, but better if you double-check of course. The fit of the PH plot is now done in O2 in the T0 calibration. Many classes were obsolete as far as I can tell.
The TRDQC.json I would like to keep for us to do local tests. It works with some local reconstruction from CTF files. But from the postprocessing I don't see if it is done at all or not. But also not sure if it is supposed when run locally. @deependra170598 do you know?


maroozm:
Hi @martenole ,
   Everything seems fine from my side.
Thanks,
Salman

Barthelemy:
@martenole do you want more approvals from your side before I merge ?

deependra170598:
Yes @martenole ,there is no setting for postprocessing in TRDQC.json till now. So if we want local postprocessing we have to update it. 
Regarding generating trend for t0 peak position and its chi2 value trend of fit, this job was done by TRDReductor.cxx. As you said fit is done in O2, so I have to look at how I can find trends from there.

deependra170598:
Hi @martenole , TRDQC.json contains examples of postprocessing settings for generating trends for tracklet mean etc and summerising Quality of checks. These settings were used by me for local testing and also in staging. After last PR by me on trackletCountCheck, I need to verify these settings again.

TRDReductor.cxx was responsible for generating t0peak position trend and chi2 trend of fit. As you said it is now done in O2 t0 calibration, I have to look at how can I find trend from there. Thanks.

martenole:
This is ready to be merged :+1:

PR #2074, opened 2023-12-14
Generalise CompletionPolicy callback to allow for refactoring

ktf:
Generalise CompletionPolicy callback to allow for refactoring

I will soon move to pass a DataProcessorSpec, rather than a DeviceSpec,
because the completion policy is a property of the former.

This will allow detecting consumeWhenAny policies early enough and
make dataprocessors associated to it be resilient to expendable devices
and Sporadic inputs.


ktf:
@Barthelemy @knopers8 I will need a new tag with this.

Barthelemy (approved):

PR #16, opened 2018-05-04
CCDB Web GUI backend

Barthelemy:

PR #2077, opened 2023-12-19
[QC-1079] Run type as a string

Barthelemy:


Barthelemy:
DON't MERGE

knopers8 (commented):
i did not have time for a full review, i have to go now, will try later or after christmas break

Barthelemy:
@knopers8 I don't think that the test error is legit. It is ready to be reviewed.

Framework/src/runnerUtils.cxx
```diff
@@ -61,19 +62,35 @@ bool hasChecks(const std::string& configSource)
...
-  auto run_stop_time_ms = computeActivityField<unsigned long>(services, "runEndTimeMs", fallbackActivity.mValidity.getMax());
+  auto runNumber = computeNumericalActivityField<int>(services, "runNumber", fallbackActivity.mId);
+  std::string runType = services.get<framework::RawDeviceService>().device()->fConfig->GetProperty<std::string>("runType", fallbackActivity.mType);
+  // runType used to be an integer. If we find an integer in a config file, the risk is that it is translated directly to a string (2->"2").
+  // We must rather translate the integer into the corresponding run type string if at all possible.
+  if (is_number(runType)) {
+    try {
+      runType = parameters::GRPECS::RunTypeNames.at(std::stoi(runType));
+    } catch (std::out_of_range& exc) {
+      ILOG(Warning, Ops) << "Activity type was provided as an integer. No matching activity type could be find. Using 'NONE'." << ENDM;
```
knopers8, line 81: should we warn also if the text type could be found? I assume we want users to move all the config files to use text run type instead of integer anyway?

knopers8 (changes requested):

Framework/src/ActivityHelpers.cxx
```diff
@@ -73,7 +71,7 @@ core::Activity asActivity(const std::map<std::string, std::string>& metadata, co
 core::Activity asActivity(const boost::property_tree::ptree& tree, const std::string& provenance)
 {
   core::Activity activity;
-  if (auto runType = tree.get_optional<int>(metadata_keys::runType); runType.has_value()) {
```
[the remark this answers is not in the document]
Barthelemy: I completely agree for the method that takes a tree. So something like ?
```
  if (auto runType = tree.get_optional<std::string>(metadata_keys::runType); runType.has_value()) {
    activity.mType = runType.value();
  } else if (auto runTypeInt = tree.get_optional<int>(metadata_keys::runType); runTypeInt.has_value()) {
    activity.mType = std::to_string(runTypeInt.value());
  }
```
And yes I will add a test. 

For the method that takes metadata map, I need a way to convert the int (which is passed as a string) to the equivalent representation. I will use the ECSDataAdapter in O2 for that if you agree.

knopers8:
> @knopers8 I don't think that the test error is legit. It is ready to be reviewed.

It seems it consistently fails on macOS, other PRs are not failing there:
```
## sw/BUILD/QualityControl-latest/log
 8/31 Test  #8: testCcdbDatabase ........................SIGTRAP***Exception:   1.33 sec
 ```

**Comment by iravasen** [comment]
Hi @Barthelemy, thanks! Sorry for the late comments/questions to this PR:
- do we have to modify also jsons used at P2 online? 
- any implication for the async jsons in O2DPG (I guess not but asking to be sure)? 

Thanks!

**Comment by Barthelemy** [comment]
@iravasen We will update the run type in the QCDB as a bulk. It will be backward compatible though. See the details here: https://its.cern.ch/jira/browse/QC-1079 
We are late but we carry on.

File: Framework/src/runnerUtils.cxx (hunk 1 above)
Comment (Barthelemy) [reply Framework/src/runnerUtils.cxx:81]: Yes.

**Comment by Barthelemy** [comment]
@knopers8 I have finished addressing your comments. Can you check again ? please don't merge.

**Comment by knopers8** [comment]
A bunch of tests is failing:
```
11/30 Test #15: testCheckWorkflow .......................***Failed    2.79 sec
12/30 Test #16: testWorkflow ............................***Failed    3.13 sec
18/30 Test #23: functional_test .........................***Failed    3.77 sec
19/30 Test #24: multinode_test ..........................***Failed    4.46 sec
20/30 Test #25: batch_test ..............................***Failed    3.18 sec
```

knopers8 (approved):

PR #2078, opened 2023-12-20
ITS - FEETask: adding decoding of CDW word

nicolovalle:
Dear @IsakovAD , @iravasen , please have a look at this.

The new objects are filled and published only if the condition is set to True in the config file. It is intended to run with this option only when attached to calibration run. By setting DecodeCDW = 0 there will be no interference with physics runs.

Tested locally.

### DIFF
--- a/Modules/ITS/include/ITS/ITSFeeTask.h
+++ b/Modules/ITS/include/ITS/ITSFeeTask.h
@@ -60,6 +60,30 @@ class ITSFeeTask final : public TaskInterface
     } indexWord;
   };
 
+  // TODO: use proper casting and extend to full word
+  struct CalibrationWordUserField { // CDW GBT word user field
+    union {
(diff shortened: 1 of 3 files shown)


iravasen:
Hi @nicolovalle, looks perfect to me! For the parts related to TDT, are they related to some other discussion (not related to calibration)?

nicolovalle:
> Hi @nicolovalle, looks perfect to me! For the parts related to TDT, are they related to some other discussion (not related to calibration)? I let @IsakovAD to have a look but on calibration side looks ok to me.

Hi @iravasen . Yes, the TDT part is the same as it was before. I just grouped in a single “if” statement since both TDT and CDW counts require payload parsing. 
While in the second commit I corrected a bug for compatibility with zero-padding format (both TDTs and CDWs)

IsakovAD (approved):

PR #2079, opened 2023-12-21
Fix type cast introducing an unwanted cut in ITSchi2.

mfaggin:


mfaggin:
@chiarazampolli @noferini @shahor02 I should have found the bug creating the strange eta distribution for ITS tracks below:

image

The issue was that the parameters from the `.json` file are read as `string` and then casted. By mistake, the configuration for the ITSchi2/NDF was casted to `int` instead of `float`, and given that the value passed through the configuration is `1e10f` then it was casted into `atoi("1e10f") == 1`, introducing an unwanted cut at `1`.
Below the plot after the fix:

image

iravasen:
Hi @mfaggin nice catch!! Thanks a lot. 
I approved it but maybe @Barthelemy or @knopers8 can already force-merge given that the modification is trivial.

chiarazampolli:
Hello, 

If we want this in apass1, we need to patch the QC tag too. @Barthelemy , @knopers8 , not sure you're online to do that...

Chiara

shahor02:
I would not do another build just for that. Cannot we simply use a local `itstpc.json` passed via QC_JSON_GLO_ITSTPC with 
```
 "maxChi2PerClusterITS": "10000"
```
?

chiarazampolli:
This is good idea! @mfaggin , can you open a PR to O2DPG? Then i will patch the tag.

Single json files cannot be taken, as things are at the moment, we could take a full final json. Still, i want to have ad-hoc settings. I prefer the overhead of a new tag.

mfaggin:
Hi @chiarazampolli @shahor02 if I get it properly, this PR in O2DPG should do the job as you ask: https://github.com/AliceO2Group/O2DPG/pull/1401

Barthelemy (approved):

PR #2081, opened 2024-01-05
BigScreen - Fixed retrieval of latest available QO

aferrero2707:
This PR fixes the retrieval of QOs when using periodic triggers. It introduces the ability to retrieve the latest available QO from the QCDB, instead of using the trigger's time stamp. The QO validity range is then compared to the time stamp of the trigger to check if the QO is recent enough to be displayed.

### DIFF
--- a/Modules/Common/src/BigScreen.cxx
+++ b/Modules/Common/src/BigScreen.cxx
@@ -20,6 +20,7 @@
 #include "QualityControl/MonitorObject.h"
 #include "QualityControl/DatabaseInterface.h"
 #include "QualityControl/ObjectMetadataKeys.h"
+#include "QualityControl/ActivityHelpers.h"
 #include <TDatime.h>
 #include <TPaveText.h>
 #include <chrono>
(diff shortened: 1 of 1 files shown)


Barthelemy (approved):

Modules/Common/src/BigScreen.cxx
```diff
@@ -105,19 +106,27 @@ void BigScreen::initialize(quality_control::postprocessing::Trigger t, framework
...
-  if (!qo) {
+  // find the time-stamp of the most recent object matching the current activity
+  // if ignoreActivity is true the activity matching criteria are not applied
+  Activity activity = ignoreActivity ? Activity{} : t.activity;
+  auto timestamp = t.timestamp;
+  const auto objFullPath = t.activity.mProvenance + "/" + source.path;
+  const auto filterMetadata = activity_helpers::asDatabaseMetadata(activity, false);
+  const auto objectValidity = qcdb.getLatestObjectValidity(objFullPath, filterMetadata);
+  if (objectValidity.isValid()) {
+    timestamp = objectValidity.getMax() - 1;
```
knopers8, line 117: why `objectValidity.getMax() - 1` in particular?

aferrero2707: @knopers8 honestly I simply used the same logic as in the `newObject` trigger for the object's time stamp (see [here](https://github.com/AliceO2Group/QualityControl/blob/64fa7f18968589110b8f2b2b9f34458ba0be86fc/Framework/src/Triggers.cxx#L196)).
If there is a better choice just let me know. Thanks!

knopers8: Thanks. There, `max-1` is used because it is the max timestamp that one can use to access the object. In the case of the trigger, I use it because it makes the most sense as a timestamp of data points on trending plots.

In your case, I am wondering if using `"Created-At"` instead of `"Valid-Until" - 1` could make more sense. If we use the first, you will consider an object as too old if it was not updated recently. If you use the latter, a regularly updated object, but with no new data since some time (max validity does not move) will not be consider as too old. Personally, I am OK with both alternatives, so I think it's up to you.

aferrero2707: For the purpose of the BigScreen, I think it is useful to also identify the cases where an object is regularly updated but no new data is actually received, and show it as grey if the last new data is too far in the past. What do you think?

knopers8: Yes, indeed, this is something that shifters should be aware of.

So the algorithm would be:
1) get the latest object validity
2) if `now - created > x`, consider the object as invalid, use an appropriate colour in the display
3) if `now - validity_end > y` consider the object as invalid, use an appropriate (different) colour in the display.

I would keep `x` and `y` as different parameters in case we want to tune them separately. After all, there is always some latency between the validity end and the object publication time.

Colour could be also different between the two, so it is clear which of the two cases we encounter.

aferrero2707: That's a very good suggestion! We can extend the logic in a separate PR, for example:
- grey color if `now - created > x` (the processing is not creating new QOs)
- yellow color if `now - validity_end > y` (processing creates new QOs, but from the same old data)

PR #2082, opened 2024-01-09
[QC-1083] Introduce a start/end of activity method in checkInterface …

Barthelemy:
…and aggregatorInterface


Barthelemy:
I have kept the same signature as in `TaskInterface` : `startOfActivity(const core::Activity& activity)`

I have also removed the `mActivity` from the `Interface` to be something identical to `TaskInterface`.

knopers8 (approved):

PR #2094, opened 2024-01-18
CTP: BC filling scheme check

lhusova:
fix for the BC filling scheme check


Modules/CTP/include/CTP/RawDataReaderCheck.h
```diff
@@ -18,6 +18,7 @@
 #define QC_MODULE_CTP_CTPRAWDATAREADERCHECK_H
 
 #include "QualityControl/CheckInterface.h"
+#include "TH1.h"
```
Barthelemy, line 21: You could use a forward declaration instead.

Barthelemy (changes requested):
Thank you for your PR. Just a couple of minor things to fix.

Barthelemy (approved):
Thank you

PR #2096, opened 2024-01-18
Modernize TaskRunner completion policy

knopers8:
This method was written when one could not access InputSpecs from within the callback. Also, CCDB inputs were not supported back then. This commit makes the completion policy rely on InputSpecs to recognize the type of expected inputs and differentiates between Data and Condition inputs.

Preparatory work for QC-1075.


Barthelemy:
All clear, thank you

Barthelemy (approved):

PR #2098, opened 2024-01-19
[QC-1075] Allow for multiple data sources in QC tasks

knopers8:
Which, effectively, allows to declary CCDB inputs as direct alongside of a data sampling policy.

To my impression, it seems to work with the example provided by Markus.

To be decided if allowing for multiple data sources is the best way forward. The users might try to combine direct *data* sources with sampled sources, which will normally leave plenty of partially filled timeslices which are never executed. However, this generalization is in line with multiple data sources for Checks and Aggregators.
A possible approach would be to declare a ccdb data source, which is interally treated as direct, but can be explicitely asserted that it is used in combination with a data sampling policy.

### DIFF
--- a/Framework/include/QualityControl/TaskSpec.h
+++ b/Framework/include/QualityControl/TaskSpec.h
@@ -39,13 +39,13 @@ struct TaskSpec {
 
   // minimal valid spec
   TaskSpec(std::string taskName, std::string className, std::string moduleName, std::string detectorName,
-           int cycleDurationSeconds, DataSourceSpec dataSource)
+           int cycleDurationSeconds, std::vector<DataSourceSpec> dataSources)
     : taskName(std::move(taskName)),
       className(std::move(className)),
       moduleName(std::move(moduleName)),
       detectorName(std::move(detectorName)),
       cycleDurationSeconds(cycleDurationSeconds),
-      dataSource(std::move(dataSource))
+      dataSources(std::move(dataSources))
   {
   }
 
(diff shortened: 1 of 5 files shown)


knopers8:
Hi @mfasDa , this PR, in combination with the attached config file, allows to put CCDB inputs as direct in combination with a sampling policy. On my setup, I was getting valid results, but perhaps I am missing something. Could you please give it a go as well?

The way to configure might change a bit, but the underlying behaviour should stay the same... that is assuming that you confirm this works as expected. 

[cells2.json](https://github.com/AliceO2Group/QualityControl/files/14008872/cells2.json)

knopers8:
@mfasDa Since you might be busy with something else, I will propose to move on with this PR, then later we can address any issues you might encounter while trying it.

@Barthelemy Ready for review

Barthelemy:
@knopers8 do I understand correctly that all the config files will have to be adapted ?

knopers8:
No, it is backwards-compatible.

Barthelemy:
indeed, sorry, I missed one line

Barthelemy (approved):

Barthelemy:
waiting for `o2` to finish. I don't think that the failure of O2fst/o2 is related

mfasDa:
Hi @knopers8 , Hi @Barthelemy ,

Sorry for the late reply! The solution works for us. I get both the CCDB objects and the downsampled data. Our PR can now also be merged.

Thanks a lot for your help!

PR #2101, opened 2024-01-23
CTP: Trending

lhusova:
Trending of input and class rates

### DIFF
--- a/Modules/CTP/CMakeLists.txt
+++ b/Modules/CTP/CMakeLists.txt
@@ -1,8 +1,10 @@
 # ---- Library ----
 
-add_library(O2QcCTP)
+add_library(O2QcCTP
+            src/TrendingConfigCTP.cxx
+            src/TH1ctpReductor.cxx )
 
-target_sources(O2QcCTP PRIVATE src/RawDataReaderCheck.cxx  src/CountersQcTask.cxx  src/RawDataQcTask.cxx )
+target_sources(O2QcCTP PRIVATE src/RawDataReaderCheck.cxx  src/CountersQcTask.cxx  src/RawDataQcTask.cxx  src/TH1ctpReductor.cxx  src/CTPTrendingTask.cxx)
 
 target_include_directories(
   O2QcCTP
@@ -18,11 +20,13 @@ install(TARGETS O2QcCTP
         RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR})
 
 add_root_dictionary(O2QcCTP
-                    HEADERS
-  include/CTP/CountersQcTask.h
-include/CTP/RawDataQcTask.h
-      include/CTP/RawDataReaderCheck.h
-                          LINKDEF include/CTP/LinkDef.h)
+                    HEADERS include/CTP/CountersQcTask.h
+                            include/CTP/RawDataQcTask.h
+                            include/CTP/RawDataReaderCheck.h
+                            include/CTP/TH1ctpReductor.h
+                            include/CTP/CTPTrendingTask.h
+                            include/CTP/TrendingConfigCTP.h
+                    LINKDEF include/CTP/LinkDef.h)
 
 install(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/include/CTP
   DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}/QualityControl")
@@ -42,4 +46,17 @@ foreach(test ${TEST_SRCS})
   set_property(TARGET ${test_name}
     PROPERTY RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/tests)
   set_tests_properties(${test_name} PROPERTIES TIMEOUT 20)
+endforeach()
+
+# ---- Extra scripts ----
+
+install(FILES
+        src/ctpTrending.json
+        DESTINATION etc)
+
+get_property(dirs
+             DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
+             PROPERTY INCLUDE_DIRECTORIES)
+foreach(dir ${dirs})
+  message(STATUS "dir='${dir}'")
 endforeach()
--- a/Modules/CTP/include/CTP/CTPTrendingTask.h
+++ b/Modules/CTP/include/CTP/CTPTrendingTask.h
@@ -1,0 +1,74 @@
+// Copyright 2019-2020 CERN and copyright holders of ALICE O2.
+// See https://alice-o2.web.cern.ch/copyright for details of the copyright holders.
+// All rights not expressly granted are reserved.
+//
+// This software is distributed under the terms of the GNU General Public
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+///
+/// \file     CTPTrendingTask.h
+/// \author   Lucia Anna Tarasovicova
+///
+
+#ifndef QUALITYCONTROL_CTPTTRENDINGTASK_H
+#define QUALITYCONTROL_CTPTTRENDINGTASK_H
+
+#include "QualityControl/PostProcessingInterface.h"
+#include "CTP/TrendingConfigCTP.h"
+#include "QualityControl/Reductor.h"
+#include <DataFormatsCTP/Configuration.h>
+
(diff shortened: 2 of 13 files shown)


Modules/CTP/CMakeLists.txt
```diff
@@ -43,3 +47,16 @@ foreach(test ${TEST_SRCS})
...
   set_tests_properties(${test_name} PROPERTIES TIMEOUT 20)
 endforeach()
+
+# ---- Extra scripts ----
+
+install(FILES
+        src/ctpTrending.json
+        DESTINATION etc)
+
+get_property(dirs
```
Barthelemy, line 57: could you remove this if it is only there to debug ?

Modules/CTP/src/CTPTrendingTask.cxx
```diff
@@ -0,0 +1,223 @@
...
+
+      } else {
+        ILOG(Error, Devel) << "Could not get the htemp histogram of the plot '" << plot.name << "'." << ENDM;
+      }
+
+      mPlots[plot.name] = c;
+      getObjectsManager()->startPublishing(c);
+
+      if (index < numberOfInputs)
+        inputIndex++;
```
Barthelemy, line 217: It is recommended to put braces even when there is a single line

Modules/CTP/src/CTPTrendingTask.cxx
```diff
@@ -0,0 +1,223 @@
...
+      auto* c = new TCanvas();
+      ILOG(Info, Support) << plot.varexp << " " << plot.selection << " " << plot.option << ENDM;
+      mTrend->Draw(plot.varexp.c_str(), plot.selection.c_str(), plot.option.c_str());
+
+      c->SetName(plot.name.c_str());
+      c->SetTitle(plot.title.c_str());
+
+      if (auto histo = dynamic_cast<TH1*>(c->GetPrimitive("htemp"))) {
+        if (index < numberOfInputs)
+          histo->SetTitle(ctpInputs[inputIndex].name.c_str());
```
Barthelemy, line 191: It is recommended to put braces even when there is a single line

Modules/CTP/src/CTPTrendingTask.cxx
```diff
@@ -0,0 +1,223 @@
...
+      continue;
+    else
+      trendedClassIndex.push_back(ctpClass.getIndex());
+  }
+
+  ILOG(Info, Support) << "Generating " << trendedClassIndex.size() + trendedInputIndex.size() << " plots." << ENDM;
+
+  for (const auto& plot : mConfig.plots) {
+
+    if (index < numberOfInputs && index != trendedInputIndex[inputIndex] - 1) {
```
Barthelemy, line 168: A bit of documentation on what is being done here would be nice. In particular given that there are two conditions leading in the end to the same duplicated code.

Modules/CTP/include/CTP/CTPTrendingTask.h
```diff
@@ -0,0 +1,74 @@
...
+  ~CTPTrendingTask() override = default;
+
+  /// \brief Post-processing methods inherited from 'PostProcessingInterface'.
+  void configure(const boost::property_tree::ptree& config) override;
+  void initialize(Trigger, framework::ServiceRegistryRef) override;
+  void update(Trigger, framework::ServiceRegistryRef) override;
+  void finalize(Trigger, framework::ServiceRegistryRef) override;
+
+ private:
+  struct MetaData {
```
Barthelemy, line 54: Why do you need this struct ? you could directly store the activity or just the run number if needed.

Barthelemy (commented):
Hi, 

Thank you !

I made a first quick review. Please see my comments. @knopers8 do you want to have a look too or not ? 

One thing that I would like to discuss is the number of plots. 100 more plots is non-negligible. Could some be grouped ? Could a higher dimension plot be used ? How many more are expected to come in the future ? 

It is crucial for the QCDB to limit the amount of objects that are updated at fast pace such as this one and that are then kept long term.

Modules/CTP/src/CTPTrendingTask.cxx
```diff
@@ -0,0 +1,223 @@
...
+    }
+    ILOG(Info, Support) << "Got MO " << mo << ENDM;
+    TObject* obj = mo ? mo->getObject() : nullptr;
+    if (obj) {
+      foundInputs = true;
+      mReductors[dataSource.name]->update(obj);
+    }
+  }
+
+  if (!foundInputs) {
```
knopers8, line 106: I am wondering if you really should add another row in the tree if you manage to retrieve only some objects from the QCDB. What may happen is that your new row will consists of partially obsolete values, which might be misleading. Up to you to decide if this is what you want.

If not, you could invert the logic and have `bool inputMissing = false;` which is set to true when one of the `retrieveMO` calls fails, then you can skip `mTrend->Fill()`.

Modules/CTP/src/CTPTrendingTask.cxx
```diff
@@ -0,0 +1,223 @@
...
+
+  const int numberOfInputs = 48 - nonExistingInputs.size();
+  ILOG(Info, Support) << "numberOfInputs.  " << numberOfInputs << ENDM;
+
+  std::vector<int> trendedInputIndex;
+  std::vector<int> trendedClassIndex;
+  for (int i = 0; i < ctpInputs.size(); i++) {
+    auto input = ctpInputs[i];
+    if (std::find(nonExistingInputs.begin(), nonExistingInputs.end(), input.getIndex()) != std::end(nonExistingInputs)) {
+      ILOG(Info, Support) << "Input " << input.getIndex() << " should not exist!!" << ENDM;
```
knopers8, line 143: Should it be a warning perhaps?

Modules/CTP/src/CTPTrendingTask.cxx
```diff
@@ -0,0 +1,223 @@
...
+    } else {
+      // Before we generate any new plots, we have to delete existing under the same names.
+      // It seems that ROOT cannot handle an existence of two canvases with a common name in the same process.
+      if (mPlots.count(plot.name)) {
+        getObjectsManager()->stopPublishing(plot.name);
+        delete mPlots[plot.name];
+      }
+
+      auto* c = new TCanvas();
+      ILOG(Info, Support) << plot.varexp << " " << plot.selection << " " << plot.option << ENDM;
```
knopers8, line 183: This looks like a debug or developer log.

Modules/CTP/src/CTPTrendingTask.cxx
```diff
@@ -0,0 +1,223 @@
...
+using namespace o2::ctp;
+
+void CTPTrendingTask::configure(const boost::property_tree::ptree& config)
+{
+  mConfig = TrendingConfigCTP(getID(), config);
+}
+
+void CTPTrendingTask::initialize(Trigger, framework::ServiceRegistryRef services)
+{
+  // Preparing data structure of TTree
```
knopers8, line 49: Please delete any existing pointers in `mPlots` and clear it to account for the correct START -> STOP -> START sequence

knopers8 (changes requested):

Modules/CTP/include/CTP/TH1ctpReductor.h
```diff
@@ -18,7 +18,7 @@
 
 #include "QualityControl/Reductor.h"
 
-namespace o2::quality_control_modules::ctp
+namespace o2::quality_control::postprocessing
```
knopers8, line 21: Since this Reductor is CTP-specific, I would keep it within the `o2::quality_control_modules::ctp` namespace.

Modules/CTP/include/CTP/TH1ctpReductor.h
```diff
@@ -0,0 +1,80 @@
...
+
+#include "QualityControl/Reductor.h"
+
+namespace o2::quality_control::postprocessing
+{
+
+/// \brief A Reductor which obtains the characteristics of TH1 including the bin contents.
+/// Based on the example by Piotr Jan Konopka
+/// It produces a branch in the format: "mean/D:stddev:entries:inputs[%i]:classes[%i]"
+class TH1ctpReductor : public quality_control::postprocessing::Reductor
```
knopers8, line 27:
```suggestion
class TH1ctpReductor : public quality_control::postprocessing::ReductorTObject
```
There was a change in the framework while you were working on your PR. Essentially, you have to inherit now `ReductorTObject`. You may have a look how I adapted the tasks which were already in the codebase: https://github.com/AliceO2Group/QualityControl/pull/2110/files

PR #2106, opened 2024-01-25
PostProcessingInterface inherits UserCodeInterface

knopers8:
which brings the possibility to use retrieveConditionAny also in post processing tasks


Barthelemy (approved):
that went better than I tought !
Thank you

PR #2108, opened 2024-01-26
TOF: Add trending for calibration objects (LHC phase)

ercolessi:
@noferini

### DIFF
--- a/Modules/TOF/src/TrendingCalibLHCphase.cxx
+++ b/Modules/TOF/src/TrendingCalibLHCphase.cxx
@@ -113,8 +113,8 @@ void TrendingCalibLHCphase::trendValues(const Trigger& t, repository::DatabaseIn
       } else {
         ILOG(Info, Support) << "Retrieved calibration file '" << dataSource.path << "'." << ENDM;
         mPhase = calib_object->getLHCphase(0);
-        mStartValidity = calib_object->getStartValidity();
-        mEndValidity = calib_object->getEndValidity();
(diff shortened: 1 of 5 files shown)


Modules/TOF/src/TrendingCalibLHCphase.cxx
```diff
@@ -0,0 +1,206 @@
...
+
+  mPhase = 0.;
+
+  map<string, string> metadata; // can be empty
+
+  for (auto& dataSource : mConfig.dataSources) {
+
+    if (dataSource.type == "ccdb") {
+
+      auto calib_object = mCdbApi.retrieveFromTFileAny<LHCphase>(dataSource.path, metadata, -1);
```
knopers8, line 105: Since very recently, you can just use `UserCodeInterface::retrieveConditionAny` (an inherited member of TrendingCalibLHCPhase) instead of using CcdbAPI directly.

Also, soon you should be able to write only a Reductor implementation to achieve this, as it is demonstrated in `testReductor.cxx` in https://github.com/AliceO2Group/QualityControl/pull/2110.

ercolessi: Hi @knopers8,
thanks for pointing this out! 
I am using now directly the member ``UserCodeInterface::retrieveConditionAny``, also I will keep an eye on the new Reductor implementation

njacazio (approved):
Thanks Francesca

chiarazampolli:
Very nice, I always wanted to have this!!

PR #2111, opened 2024-01-26
Fix C++20 regression

ktf:
Fix C++20 regression


ktf:
@Barthelemy @knopers8 we need a new tag with this...

knopers8 (approved):

PR #2116, opened 2024-01-31
Fix deprecated-enum-float-conversion to keep C++20 happy

ktf:
Fix deprecated-enum-float-conversion to keep C++20 happy


ktf:
@knopers8 @Barthelemy

ktf:
@knopers8 @Barthelemy this is hopefully the last bit before we enable C++20. Can we have a release with it and propagate it to alidist?

Barthelemy (approved):

Barthelemy:
@ktf Here is the release: https://github.com/alisw/alidist/pull/5340

PR #2118, opened 2024-01-31
ITS: fixed mGeometry loading

IsakovAD:
Fixed mGeomtry loading for Cluster and FHR task. 

Now it triggers each time when mGeom object is not initialized;

### DIFF
--- a/Modules/ITS/itsCluster.json
+++ b/Modules/ITS/itsCluster.json
@@ -52,7 +52,7 @@
           "askGRPMagField": "false",
           "askMatLUT": "false",
           "askTime": "false",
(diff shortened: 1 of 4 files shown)


IsakovAD:
@Barthelemy @knopers8 this PR should fix the problem with bad initialization of mGeom file because of no `RunChanged` flag after recent FLP suite upgrade. 

Would it be possible to include it to the next fix?

Barthelemy:
Hi, 
I will create a patch release for it. 
Cheers,
Barth

Barthelemy (approved):

PR #2124, opened 2024-02-01
[WIP] TPC QC: adds occupancy histogram in Cluster task

sbhawani:


Modules/TPC/include/TPC/ClusterVisualizer.h
```diff
@@ -65,6 +65,12 @@ class ClusterVisualizer final : public quality_control::postprocessing::PostProc
   /// \param services Interface containing optional interfaces, for example DatabaseInterface
   void finalize(quality_control::postprocessing::Trigger, framework::ServiceRegistryRef) override;
 
+  template <class T>
+  void makeRadialProfile(o2::tpc::CalDet<T>& calDet, TCanvas* canv, int nbinsY, float yMin, float yMax);
```
wiechula, line 69: Most probably these functions will never be called outside this code. So the can be in the `private` part of the class.
Also, the binning is not needed here, since it is only taken from `mRanges` via the calDet name. Also, the object should not be modified, so it can be `const.
```suggestion
  void makeRadialProfile(const o2::tpc::CalDet<T>& calDet, TCanvas* canv);
```

Modules/TPC/src/ClusterVisualizer.cxx
```diff
@@ -194,6 +209,14 @@ void ClusterVisualizer::update(Trigger t, framework::ServiceRegistryRef)
...
   o2::tpc::painter::makeSummaryCanvases(calDet, int(mRanges[calDet.getName()].at(0)), mRanges[calDet.getName()].at(1), mRanges[calDet.getName()].at(2), false, &vecPtr);
   calDetIter++;
+
+  calDet = clusters.getOccupancy();
+  vecPtr = toVector(mCalDetCanvasVec.at(calDetIter));
+  o2::tpc::painter::makeSummaryCanvases(calDet, int(mRanges[calDet.getName()].at(0)), mRanges[calDet.getName()].at(1), mRanges[calDet.getName()].at(2), false, &vecPtr);
+  calDetIter++;
+  vecPtr = toVector(mCalDetCanvasVec.at(calDetIter));
+  makeRadialProfile(calDet, vecPtr.at(0), int(mRanges[calDet.getName()].at(0)), mRanges[calDet.getName()].at(1), mRanges[calDet.getName()].at(2));
+  calDetIter++;
```
wiechula, line 219: Here is a problem in the logic, since `calDet` is a __reference__ to the `NClusters` cal det. So all subsequent assignments will overwrite the `NClusters` calDet. Also, there is a lot of unnecessary repetitive code here. I propose to add a simple lambda and replace everything below
```c++
auto& clusters = clusterData->getClusters();
```
with
```c++
auto fillCanvases = [&calDetIter, this](const CalDet& calDet) {
  auto vecPtr = toVector(mCalDetCanvasVec.at(calDetIter++));
  const auto& ranges = mRanges[calDet.getName()];
  o2::tpc::painter::makeSummaryCanvases(calDet, int(ranges.at(0)), ranges.at(1), ranges.at(2), false, &vecPtr);
};

  fillCanvases(clusters.getNClusters());
  fillCanvases(clusters.getQMax());

  if (mIsClusters) {
    fillCanvases(clusters.getQTot());
    fillCanvases(clusters.getSigmaPad());
    fillCanvases(clusters.getSigmaTime());
  }

  fillCanvases(clusters.getTimeBin());
  fillCanvases(clusters.getOccupancy());

  makeRadialProfile(calDet, mCalDetCanvasVec.at(calDetIter++).at(0).get());
```

This should be much more compact and easier to read.

Modules/TPC/src/ClusterVisualizer.cxx
```diff
@@ -205,4 +228,64 @@ void ClusterVisualizer::finalize(Trigger t, framework::ServiceRegistryRef)
   }
 }
 
+template <class T>
+void ClusterVisualizer::makeRadialProfile(o2::tpc::CalDet<T>& calDet, TCanvas* canv, int nbinsY, float yMin, float yMax)
+{
+  const std::string_view calName = calDet.getName();
```
wiechula, line 234:
```suggestion
void ClusterVisualizer::makeRadialProfile(const o2::tpc::CalDet<T>& calDet, TCanvas* canv)
{
  const std::string_view calName = calDet.getName();
  const auto& ranges = mRanges[calName.data()];
  
  const int nbinsY = int(ranges.at(0));
  const float yMin = ranges.at(1);
  const float yMax = ranges.at(2);
```

wiechula (changes requested):

wiechula:
@sbhawani , please fix the clang-format

PR #2125, opened 2024-02-01
ITS: cut for real primary particles for efficiency calculation

iravasen:
Before all the efficiencies where not the real efficiencies for primary particles. 
Now we select real primary particles from MC. This helps in removing the contribution from secondaries in the efficiency which is not the object of study of this task.


Barthelemy (approved):

Barthelemy:
I am merging for it to be in the next release.

iravasen:
Thank you!

PR #2127, opened 2024-02-05
init pointer data members to nullptr

shahor02:
@IsakovAD I suspect non-initialized pointers is the reason of the crash on staging:
```
#0  0x00007fad6a2ec7dd in ROOT::Math::Impl::Transform3D<double>::operator() (p=..., this=<optimized out>) at /home/alibuild/sw/SOURCES/QualityControl/v1.131.0/v1.131.0/Modules/ITS/src/ITSFhrTask.cxx:421
#1  ROOT::Math::Impl::Transform3D<double>::operator()<ROOT::Math::Cartesian3D<float> > (p=..., this=<optimized out>) at /home/alibuild/sw/slc8_x86-64/ROOT/v6-30-01-alice2-local2/include/Math/GenVector/Transform3D.h:643
#2  o2::quality_control_modules::its::ITSFhrTask::monitorData (this=0x7fad69a59010, ctx=...) at /home/alibuild/sw/SOURCES/QualityControl/v1.131.0/v1.131.0/Modules/ITS/src/ITSFhrTask.cxx:423
#3  0x00007fadc391db45 in o2::quality_control::core::TaskRunner::run (this=0x2222080, pCtx=...) at /opt/rh/gcc-toolset-12/root/usr/include/c++/12/bits/shared_ptr_base.h:1665
#4  0x00007fadbd89b1c5 in std::function<void (o2::framework::ProcessingContext&)>::operator()(o2::framework::ProcessingContext&) const (__args#0=..., this=<optimized out>) at /opt/rh/gcc-toolset-12/root/usr/include/c++/12/bits/std_function.h:587
#5  operator() (__closure=<optimized out>, action=...) at /home/alibuild/sw/SOURCES/O2/daily-20240125-0100/daily-20240125-0100/Framework/Core/src/DataProcessingDevice.cxx:2355
#6  0x00007fadbd8a9864 in o2::framework::DataProcessingDevice::tryDispatchComputation (ref=..., completed=...) at /home/alibuild/sw/SOURCES/O2/daily-20240125-0100/daily-20240125-0100/Framework/Core/src/DataProcessingDevice.cxx:2405
```

### DIFF
--- a/Modules/ITS/include/ITS/ITSFhrTask.h
+++ b/Modules/ITS/include/ITS/ITSFhrTask.h
@@ -93,7 +93,7 @@ class ITSFhrTask final : public TaskInterface
   int mNThreads = 1;
   std::unordered_map<unsigned int, int> mHitPixelID_Hash[7][48][2][14][14]; // layer, stave, substave, hic, chip
 
-  o2::itsmft::RawPixelDecoder<o2::itsmft::ChipMappingITS>* mDecoder;
+  o2::itsmft::RawPixelDecoder<o2::itsmft::ChipMappingITS>* mDecoder = nullptr;
   ChipPixelData* mChipDataBuffer = nullptr;
   std::vector<ChipPixelData> mChipsBuffer;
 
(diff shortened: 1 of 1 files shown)


IsakovAD:
Thank you for preparing the fix, @shahor02 

Do we need to have the same fix also for Cluster Task? Because we have exactly the same code to initialize Geometry file there

shahor02:
I would say so, one should always init data members.

Barthelemy (approved):

PR #2128, opened 2024-02-09
[HMP] Don't flood the IL in case of decoding error

Barthelemy:
@gvolpe79

### DIFF
--- a/Modules/HMPID/src/HmpidTask.cxx
+++ b/Modules/HMPID/src/HmpidTask.cxx
@@ -233,6 +233,7 @@ void HmpidTask::monitorData(o2::framework::ProcessingContext& ctx)
   mDecoder->init();
   mDecoder->setVerbosity(2); // this is for Debug
                              //  static const Int_t numCham = 7;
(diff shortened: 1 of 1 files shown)


Modules/HMPID/src/HmpidTask.cxx
```diff
@@ -245,7 +246,7 @@ void HmpidTask::monitorData(o2::framework::ProcessingContext& ctx)
       }
       mDecoder->setUpStream(ptrToPayload, (long int)payloadSize);
       if (!mDecoder->decodeBufferFast()) {
-        ILOG(Error, Devel) << "Error decoding the Superpage !" << ENDM;
+        ILOG_INST.log(msgLimit, "Error decoding the Superpage !");
```
knopers8, line 249: is there a way of knowing from the logs, that the future messages of this kind are suppressed?

Barthelemy: When the limit is reached it says: `Warning - [auto-mute] Verbosity threshold exceeded (1 msg < 600 s), discarding next similar messages`

knopers8: cool, thanks!

knopers8 (approved):

PR #2131, opened 2024-02-12
Take any latest trend when resuming. 

knopers8:
It does not really make sense to restrict resuming the trend to the existing run only, since in case of failure, we are not able to recover a task anyway in the same run. The main use-case was always to perform a trend across many runs or even periods, thus we cannot restrict the retrieval to the current activity.

Also, this commit fixes the retrieval by taking the latest object, not the one matching the current timestamp, which would not work with the new validity rules since a few months.


Barthelemy (approved):
agreed

PR #2133, opened 2024-02-13
TOF: adding pads in the HitMap

sofiatomassini:
New pads are added in the Hit map in order to localize the mismatching between the Hit and the Reference map. The pads can be turned on or off in the configuration.
the condition in the checker  is also changed and the number of mismatching is now used to determine the quality of the checker.


njacazio (approved):
Starting ci

njacazio:
@Barthelemy It seems that the ci might be a false positive?

Barthelemy:
yes, spurious errors, merging

PR #2137, opened 2024-02-15
QC-1100 Allow to stop publishing an object without accessing GetName

knopers8:
Effectively, this allows to use stopPublishing on an object allocated on heap after it has been deleted, but still having the pointer, which may avoid some segmentation faults. Unfortunately, this will not avoid segfaults when re-registering an object, since then TObject::GetName has to be called as the only source of the object name.


Barthelemy:
Make sense, thank you

Barthelemy (approved):

PR #2140, opened 2024-02-16
TOF: definition of two configurable thresholds in the checker and new shifter message

sofiatomassini:
two different configurable thresholds are added in the checker. 
The shifter message is modified 
The HitMapNoiseFiltered id used as monitoring object in the checker and in the PostProcessing

### DIFF
--- a/Modules/TOF/src/CheckHitMap.cxx
+++ b/Modules/TOF/src/CheckHitMap.cxx
@@ -24,10 +24,6 @@
 #include "DataFormatsTOF/TOFFEElightInfo.h"
 #include "TOFBase/Geo.h"
 
-// ROOT
-#include "TLine.h"
-#include "TLatex.h"
-
 using namespace std;
 
 namespace o2::quality_control_modules::tof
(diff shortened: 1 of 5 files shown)


njacazio (approved):

njacazio (approved):

njacazio:
@Barthelemy ci looks spurious again

sofiatomassini:
@Barthelemy  @knopers8 could you have a look? this seems to be spurious, it would be important to have this merged for the next release. thank you!

Modules/TOF/src/CheckHitMap.cxx
```diff
@@ -136,17 +136,17 @@ void CheckHitMap::beautify(std::shared_ptr<MonitorObject> mo, Quality checkResul
     ILOG(Error, Support) << "Cannot beautify MO " << mo->getName() << " which is not of type " << getAcceptedType() << ENDM;
     return;
   }
-  if (mo->getName() == mAcceptedName) {
+  if (1) {
```
knopers8, line 139: Shouldn't the `if` be removed in this case?

njacazio: that's true, I could not reapply the formatting online. Will do at the next iteration. Thanks!

PR #2145, opened 2024-02-20
Increase newobject warning log level

knopers8:


knopers8:
There is basically a flood of those when starting any environment in production, they are not as important anymore.

Barthelemy (approved):
Thank you. I fixed the formatting.

PR #2146, opened 2024-02-20
QC-1102 ObjectManager should survive being fed with nullptr

knopers8:
Fixes an issue reported in QC-1102.


Framework/src/ObjectsManager.cxx
```diff
@@ -61,6 +61,10 @@ ObjectsManager::~ObjectsManager()
 
 void ObjectsManager::startPublishing(TObject* object)
 {
+  if (!object) {
+    ILOG(Warning, Support) << "A nullptr provided to ObjectManager::stopPublishing" << ENDM;
```
Barthelemy, line 65:
```suggestion
    ILOG(Warning, Support) << "A nullptr provided to ObjectManager::startPublishing" << ENDM;
```

Barthelemy (changes requested):
Thank you for these extra guards. There is just a typo in one of the logs.

PR #2148, opened 2024-02-21
QC-1103 BigScreen: create the canvas, paves and labels in configure()

knopers8:
mCanvas, mPaves and mLabels are now created just once, during configuration. Thus, we avoid any issues with republishing mCanvas and making sure it does not delete something already deleted while going 2nd time through initialize() (START STOP START)


knopers8:
@aferrero2707 Could you have a look as well in case I am missing something here?

aferrero2707:
@knopers8 I do not see any potential problems by moving the canvas initialization in `configure()`. For me it is ok.

PR #2149, opened 2024-02-21
QC-1104 Reset sh. ptr before creating a new histogram in MCH PP

knopers8:
This makes sure that the old object are deleted before the new ones are created, which prevents us from potentially sharing the same resources with both old and new, and prevents ROOT from producing warnings about histograms with duplicate names and potential memory leaks.


knopers8:
@aferrero2707 this fixes my laptop reproducer. I am not entirely sure it will fix P2 as well, but it should be a good first step to do.

aferrero2707:
@knopers8 thanks, it perfectly makes sense! 
There is one more source file with a similar structure that most likely requires a similar treatment: [DecodingPostProcessing.cxx](https://github.com/AliceO2Group/QualityControl/blob/master/Modules/MUON/MCH/src/DecodingPostProcessing.cxx).

knopers8:
indeed, i will then include it in this PR

Barthelemy (approved):

PR #2152, opened 2024-02-24
[TRD]: Check on hc2d

deependra170598:
Hi @martenole and others!
In this PR, I tried to implement check on Half Chamber status.
It contains two sub checks...

- Checks if any bin of `trackletsperHC2D`  is empty
- Checks if any chamber is masked.

Any masked half-chamber or empty chamber results in bad quality otherwise good.

### DIFF
--- a/Modules/TRD/src/CheckOnHc2d.cxx
+++ b/Modules/TRD/src/CheckOnHc2d.cxx
@@ -27,6 +27,7 @@
 
 using namespace std;
 using namespace o2::quality_control;
+using Helper = o2::trd::HelperMethods;
 
 namespace o2::quality_control_modules::trd
 {
@@ -40,10 +41,23 @@ void CheckOnHc2d::configure()
     mTimestamp = o2::ccdb::getCurrentTimestamp();
     ILOG(Debug, Support) << "configure() : using default timestam of now = " << mTimestamp << ENDM;
   }
-  auto& mgr = o2::ccdb::BasicCCDBManager::instance();
-  mgr.setTimestamp(mTimestamp);
 
-  mChamberStatus = mgr.get<std::array<int, MAXCHAMBER>>("TRD/Calib/DCSDPsFedChamberStatus");
+  if (auto param = mCustomParameters.find("ccdbPath"); param != mCustomParameters.end()) {
+    mCcdbPath = mCustomParameters["ccdbPath"];
+    ILOG(Debug, Support) << "configure() : using ccdb path = " << mCcdbPath << ENDM;
+  } else {
+
+    ILOG(Warning, Trace) << "provided ccdb path in config not found" << ENDM;
+    mCcdbPath = "TRD/Calib/DCSDPsFedChamberStatus";
+    ILOG(Warning, Trace) << "using ccdb path =" << mCcdbPath << ENDM;
+  }
+
+  // auto& mgr = o2::ccdb::BasicCCDBManager::instance();
+  // mgr.setTimestamp(mTimestamp);
+  // mChamberStatus = mgr.get<std::array<int, MAXCHAMBER>>("TRD/Calib/DCSDPsFedChamberStatus");
+
+  mChamberStatus = retrieveConditionAny<std::array<int, MAXCHAMBER>>(mCcdbPath, {}, mTimestamp);
+
   if (mChamberStatus == nullptr) {
     ILOG(Info, Support) << "mChamberStatus is null, no chamber status to display" << ENDM;
   }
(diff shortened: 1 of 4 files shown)


Modules/TRD/src/CheckOnHc2d.cxx
```diff
@@ -0,0 +1,138 @@
...
+{
+  if (auto param = mCustomParameters.find("ccdbtimestamp"); param != mCustomParameters.end()) {
+    mTimestamp = std::stol(mCustomParameters["ccdbtimestamp"]);
+    ILOG(Debug, Support) << "configure() : using ccdbtimestamp = " << mTimestamp << ENDM;
+  } else {
+    mTimestamp = o2::ccdb::getCurrentTimestamp();
+    ILOG(Debug, Support) << "configure() : using default timestam of now = " << mTimestamp << ENDM;
+  }
+  auto& mgr = o2::ccdb::BasicCCDBManager::instance();
+  mgr.setTimestamp(mTimestamp);
```
martenole, line 44: Should checks still use the BasicCCDBManager? Should one not use the DPL CCDB fetching mechanism instead? This we do for example in the TrackletsTask or DigitsTask

Modules/TRD/src/CheckOnHc2d.cxx
```diff
@@ -0,0 +1,138 @@
...
+      }
+
+      result1 = Quality::Good;
+
+      for (int i = 0; i < h->GetNbinsX(); i++) {
+        for (int j = 0; j < h->GetNbinsY(); j++) {
+          double content = h->GetBinContent(i, j);
+          if (content == 0) {
+            result1 = Quality::Bad;
+            result1.addReason(FlagReasonFactory::Unknown(), "some half chambers are empty");
```
martenole, line 75: But if the bin is empty and the half-chamber is already flagged as bad, then the quality should not be bad. It should still be good.

Modules/TRD/src/CheckOnHc2d.cxx
```diff
@@ -0,0 +1,138 @@
...
+        }
+      }
+
+      result2 = Quality::Good;
+
+      for (int hcid = 0; hcid < MAXCHAMBER * 2; ++hcid) {
+        if (TRDHelpers::isHalfChamberMasked(hcid, mChamberStatus)) {
+          ILOG(Debug, Trace) << "Masked half Chamber id =" << hcid << ENDM;
+          result2 = Quality::Bad;
+          result2.addReason(FlagReasonFactory::Unknown(), "some chmabers are masked");
```
martenole, line 87: For some masked chambers everything is still OK. We will always have that, so your check would never yield good quality. I would change quality to bad based on masked chambers only if we have either a full sector masked or more than 40% of all half-chambers masked from DCS side

martenole (commented):
Thanks for the work Deependra, but I think we need to refine the check a bit. And please make check if other checkers are not using the DPL CCDB fetching mechanism as well.

Modules/TRD/src/CheckOnHc2d.cxx
deependra170598: As I discussed with Piotr, He told me this is not possible.

Modules/TRD/src/CheckOnHc2d.cxx
deependra170598: I tried to address this also. let me know if something wrong.

martenole:
Hi @deependra170598 do you have time for a debug session maybe tomorrow at some point? Concerning the timestamp I saw this code in MFT where they obtain the time directly from the MO. That makes more sense to me then setting it as parameter..
https://github.com/AliceO2Group/QualityControl/blob/8991a03f8e48bff0369a9c45c08e83063f46016d/Modules/MFT/src/QcMFTDigitCheck.cxx#L131-L134

deependra170598:
Hi Ole, I used the same way of fetching ccdbTimeStamp as you mentioned. Right now I am getting the same flag message as shown in the above picture for the chamber map shown in this picture.
Screenshot 2024-03-15 023917

Modules/TRD/src/CheckOnHc2d.cxx
```diff
@@ -0,0 +1,204 @@
...
+#include "DataFormatsTRD/HelperMethods.h"
+
+using namespace std;
+using namespace o2::quality_control;
+using Helper = o2::trd::HelperMethods;
+
+namespace o2::quality_control_modules::trd
+{
+
+void CheckOnHc2d::configure()
```
martenole, line 35: do we need this method at all?

Modules/TRD/src/CheckOnHc2d.cxx
```diff
@@ -0,0 +1,204 @@
...
+      result2 = Quality::Good;
+      int nMaskedHC = 0;        // total number of masked half-chamber
+      int nMaskedHChInASec = 1; // total number of masked half-chamber in a sector
+      int currentSector = -1, oldSector = -1;
+      for (int hcid = 0; hcid < MAXCHAMBER * 2; ++hcid) {
+        currentSector = Helper::getSector(hcid / 2);
+        if (TRDHelpers::isHalfChamberMasked(hcid, mChamberStatus)) {
+          ILOG(Debug, Trace) << "Masked half Chamber id =" << hcid << ENDM;
+          nMaskedHC++;
+          if (nMaskedHC > (NCHAMBER * 2 * 0.4)) { // if nMaskedHC is greator than 40% of all chambers
```
martenole, line 137: The 40% should be a parameter that we can change from the json file

Modules/TRD/src/CheckOnHc2d.cxx
```diff
@@ -0,0 +1,204 @@
...
+
+std::string CheckOnHc2d::getAcceptedType() { return "TH2"; }
+
+void CheckOnHc2d::beautify(std::shared_ptr<MonitorObject> mo, Quality checkResult)
+{
+  if (mo->getName() == "trackletsperHC2D") {
+    auto* h = dynamic_cast<TH2F*>(mo->getObject());
+
+    if (checkResult == Quality::Good) {
+      h->SetFillColor(kGreen);
```
martenole, line 176: Does one see it if the histogram fill color for a TH2 changes if we draw it with COLZ option? I think we need a text box here

martenole (commented):
I am adding some small comments now. But I think we can also go through this PR together next week

deependra170598:
Right now I am closing this PR. I will first merge TrackletsCheck and TrackletCountsCheck then will work on 2DHC check.

PR #2156, opened 2024-02-27
ITS: fix of wrong vector size infologger message

IsakovAD:
Added template for float variables in string parser which should fix problematic vector size error messages


iravasen (approved):

iravasen:
Thanks @IsakovAD, all good for me.

PR #2162, opened 2024-02-29
[WIP] TRD: Trending of calibration parameters

maroozm:
This PR will provide trending of calibration parameters (vdrift, gain etc) from ccdb using `ReductorConditionAny`


martenole (commented):
Thanks a lot @maroozm this looks like a good start to me. Two things I would like to have as well in this PR:
- Can you add also trending for ExB? You can retrieve that from the same object as vDrift so it should be straight-forward. Then I can see the changes needed for one additional calibration value
- Can you in addition to the graph also plot a 2D histogram with run number or time on the x-axis and for every time stamp the different values of vDrift on the y-axis? This would allow for a better understanding of mean and RMS. For example for all chambers where the fit is not possible we put the default values

maroozm:
Thanks @martenole :)
I added both the plots: exb and vdrift values for a specific run number.
the second one will look like:
image

martenole:
Thanks a lot @maroozm ! Are only graphs supported? Because I would prefer a 2D histogram for the second plot. From the graph its impossible to judge how many entries are at which drift velocity

maroozm:
> orted? Because I would prefer a 2D histogram for the second plot. From the graph its impossible to judge how many entries are at which drift velocity

Thanks. I checked, but it seems it only produces graphs as per `o2::quality_control::postprocessing::TrendingTask`. I can implement our own trending task for TRD to produce this.

martenole:
Thanks, good that 2D histogram is supported. But why do we have negative vDrift values? In the graph this was not the case and it does not make sense.

maroozm:
> Thanks, good that 2D histogram is supported. But why do we have negative vDrift values? In the graph this was not the case and it does not make sense.

hmm. am looking into it. but i don't know, i mean, when i draw only one axis, its fine, as here.
image
i also checked the tree that reductor creates, it also doesn't have any neg values.

Modules/TRD/calibParamTRD.json
```diff
@@ -0,0 +1,82 @@
...
+          },
+          {
+            "name": "vdrift_trend2",
+            "title": "vDrift trend",
+            "varexp": "CalVdriftExB.vdrift[]:runNumber",
+            "selection": "",
+            "option": "colz logz",
+            "graphYRange": "-2:3.5",
+            "graphXRange": "",
+            "graphAxisLabel": "vdrift:time"
```
martenole, line 56: But this still looks like the definition of a graph. Can't you also set the number of bins for example?

maroozm: it doesn't seem to have any option, related to binning. i am working on implementing our own trending task 🙂

maroozm:
Hi Ole, these are the three objects, we have right now :)
image
image
image

martenole:
Hi @maroozm could you please rebase your branch on top of master and resolve the merge conflicts?

maroozm:
> Hi @maroozm could you please rebase your branch on top of master and resolve the merge conflicts?

Hi @martenole, thanks
done :)

maroozm:
Hi. This can be merged :)

martenole (approved):
This is fine to be merged from my side @knopers8

knopers8 (approved):

PR #2166, opened 2024-03-05
Tips for memory usage reduction

knopers8:


Barthelemy (approved):
Thanks very clear and instructing.

PR #2167, opened 2024-03-06
Add histo for TrendingTask

ValerieRamillien:


Modules/MUON/MID/src/DigitsQcCheck.cxx
```diff
@@ -120,14 +120,20 @@ Quality DigitsQcCheck::check(std::map<std::string, std::shared_ptr<MonitorObject
...
+        nBadLB = 0;
+        maxVal = 0;
+        minVal = 1000;
         auto histo = static_cast<TH2F*>(item.second->getObject());
         mHistoHelper.normalizeHistoTokHz(histo);
         for (int bx = 1; bx < 15; bx++) {
           for (int by = 1; by < 37; by++) {
             if (!((bx > 6) && (bx < 9) && (by > 15) && (by < 22))) { // central zone empty
               double val = histo->GetBinContent(bx, by);
+              if (val > maxVal)
```
Barthelemy, line 133: In general it is recommended to always use `{` and `}` even when there is a single line.

Modules/MUON/MID/src/DigitsQcCheck.cxx
```diff
@@ -159,6 +165,22 @@ Quality DigitsQcCheck::check(std::map<std::string, std::shared_ptr<MonitorObject
         mQualityMap[item.second->getName()] = qual;
       } // if mNTFInSeconds > 0.
     }
+    if (item.second->getName() == "NbLBEmpty") {
+      auto histo = static_cast<TH1F*>(item.second->getObject());
```
Barthelemy, line 169: why not use a dynamic cast and make sure that the result is not a nullptr ?

Barthelemy (commented):
Thank you, I have put a couple of comment that I will let you address.

Modules/MUON/MID/src/DigitsQcCheck.cxx
Barthelemy: In general, you could try to use 
```
      auto histo = dynamic_cast<TH1F*>(item.second->getObject());
      if (histo) {
             // do something with the plot
      } else { 
             // something went wrong
      }
```
I let you investigate this safer option and perhaps submit another PR as to not block this one. This could apply to all the static casts.

Barthelemy (approved):

PR #2169, opened 2024-03-08
[QC-1125] Avoid MonitorObjectCollection to grow in size

aferrero2707:
Currently the published objects are inserted in the `MonitorObjectCollection` via the `TObjArray::Add()` method, which always appends new elements at the end of the array.

When un-publishing objects, they are removed via the `TObjArray::Remove()` method, which does not reduce the size of the TObjArray. Instead, the removed slots are set to `nullptr`.

The proposed solution is to replace `TObjArray::Add()` with `TObjArray::AddAtFree()` in `ObjectsManager::startPublishing()`, which will re-use NULL slots instead of systematically appending new ones at the end.

### DIFF
--- a/Framework/src/ObjectsManager.cxx
+++ b/Framework/src/ObjectsManager.cxx
@@ -131,6 +131,7 @@ void ObjectsManager::stopPublishing(const string& objectName)
 {
   auto* mo = dynamic_cast<MonitorObject*>(getMonitorObject(objectName));
   mMonitorObjects->Remove(mo);
+  mMonitorObjects->Compress();
 }
 
 void ObjectsManager::stopPublishingAll()


knopers8 (changes requested):
Thanks @aferrero2707 ! Yet another ROOT intricacy that I was not now aware of...

I would also propose to use `Compress()` after each `Remove()`, which according to the doc should "Remove empty slots from array. ". I believe the performance impact should be fairly minimal, while we would benefit from not storing there `nullptr` at any time, thus avoiding issues with storing the MonitorObjectCollection to files at any moment.

aferrero2707:
@knopers8 indeed, calling `Compress()` after `Remove()` should have the same effect. In this case one can keep using `Add()`. I have checked locally that `Compress()` does what is advertized, and I changed the code accordingly.

knopers8:
Great, thank you!

PR #2170, opened 2024-03-08
[QC-1031]: changed origin for QC Check

justonedev1:


Framework/src/Check.cxx
```diff
@@ -260,14 +260,17 @@ CheckConfig Check::extractConfig(const CommonSpec& commonSpec, const CheckSpec&
...
   };
 }
 
-framework::OutputSpec Check::createOutputSpec(const std::string& checkName)
+framework::OutputSpec Check::createOutputSpec(const std::string& detector, const std::string& checkName)
 {
-  return { "QC", createCheckDataDescription(checkName), 0, framework::Lifetime::Sporadic };
+  using Origin = o2::header::DataOrigin;
+  Origin header;
+  header.runtimeInit(std::string{ "C" }.append(detector.substr(0, Origin::size)).c_str());
```
knopers8, line 272:
```suggestion
  header.runtimeInit(std::string{ "C" }.append(detector.substr(0, Origin::size - 1)).c_str());
```
If you get a 4-letter detector code, won't you pass a 5-letter string to `runtimeInit`?

knopers8: Also, I would propose to extract the logic into a separate `createCheckDataOrigin` method.

PR #2171, opened 2024-03-08
[QC 1123]: using hash when description is too long

justonedev1:


knopers8:
There are some tests failing on macOS. It's not clear to me why only there...

Framework/include/QualityControl/AggregatorRunner.h
```diff
@@ -109,7 +109,7 @@ class AggregatorRunner : public framework::Task
   static framework::DataProcessorLabel getLabel() { return { "qc-aggregator" }; }
   static std::string createAggregatorRunnerIdString() { return "qc-aggregator"; };
   static std::string createAggregatorRunnerName();
-  static header::DataDescription createAggregatorRunnerDataDescription(const std::string& aggregatorName);
+  static header::DataDescription createAggregatorRunnerDataDescription(const std::string& aggregatorName, size_t hashSize = 4);
```
knopers8, line 112:
```suggestion
  static header::DataDescription createAggregatorRunnerDataDescription(const std::string& aggregatorName);
```
This could be discussed, but I think it would be safer to assume always the same hash length for a given QC actor type. Otherwise, someone who wants to subscribe to any QC Data Processor would have to check the `InfrastructureGenerator` implementation (instead of just relying on information in a header) to use the correct hash length and would be exposed to incompatibilities if we change it.

I would propose that hashLength is a constant inside each of such `create*DataDescription` methods.

Framework/src/HashDataDescription.cxx
```diff
@@ -0,0 +1,45 @@
...
+o2::header::DataDescription createDataDescription(const std::string& name, size_t hashLength)
+{
+  o2::header::DataDescription description{};
+  if (name.size() <= o2::header::DataDescription::size) {
+    description.runtimeInit(name.c_str());
+    return description;
+  } else {
+
+    std::stringstream ss{};
+    ss << std::hex << std::hash<std::string>{}(name);
```
knopers8, line 37: Perhaps mac failures are related to the implementation of `std::hash`? I think it would be good to have the hash to return the same result on any platform and any execution, because we might generate a workflow template on one OS and use it in a different one.

knopers8 (changes requested):
Thanks a lot! I have some improvement suggestions, but it's good to go in after they are addressed in any way.

Framework/src/HashDataDescription.cxx
justonedev1: yes, it does... I thought that it might happen, but I still wanted to try it

justonedev1: I had a implementation of djb2 hash, but I too it away when I saw that `std::hash` is used for timers.

Framework/include/QualityControl/AggregatorRunner.h
justonedev1: sure, I will change it for all of the factory methods. I just saw there is written N for hash length in ticket, so I kinda propagated it everywhere...

knopers8: to be clear, the general function in `HashDataDescription` should still have the argument, the component-specific ones should use constants.

Framework/src/HashDataDescription.cxx
knopers8: then indeed it's weird that it does not fail for timers...

justonedev1: https://github.com/AliceO2Group/QualityControl/blob/master/Framework/src/TaskRunner.cxx#L307-L320

I guess that there is not a test for timers, that tests for exact task name

knopers8 (approved):
Thank you!

PR #2173, opened 2024-03-11
TOF: change fetching of conditions from the ccdb

sofiatomassini:
the url to get the reference map in the ccdb is now set correctly. 
@Barthelemy  could you please merge this PR? Thank you!


njacazio:
@Barthelemy we were wondering if this is how we should set up the CCDB call to customize the URL. Thanks and cheers,
N.

Barthelemy (changes requested):
Hi, 

I understand that you want to retrieve something from the CCDB ? 

There are several ways of doing it. You should not set the url yourself. The URL is set in the config file. 

The documentation is here: 
https://github.com/AliceO2Group/QualityControl/blob/master/doc/Advanced.md#accessing-objects-in-ccdb

The recommended way is to use Lifetime::Condition DPL input. There are many examples in the Modules. 

We don't recommend but don't forbid either to use `retrieveConditionAny` which would let you retrieve the CCDB data as well. It is ready to use and does not require any extra configuration.

knopers8:
PostProcessingTasks do not take DPL inputs, so in this case `retrieveConditionAny` should be used.

sofiatomassini:
Indeed, however the `PostProcessingInterface.h` does not depend on `UserCodeInterface.h` where the `retrieveConditionAny` is defined. For instance in `Modules/CPV/src/PhysicsTask.cxx` it can be used  because the `TaskInterface.h` is dependent on `UserCodeInterface.h`. Did I understand correctly how It works? Thank you !

knopers8:
PostProcessingInterface does inherit UserCodeInterface since two months: https://github.com/AliceO2Group/QualityControl/blob/07bdae6d68188960dd14c5be555d149c4bd17ee9/Framework/include/QualityControl/PostProcessingInterface.h#L37
Perhaps you have to rebase with the master branch?

Barthelemy:
I am updating the documentation accordingly

Modules/TOF/include/Base/MessagePad.h
```diff
@@ -35,13 +35,13 @@ namespace o2::quality_control_modules::tof
 
 struct MessagePad {
   float mPadLowX = 0.6;                 /// Position of the message PAD in low x
-  float mPadLowY = 0.5;                 /// Position of the message PAD in low y
```
njacazio, line 38: don't change these

Modules/TOF/include/Base/MessagePad.h
```diff
@@ -140,7 +140,7 @@ struct MessagePad {
     ILOG(Info, Devel) << "Message pad '" << mName << "' is enabled" << ENDM;
 
     mMessagePad = new TPaveText(mPadLowX, mPadLowY, mPadHighX, mPadHighY, padOpt);
-    mMessagePad->SetName(Form("%s_msg", histogram->GetName()));
```
njacazio, line 143: this is ok

Modules/TOF/include/Base/MessagePad.h
```diff
@@ -35,13 +35,13 @@ namespace o2::quality_control_modules::tof
...
-  float mPadLowY = 0.5;                 /// Position of the message PAD in low y
+  float mPadLowY = 0.9;                 /// Position of the message PAD in low y
   float mPadHighX = 0.9;                /// Position of the message PAD in high x
-  float mPadHighY = 0.75;               /// Position of the message PAD in high y
+  float mPadHighY = 1.5;                /// Position of the message PAD in high y
   std::vector<std::string> mMessages{}; /// Message to print on the pad, this is reset at each call of MakeMessagePad
   TPaveText* mMessagePad = nullptr;     /// Text pad with the messages
   int mEnabledFlag = 1;                 /// Flag to enable or disable the pad
-  const std::string mName = "";         /// Name of the message pad, can be used to identify the pad if multiple are used
+  const std::string mName = "msg";      /// Name of the message pad, can be used to identify the pad if multiple are used
```
njacazio, line 44: this is ok

Modules/TOF/src/PostProcessHitMap.cxx
```diff
@@ -60,22 +63,24 @@ void PostProcessHitMap::initialize(Trigger, framework::ServiceRegistryRef servic
 {
   // Setting up services
   mDatabase = &services.get<o2::quality_control::repository::DatabaseInterface>();
+  o2::ccdb::BasicCCDBManager::instance().setURL(mCCDBUrl);
 }
 
 void PostProcessHitMap::update(Trigger t, framework::ServiceRegistryRef services)
 {
   // Getting the hit map
-  const auto mo = mDatabase->retrieveMO(mCCDBPath, mCCDBPathObject, t.timestamp, t.activity);
```
njacazio, line 68: Don't change this

Modules/TOF/src/PostProcessHitMap.cxx
```diff
@@ -60,22 +63,24 @@ void PostProcessHitMap::initialize(Trigger, framework::ServiceRegistryRef servic
...
     return;
   }
   // Getting the reference map
-  auto& ccdb = services.get<repository::DatabaseInterface>();
-  map<string, string> metadata; // can be empty
-  const auto* refmap = static_cast<o2::tof::TOFFEElightInfo*>(ccdb.retrieveAny(typeid(o2::tof::TOFFEElightInfo), mRefMapCcdbPath, metadata, mRefMapTimestamp));
+  const o2::tof::TOFFEElightInfo* refmap = nullptr;
+  refmap = o2::quality_control::core::UserCodeInterface::retrieveConditionAny<o2::tof::TOFFEElightInfo>(mRefMapCcdbPath);
 
+  // const auto* refmap = o2::ccdb::BasicCCDBManager::instance().getForTimeStamp<o2::tof::TOFFEElightInfo>(mRefMapCcdbPath, mRefMapTimestamp);
```
njacazio, line 82: This can be cleaned

Modules/TOF/src/PostProcessHitMap.cxx
```diff
@@ -24,7 +24,10 @@
 #include "QualityControl/MonitorObject.h"
 #include "QualityControl/QcInfoLogger.h"
 #include "TOF/PostProcessHitMap.h"
+#include "CCDB/CcdbApi.h"
```
njacazio, line 27: These new includes are not needed then, right?

Modules/TOF/src/PostProcessHitMap.cxx
```diff
@@ -46,8 +49,8 @@ void PostProcessHitMap::configure(const boost::property_tree::ptree& config)
   const std::string baseJsonPath = "qc.postprocessing." + getID() + ".customization.";
   mCCDBPath = config.get<std::string>(baseJsonPath + "CCDBPath", "TOF/MO/TaskDigits/");
   ILOG(Info, Support) << "Setting CCDBPath to " << mCCDBPath << ENDM;
-  mCCDBPathObject = config.get<std::string>(baseJsonPath + "CCDBPathObject", "HitMapNoiseFiltered");
```
njacazio, line 49: Keep both

Modules/TOF/src/PostProcessHitMap.cxx
```diff
@@ -60,22 +63,24 @@ void PostProcessHitMap::initialize(Trigger, framework::ServiceRegistryRef servic
 {
   // Setting up services
   mDatabase = &services.get<o2::quality_control::repository::DatabaseInterface>();
+  o2::ccdb::BasicCCDBManager::instance().setURL(mCCDBUrl);
```
njacazio, line 66: This might not be needed

njacazio (changes requested):

njacazio (approved):

njacazio:
Ok @Barthelemy it should be good to merge

Barthelemy:
why do you need `mCCDBUrl` ?

njacazio:
Indeed it was not needed

Barthelemy (approved):

njacazio (approved):

PR #2174, opened 2024-03-12
MCH: reduced memory footprint of Digits task

aferrero2707:
- experts plots are only allocated if needed
- occupancy TH2 ratio now uses uniform scaling instead of bin-by-bin scaling
- reduced number of bins for orbits plot

This should reduce the memory footprint by approximately a factor of 3. The changes have no impcat on the contents of the generated plots, apart from the different binning of the orbits plot.

### DIFF
--- a/Modules/MUON/MCH/src/TH2ElecMapReductor.cxx
+++ b/Modules/MUON/MCH/src/TH2ElecMapReductor.cxx
@@ -187,7 +187,9 @@ void TH2ElecMapReductor::update(TObject* obj)
 
       deNumPads[cathode][deIndex] += 1;
 
-      Float_t stat = hr->getDen()->GetBinContent(i, j);
+      // here we assume that if the number of bins differs between numerator and denominator,
+      // the denominator contains a single common scaling factor in the first bin (uniform scaling)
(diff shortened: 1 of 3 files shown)


Modules/MUON/MCH/src/TH2ElecMapReductor.cxx
```diff
@@ -187,7 +187,9 @@ void TH2ElecMapReductor::update(TObject* obj)
 
       deNumPads[cathode][deIndex] += 1;
 
-      Float_t stat = hr->getDen()->GetBinContent(i, j);
+      // here we assume that if the number of bins differs between numerator and denominator,
+      // the denominator contains a single common scaling factor in the first bin (uniform scaling)
+      Float_t stat = (hr->getNum()->GetNcells() == hr->getDen()->GetNcells()) ? hr->getDen()->GetBinContent(i, j) : hr->getDen()->GetBinContent(1, 1);
```
Barthelemy, line 192: I did not notice it before, but for the future please use "normal" modern c++ types instead of ROOT types.

Barthelemy (approved):
thank you. 

I have one comment but no need to change it now in this PR.

PR #2178, opened 2024-03-14
[QC-1118] Add option to entirely disable the infologger

Barthelemy:


Framework/src/QcInfoLogger.cxx
```diff
@@ -59,6 +60,13 @@ void QcInfoLogger::setPartition(const std::string& partitionName)
   ILOG(Debug, Devel) << "IL: Partition set to " << partitionName << ENDM;
 }
 
+void QcInfoLogger::disable(bool disabled)
+{
+  QcInfoLogger::disabled = disabled;
```
knopers8, line 65: Since the method lets you also enable IL by setting `disabled` to `false`, the filtering settings should come back to the selected values in the config file.

Alternatively, you could make the `disable()` allow only to disable, but not the other way.

knopers8 (changes requested):
Thanks. I am wondering if this is going to address the problem of the error message. If it appears when a socket or a singleton is created, will enabling full filtering make it dissappear? This being said, I don't know where exactly it comes from.

Also, this is not exactly related to the PR, but I am becoming confused what the structure `DiscardFileParameters` is for, as you use it also to set how the logs are filtered regardless of redirecting them to files. Also it has members `debug` and `debugInDiscardFile` which suggest that the first one does not apply to the log files. Perhaps it should be renamed to `InfoLoggerParameters`?

Framework/src/QcInfoLogger.cxx
Barthelemy: Yes, let's not aim for the moon, I will only let the disabling :-D

Barthelemy:
Concerning `DiscardFileParameters` I agree that it is confusing. A better name would be `DiscardParameters` because all parameters are related to this topic. I am also adding some comments. I will take the opportunity to do it here.

Barthelemy:
@knopers8 I pushed the changes. Now, I don't know if it will prevent the error message  when a socket or a singleton is created but that's the best I can do for now.

knopers8 (approved):
Well, `DiscardParemeters` as a name is still not clear about the fact that it concerns discarding logs, but is not related to `CustomParameters` for example. So please consider renaming it to e.g. `LogDiscardParameters` if you agree with me. It's not very important though, so I am fine with merging it as it is otherwise.

Barthelemy:
Agreed, I will rename it in this PR and then we can merge

Barthelemy:
done

PR #2180, opened 2024-03-15
use correct check data origin as inputs of aggregators

knopers8:


knopers8:
still testing on my laptop to be sure, but this should address the issue in staging

Barthelemy (approved):

Barthelemy:
thank you, seems ok. I let you merge when you are ready.

PR #2190, opened 2024-03-19
QC-1132 Correct validity in PostProcessingTasks across run restart

knopers8:
This makes sure that the SOR&EOR timestamps are retrieved from ECS and passed to the user task with UserOrControl trigger, but they are not used to establish the validity of produced objects. Also, we do not publish objects if the validity is going to be invalid, since they will not be stored anyway. We produce appropriate warnings in such case.

The root cause for having EOR timestamp earlier than SOR for the 2nd run was discovered to come from ECS sending incorrect values. However, with this commit, these values will not be used anyway for determining objects validity.


Barthelemy (approved):
ok, and indeed this might fix the issue seen by TPC

PR #2192, opened 2024-03-19
GLO: updated BigScreen task

aferrero2707:
This PR improves and simplifies the Big Screen task:
- moved canvas object to a separate source file
- added "old" state for objects that exist for the current activity but are older than the maximum age defined in the configuration
- moved canvas publishing in the `initialize()` method
- use the extended task parameters for the configuration

The modifications are split is few commits, to hopefully simplify the review process.

### DIFF
--- a/Modules/Common/include/Common/BigScreen.h
+++ b/Modules/Common/include/Common/BigScreen.h
@@ -47,7 +47,7 @@ class BigScreen : public quality_control::postprocessing::PostProcessingInterfac
   /// \brief maximum allowed age of quality objects, in seconds (default: 10 minutes)
   int mMaxObjectTimeShift{ 600 };
   /// \brief read quality objects from all runs
-  int mIgnoreActivity{ 0 };
+  bool mIgnoreActivity{ false };
   /// \brief configuration parameters
   BigScreenConfig mConfig;
   /// \brief canvas with human-readable quality states
--- a/Modules/Common/include/Common/BigScreen.h
+++ b/Modules/Common/include/Common/BigScreen.h
@@ -20,12 +20,8 @@
 
 #include "QualityControl/PostProcessingInterface.h"
 #include "Common/BigScreenConfig.h"
+#include "Common/BigScreenCanvas.h"
 #include "QualityControl/Quality.h"
-#include <TCanvas.h>
-#include <TPaveText.h>
-#include <TText.h>
-#include <string>
-#include <map>
 
 namespace o2::quality_control_modules::common
 {
@@ -50,14 +46,8 @@ class BigScreen : public quality_control::postprocessing::PostProcessingInterfac
  private:
   /// \brief configuration parameters
   BigScreenConfig mConfig;
-  /// \brief colors associated to each quality state (Good/Medium/Bad/Null)
-  std::unordered_map<std::string, int> mColors;
-  /// \brief
-  std::map<std::string, std::unique_ptr<TText>> mLabels;
-  /// \brief
-  std::map<std::string, std::unique_ptr<TPaveText>> mPaves;
-  /// \brief canvas with human-readable quality states and messages
-  std::unique_ptr<TCanvas> mCanvas;
+  /// \brief canvas with human-readable quality states
+  std::unique_ptr<BigScreenCanvas> mCanvas;
 };
 
 } // namespace o2::quality_control_modules::common
--- a/Modules/Common/include/Common/BigScreen.h
+++ b/Modules/Common/include/Common/BigScreen.h
@@ -44,6 +44,10 @@ class BigScreen : public quality_control::postprocessing::PostProcessingInterfac
   void finalize(quality_control::postprocessing::Trigger, framework::ServiceRegistryRef) override;
 
  private:
+  /// \brief maximum allowed age of quality objects, in seconds (default: 10 minutes)
+  int mMaxObjectTimeShift{ 600 };
+  /// \brief read quality objects from all runs
+  int mIgnoreActivity{ 0 };
   /// \brief configuration parameters
   BigScreenConfig mConfig;
   /// \brief canvas with human-readable quality states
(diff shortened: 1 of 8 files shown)


aferrero2707:
@Barthelemy @knopers8 apart from making the BigScreen canvas reusable, the new code is hopefully simpler, more readable and easier to maintain.

Modules/Common/include/Common/BigScreenCanvas.h
```diff
@@ -0,0 +1,64 @@
...
+struct BigScreenElement;
+
+class BigScreenCanvas : public TCanvas
+{
+ public:
+  BigScreenCanvas(std::string name, std::string title, int nRows, int nCols, int borderWidth);
+  ~BigScreenCanvas() = default;
+
+  /// \brief add a box in the canvas at a given index, with "name" displayed above the box
+  void add(std::string name, int index);
```
knopers8, line 38:
```suggestion
  void addBox(std::string name, int index);
```
How about addBox then?

Also, could you document how are the boxes indexed? I.e. does the index start at 0 or 1, does it go vertically or horizontally first?

Modules/Common/include/Common/BigScreenCanvas.h
```diff
@@ -0,0 +1,64 @@
...
+  /// \brief set the text message and color of the box identified by "name"
+  void set(std::string name, int color, std::string text);
```
knopers8, line 40:
```suggestion
  void setText(std::string boxName, int color, std::string text);
```

Modules/Common/include/Common/BigScreenCanvas.h
```diff
@@ -0,0 +1,64 @@
...
+  /// \brief set the text message and color of the box identified by "name", based on the specified quality flag
+  void set(std::string name, o2::quality_control::core::Quality quality);
```
knopers8, line 42:
```suggestion
  void setQuality(std::string boxName, o2::quality_control::core::Quality quality);
```

Modules/Common/src/BigScreen.cxx
```diff
@@ -33,69 +31,79 @@ namespace o2::quality_control_modules::common
...
+  auto parOpt = customParameters.atOptional(parName, activity);
+  if (parOpt.has_value()) {
+    parValue = parOpt.value();
+  } else {
+    parOpt = customParameters.atOptional(parName);
+    if (parOpt.has_value()) {
+      parValue = parOpt.value();
+    }
+  }
+  return parValue;
```
knopers8, line 52: @Barthelemy it seems like a lot of code to use the CustomParameters. Could you have a look if it could not be done with less?

Modules/Common/include/Common/BigScreen.h
```diff
@@ -48,16 +44,14 @@ class BigScreen : public quality_control::postprocessing::PostProcessingInterfac
   void finalize(quality_control::postprocessing::Trigger, framework::ServiceRegistryRef) override;
 
  private:
+  /// \brief maximum allowed age of quality objects, in seconds (default: 10 minutes)
+  int mMaxObjectTimeShift{ 600 };
+  /// \brief read quality objects from all runs
+  int mIgnoreActivity{ 0 };
```
knopers8, line 50:
```suggestion
  bool mIgnoreActivity{ false };
```
Why not make it a bool for clarity?

Modules/Common/include/Common/BigScreenCanvas.h
knopers8: Could you also document that the color expected is de facto as defined with the constants in ROOT, such as `kYellow`?

Modules/Common/src/BigScreenConfig.cxx
```diff
@@ -15,83 +15,36 @@
...
       for (const auto& sourceName : sourceNames.value()) {
-        std::string sourceDet, sourcePath;
-        splitDataSourceName(sourceName.second.data(), sourceDet, sourcePath);
-        if (sourceDet.empty()) {
+        auto tokens = o2::utils::Str::tokenize(sourceName.second.data(), ':', false, true);
+        if (tokens.size() != 2 || tokens[0].empty()) {
           continue;
         }
-        dataSources.push_back({ sourceDet, sourcePath });
+        dataSources.push_back({ tokens[0], tokens[1] });
```
knopers8, line 34:
```suggestion
        const auto& sourceDet = tokens[0];
        const auto& sourcePath = tokens[1];
        dataSources.push_back({ sourceDet, sourcePath });
```
one could still keep clear what the tokens stand for

knopers8 (commented):
Thanks! Since one of the goals of the PR is improving the readability, I would have another few suggestions, but regarding the general code quality, I think it's OK.

Modules/Common/include/Common/BigScreenCanvas.h
aferrero2707: Done

Modules/Common/include/Common/BigScreenCanvas.h
aferrero2707: Done

Modules/Common/include/Common/BigScreenCanvas.h
aferrero2707: Done

Modules/Common/include/Common/BigScreen.h
aferrero2707: Good point, done.

Modules/Common/src/BigScreenConfig.cxx
aferrero2707: Again good point, done as well.

Modules/Common/src/BigScreen.cxx
aferrero2707: This is how I wrapped the CustomParameters calls in the MUON code: https://github.com/AliceO2Group/QualityControl/blob/ddb85738c291bac84c38eb7e396a974ea3b74ec4/Modules/MUON/Common/include/MUONCommon/Helpers.h#L36

aferrero2707:
@knopers8 thanks for the review! All comments included in the last commit, see my replies.

knopers8 (approved):
thank you!

PR #2193, opened 2024-03-20
BigScreen: changed message severity to Info

aferrero2707:
The message about missing QCDB objects does not need to be a Warning, since the corresponding box is painted in grey in this case.

The severity has been changed to `(Info, Support)` to avoid spamming the InfoLogger with unnecessary warnings.


aferrero2707:
@Barthelemy @knopers8 this is a simple code change to reduce the severity of one ILG message

Barthelemy (approved):

PR #2194, opened 2024-03-20
Uniform postprocessing task facility in IL

knopers8:
Before we suffered from having two different facility fields for logs from different parts of the code, this should fix it.


Barthelemy (approved):
thank you, it is true we were missing that.

PR #2201, opened 2024-03-21
TRD: merging TracketCheck and TrackletCountCheck

deependra170598:
Hi @martenole !
In this PR , I tried to merge two checks: TrackletCheck and TrackletCountCheck.
Please check if everything seems fine. If everything is fine I will remove commented statements.

One issue I see is that check class can read default parameters written in code
for example: 
```mStatThresholdPerTrigger = std::stof(mCustomParameters.atOrDefaultValue("StatThresholdPerTrigger", "1000.0", *mActivity));```

 but can not read default parameter provided with config 
for example:
Screenshot 2024-03-22 041100

### CODE REVIEW

File: Modules/TRD/src/TrackletCountCheck.cxx [hunk 1]
```diff
@@ -196,4 +212,91 @@ void TrackletCountCheck::beautify(std::shared_ptr<MonitorObject> mo, Quality che
...
+}
+
+void TrackletCountCheck::startOfActivity(const Activity& activity)
+{
+  ILOG(Debug, Devel) << "TrackletCountCheck::start : " << activity.mId << ENDM;
+  mActivity = make_shared<Activity>(activity);
+
+  ILOG(Debug, Devel) << "initializing TrackletCountCheck" << ENDM;
+
+  mThresholdMeanHighPerTimeFrame = std::stof(mCustomParameters.atOrDefaultValue("UpperthresholdPerTimeFrame", "600000.0" /*default value*/, *mActivity));
```
martenole, line 227: I stumbled upon this as well and don't really like the framework API. Basically you should not use the `atOrDefaultValue` method, as this will give you always the default value from the code and not from the json unless you have provided a value for the explicit activity `type` and `beamType`. Took me also a bit to figure out what to use. Please change all these to use `atOptional` like below:
```suggestion
  mThresholdMeanHighPerTimeFrame = std::stof(mCustomParameters.atOptional("UpperthresholdPerTimeFrame", *mActivity).value_or("6e5"));
```
I also think that with the scientific notation long numbers are more easy to read..

martenole (approved):
Did not go through everything in detail, but looks good to me

Barthelemy (approved):

PR #2203, opened 2024-03-22
ITS: memory improvements and new normalization

iravasen:
This PR reduces the memory footprint in async/sync operations:

- TrackTask:
  - From TH*D to TH*F 
  - Flag to ignore histogram with several bins (the biggest in the task which is not used in async)
- Cluster task:
  -  Improved flag to ignore 1296 (yes: 1296) histograms in async. They were not published but memory was allocated. They take less than a MB by making the calculation with the number of bins but I'm almost sure that when added into the framework they take more (still to evaluate). 
  - Remove three 3D C-arrays used as counters 

For cluster task I checked that the results produced are exactly the same as for the older cluster task. 

cc @JianLIUhep @IsakovAD


IsakovAD (approved):

IsakovAD:
All is good, but do you use mPublishMore?

iravasen:
thanks! Yes I used it here for async: https://github.com/AliceO2Group/O2DPG/pull/1549

iravasen:
@IsakovAD I fixed the bug with `getClusterCoordinates` in my latest commit. The issue was that we were not using that function in the right way. Now copying from the O2 I have implemented the correct solution which treats properly the clusterID 2047 which is exactly `o2::itsmft::CompCluster::InvalidPatternID`. Not sure why we never observed this crash with the full geometry but only with the light-weight one. The root cause is that with the previous code `getXCOG` with id 2047 was looking  a vector with size 2046.

iravasen:
ok I'm done

PR #2204, opened 2024-03-22
CTP: new check and trending histograms

lhusova:
add trending and check for input and class ratios + configurable parameters for check

### DIFF
--- a/Modules/CTP/src/CTPTrendingTask.cxx
+++ b/Modules/CTP/src/CTPTrendingTask.cxx
@@ -112,7 +112,6 @@ void CTPTrendingTask::trendValues(const Trigger& t, repository::DatabaseInterfac
       ILOG(Info, Support) << "no MO object" << ENDM;
       continue;
     }
-    ILOG(Info, Support) << "Got MO " << mo << ENDM;
     TObject* obj = mo ? mo->getObject() : nullptr;
     if (!obj) {
       ILOG(Info, Support) << "inputs not found" << ENDM;
@@ -139,7 +138,11 @@ void CTPTrendingTask::generatePlots()
     return;
   }
 
-  int index = 0;
+  int index = 0; // for keeping track what is trended:
+                 // 0 <= index < 5 - absolute input rates are trended
+                 // 5 <= index < 10 - absolute class rates are trended
+                 // 10 <= index < 15 - input rate ratios are trended
+                 // 15 <= index < 19 - class rate ratios are trended
 
   for (const auto& plot : mConfig.plots) {
 
@@ -150,8 +153,14 @@ void CTPTrendingTask::generatePlots()
       delete mPlots[plot.name];
     }
 
-    if (index > 4 && mClassIndex[index - 5] == 65) {
+    if (index > 4 && index < 10 && mClassIndex[index - 5] == 65) { // if the class index == 65, this class is not defined in the config, so it won't be trended
       ILOG(Info, Support) << "Class " << mClassNames[index - 5] << " is not trended." << ENDM;
+      index++;
+      continue;
+    }
+
+    if (index > 13 && (mClassIndex[index - 13] == 65 || mClassIndex[0] == 65)) { // if the class index == 65, this class is not defined in the config, so it won't be trended
+      ILOG(Info, Support) << "Class ratio " << mClassNames[index - 13] << " / " << mClassNames[0] << " is not trended." << ENDM;
       index++;
       continue;
     }
@@ -165,9 +174,16 @@ void CTPTrendingTask::generatePlots()
     if (auto histo = dynamic_cast<TH1*>(c->GetPrimitive("htemp"))) {
       if (index < 5)
         histo->SetTitle(mInputNames[index].c_str());
-      else
+      else if (index < 10)
         histo->SetTitle(mClassNames[index - 5].c_str());
-      histo->GetYaxis()->SetTitle("rate [Hz]");
+      else if (index < 14)
+        histo->SetTitle(Form("%s/%s", mInputNames[index - 9].c_str(), mInputNames[0].c_str()));
+      else
+        histo->SetTitle(Form("%s/%s", mClassNames[index - 13].c_str(), mClassNames[0].c_str()));
+      if (index < 10)
+        histo->GetYaxis()->SetTitle("rate [Hz]");
+      else
+        histo->GetYaxis()->SetTitle("rate ratio");
       c->Update();
 
       if (plot.varexp.find(":time") != std::string::npos) {
--- a/Modules/CTP/src/CTPTrendingTask.cxx
+++ b/Modules/CTP/src/CTPTrendingTask.cxx
@@ -172,18 +172,21 @@ void CTPTrendingTask::generatePlots()
     c->SetName(plot.name.c_str());
 
     if (auto histo = dynamic_cast<TH1*>(c->GetPrimitive("htemp"))) {
-      if (index < 5)
+      if (index < 5) {
         histo->SetTitle(mInputNames[index].c_str());
-      else if (index < 10)
+      } else if (index < 10) {
         histo->SetTitle(mClassNames[index - 5].c_str());
-      else if (index < 14)
+      } else if (index < 14) {
         histo->SetTitle(Form("%s/%s", mInputNames[index - 9].c_str(), mInputNames[0].c_str()));
-      else
+      } else {
         histo->SetTitle(Form("%s/%s", mClassNames[index - 13].c_str(), mClassNames[0].c_str()));
-      if (index < 10)
+      }
+
+      if (index < 10) {
         histo->GetYaxis()->SetTitle("rate [Hz]");
-      else
+      } else {
         histo->GetYaxis()->SetTitle("rate ratio");
+      }
       c->Update();
 
       if (plot.varexp.find(":time") != std::string::npos) {
--- a/Modules/CTP/src/RawDataReaderCheck.cxx
+++ b/Modules/CTP/src/RawDataReaderCheck.cxx
@@ -40,29 +40,53 @@ namespace o2::quality_control_modules::ctp
 
 void RawDataReaderCheck::configure()
 {
-
+  float casted;
   std::string param = mCustomParameters["thresholdRateBad"];
-  float casted = std::stof(param);
+  if (isdigit(param[0])) {
+    casted = std::stof(param);
+  } else {
+    casted = 0.15;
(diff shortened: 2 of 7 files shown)


Modules/CTP/src/CTPTrendingTask.cxx
```diff
@@ -165,9 +174,16 @@ void CTPTrendingTask::generatePlots()
     if (auto histo = dynamic_cast<TH1*>(c->GetPrimitive("htemp"))) {
       if (index < 5)
         histo->SetTitle(mInputNames[index].c_str());
-      else
+      else if (index < 10)
```
Barthelemy, line 177: The guidelines recommend using curvy braces whether there is a single line or not.

Modules/CTP/src/RawDataReaderCheck.cxx
```diff
@@ -37,7 +38,33 @@ using namespace o2::ctp;
 namespace o2::quality_control_modules::ctp
 {
 
-void RawDataReaderCheck::configure() {}
+void RawDataReaderCheck::configure()
+{
+
+  std::string param = mCustomParameters["thresholdRateBad"];
+  float casted = std::stof(param);
```
Barthelemy, line 45: what happens if `stof` fails ?

Modules/CTP/src/RawDataReaderCheck.cxx
```diff
@@ -131,6 +161,52 @@ Quality RawDataReaderCheck::check(std::map<std::string, std::shared_ptr<MonitorO
       else
         result = Quality::Good;
     }
+    if (mo->getName() == "inputRatio" || mo->getName() == "classRatio") {
```
Barthelemy, line 164: This block is a pretty difficult to understand. Could you perhaps add a few comments ? There are also quite many `if mo->getName() =="XXX"` that gets repeated. perhaps restructuring slightly would help.

Modules/CTP/src/RawDataReaderCheck.cxx
```diff
@@ -131,6 +161,52 @@ Quality RawDataReaderCheck::check(std::map<std::string, std::shared_ptr<MonitorO
...
+          inputRatioCycleCounter++;
+        }
+        if (mo->getName() == "classRatio") {
+          fHistClassRatioPrevious = (TH1F*)h->Clone();
+          classRatioCycleCounter++;
+        }
+      }
+      if (vIndexRateRatioBad.size() > 0)
+        result = Quality::Bad;
+      else if (vIndexRateRatioMedium.size() > 0)
```
Barthelemy, line 205: use curvy braces.

Barthelemy (changes requested):
Thank you, I have a few comments that I hope you can address. 
I did not comment each time the curly braces were missing, but please add them everywhere.

Modules/CTP/src/RawDataReaderCheck.cxx
lhusova: I added some comments and in the last commit I've rewritten the code. It is shorter now and hopefully also simpler to understand.

Modules/CTP/src/RawDataReaderCheck.cxx
```diff
@@ -37,7 +38,57 @@ using namespace o2::ctp;
 namespace o2::quality_control_modules::ctp
 {
 
-void RawDataReaderCheck::configure() {}
+void RawDataReaderCheck::configure()
+{
+  float casted;
+  std::string param = mCustomParameters["thresholdRateBad"];
+  if (isdigit(param[0])) {
```
knopers8, line 45: this can work only by some luck. you are checking if the first character in the string is a digit. Thus it will fail if it is e.g. "-1.0" or ".3" which are valid floating point numbers. Needless to say, it will also collapse for cases like "1xyz".

I think you will have to rely on handling exceptions thrown by `std::stof`, i don't know a more elegant way in C++ :/
I would propose to try this:
```
try {
  mThresholdRateBad = std::stof(mCustomParameters["thresholdRateBad"]);
} catch (const std::exception& e) {
  mThresholdRateBad = 0.15;
}
```

Modules/CTP/src/RawDataReaderCheck.cxx
```diff
@@ -81,55 +135,89 @@ Quality RawDataReaderCheck::check(std::map<std::string, std::shared_ptr<MonitorO
...
+      } else if (vMediumBC.size() > 0) {
         result = Quality::Medium;
-      else
+      } else {
         result = Quality::Good;
-    }
-    if (mo->getName() == "inputs" || mo->getName() == "classes") {
+      }
+    } else {
       auto* h = dynamic_cast<TH1F*>(mo->getObject());
```
knopers8, line 146: Are you certain that anything else you might get will be a `TH1F*` and can be treated with this part of the code?

Also, please check if `h` is not a nullptr (which could be when the cast has failed) before you access it, so your code does not crash in such a case, but produces a valid error message instead.

Modules/CTP/src/RawDataReaderCheck.cxx
```diff
@@ -81,55 +135,89 @@ Quality RawDataReaderCheck::check(std::map<std::string, std::shared_ptr<MonitorO
...
-      } else {
+      std::string moName = mo->getName();
+      if (moName.find("Ratio") != std::string::npos) {
+        relativeRates = true;
+      }
+      if (moName.find("input") != std::string::npos) {
+        inputRates = true;
+      }
+      if (relativeRates && inputRates) {
+        h->Scale(1. / h->GetBinContent(3));
```
knopers8: FYI, this modifies the histogram you are checking, it will be stored in QCDB like that. Perhaps you may want to move modification to `beautify()` for clarity, but if this is what you want, it's OK. Please keep in mind though, that if you ever apply more than one check on this object, the 2nd one will get the modified MO.

Modules/CTP/src/RawDataReaderCheck.cxx
```diff
@@ -81,55 +135,89 @@ Quality RawDataReaderCheck::check(std::map<std::string, std::shared_ptr<MonitorO
...
+      } else if (relativeRates) {
+        h->Scale(1. / h->GetBinContent(mIndexMBclass));
+      }
+
+      if (cycleCounter > 1) { // skipping the check for first two cycles, as the first one is arbitrary long
         TH1F* fHistDifference = (TH1F*)h->Clone();
```
knopers8, line 165: I don't see anywhere that you assign this pointer elsewhere or delete it. It looks like a memory leak.

Modules/CTP/src/RawDataReaderCheck.cxx
```diff
@@ -81,55 +135,89 @@ Quality RawDataReaderCheck::check(std::map<std::string, std::shared_ptr<MonitorO
...
+        float thrMedium = mThresholdRateMedium;
+        if (relativeRates) {
+          thrBad = mThresholdRateRatioBad;
+          thrMedium = mThresholdRateRatioMedium;
         }
 
-        if (mo->getName() == "inputs") {
-          fHistInputPrevious = (TH1F*)h->Clone();
-          inputCycleCounter++;
+        for (size_t i = 1; i < fHistDifference->GetXaxis()->GetNbins() + 1; i++) { // Check how many imputs/classes changed more than a theshold value
```
knopers8, line 193:
```suggestion
        for (size_t i = 1; i < fHistDifference->GetXaxis()->GetNbins() + 1; i++) { // Check how many inputs/classes changed more than a threshold value
```

Modules/CTP/src/RawDataReaderCheck.cxx
```diff
@@ -81,55 +135,89 @@ Quality RawDataReaderCheck::check(std::map<std::string, std::shared_ptr<MonitorO
...
+        fHistInputPrevious = (TH1F*)h->Clone();
+      }
+      if (inputRates && relativeRates) {
+        fHistInputRatioPrevious = (TH1F*)h->Clone();
+      }
+      if (!inputRates && !relativeRates) {
+        fHistClassesPrevious = (TH1F*)h->Clone();
+      }
+      if (!inputRates && relativeRates) {
+        fHistClassRatioPrevious = (TH1F*)h->Clone();
```
knopers8, line 219: for all of those, you are not deleting any object that could be at this pointer before the new `Clone()`. This looks like a memory leak.

Modules/CTP/src/RawDataReaderCheck.cxx
```diff
@@ -176,30 +265,34 @@ void RawDataReaderCheck::beautify(std::shared_ptr<MonitorObject> mo, Quality che
...
 
     h->GetYaxis()->SetRangeUser(0, h->GetMaximum() * 1.5);
-  }
-  if (mo->getName() == "inputs" || mo->getName() == "classes") {
+  } else {
     auto* h = dynamic_cast<TH1F*>(mo->getObject());
     h->SetStats(kFALSE);
     msg = std::make_shared<TLatex>(0.45, 0.8, Form("Overall Quality: %s", (checkResult.getName()).c_str()));
+    std::string groupName = "Input";
+    std::string relativness = "relative";
```
knopers8, line 277:
```suggestion
    std::string relativeness = "relative";
```

knopers8 (changes requested):
Hello @lucia , I found some more problems which could break your code, some memory leaks and a few minor corrections. Please have a look.

Modules/CTP/src/RawDataReaderCheck.cxx
lhusova: The case like  "1xyz", I tried and it does not fail. The casted value is then 1. But you are right, that I didn't think about cases like "-1.0" or ".3". In this case a negative value doesn't make sense, so my check would be ok. So I will do the exception check + logic check that the threshold value should be 0 < thr < 1.

Modules/CTP/src/RawDataReaderCheck.cxx
lhusova: yes, we are producing 5 histograms in the task. One needs to be treated separately and the other 4 can be handled with the code below "else"
The check for h being a null pointer will be added.

Modules/CTP/src/RawDataReaderCheck.cxx
lhusova: This is what we need to do. In this case, we want to check how the ratio between inputs and the specified minimum bias input changed after each cycle. As the bins represent different inputs, we need to scale the histogram first and then perform the check. There is only one check performed on the histogram.

knopers8 (approved):

PR #2206, opened 2024-03-22
GLO: improved mean vertex post-processing

aferrero2707:
Some improvements to the MeanVertexPostProcessing task:
- added title to graphs
- moved TrendGraph declaration into source file
- adapted code for START/STOP/START compatibility
- removed obsolete histograms with vertex position 
    - the `MeanVtxTrending` task will have to be removed as well from all configurations
- added checker for vertex trending plots

The PR is split in few separate commits to simplify the review.

### DIFF
--- a/Modules/GLO/CMakeLists.txt
+++ b/Modules/GLO/CMakeLists.txt
@@ -2,7 +2,7 @@
 
 add_library(O2QcGLO)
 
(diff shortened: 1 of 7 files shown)


aferrero2707:
@chiarazampolli @shahor02 @Barthelemy @knopers8 this is a simplified and cleaned-up version of the MeanVertex post-processing task. It reduces the complexity of the code, moves some declarations out of the header files, and removes the code required for the obsolete trending task. It also adapts the instantiation and publishing of the plots such that they are compatible with the START/STOP/START (to be double-checked).

In addition, the code introduces a checker for the trending graphs, with configurable ranges for each of the graphs.

Barthelemy (approved):

PR #2209, opened 2024-03-26
Small fix for 1_per_run to make sure we delete the objects created la…

Barthelemy:
…st when several objects share the same validity start. This was already at P2 but I forgot ot put commit it.


knopers8 (approved):
I guess you meant "make sure we *do not* delete the objects created last"

PR #2211, opened 2024-03-27
[EMCAL-1114] Basic QC task for pedestal calibration

mfasDa:
Monitoring of pedestal values per channel for FEC
and LEDMON channels both high gain and low gain.
Monitoring is done vs. module ID and module position.


mfasDa:
Needs https://github.com/AliceO2Group/AliceO2/pull/12882 and https://github.com/AliceO2Group/AliceO2/pull/12903 (interface change for LEDMON support)

Barthelemy (approved):

PR #2214, opened 2024-03-28
TOF: Allow to configure source of ccdb MOs

njacazio:


Modules/TOF/src/PostProcessDiagnosticPerCrate.cxx
```diff
@@ -30,14 +30,26 @@ using namespace o2::quality_control::postprocessing;
...
 const int PostProcessDiagnosticPerCrate::mNSlots = 14;
 
 PostProcessDiagnosticPerCrate::~PostProcessDiagnosticPerCrate()
 {
 }
 
+void PostProcessDiagnosticPerCrate::configure(const boost::property_tree::ptree& config)
+{
+  const std::string baseJsonPath = "qc.postprocessing." + getID() + ".customization.";
+  mCCDBPath = config.get<std::string>(baseJsonPath + "CCDBPath", "TOF/MO/TaskRaw/");
```
knopers8, line 43: Why use a dedicated key instead of `qc.config.conditionDB.url` ? Can yours be different than the main one?

knopers8: OK, I think I misunderstood what the variable stands for in your case...

knopers8 (approved):

PR #2216, opened 2024-04-01
ctpqc: opening CCDB in update

lietava:
In the first tests it appeared that when initialize is called CTPConfig is not yet in CCDB. Therefore initalisation was moved to first call of update.

### DIFF
--- a/Modules/CTP/src/CTPTrendingTask.cxx
+++ b/Modules/CTP/src/CTPTrendingTask.cxx
@@ -41,10 +41,8 @@ void CTPTrendingTask::configure(const boost::property_tree::ptree& config)
 {
   mConfig = TrendingConfigCTP(getID(), config);
 }
-
-void CTPTrendingTask::initialize(Trigger t, framework::ServiceRegistryRef services)
+void CTPTrendingTask::initCTP(Trigger& t)
 {
-  // // read out the CTPConfiguration
   std::string run = std::to_string(t.activity.mId);
   CTPRunManager::setCCDBHost("https://alice-ccdb.cern.ch");
   mCTPconfig = CTPRunManager::getConfigFromCCDB(t.timestamp, run);
@@ -77,14 +75,22 @@ void CTPTrendingTask::initialize(Trigger t, framework::ServiceRegistryRef servic
     reductor->SetTVXPHOIndex(mClassIndex[4]);
     mTrend->Branch(sourceName.c_str(), reductor->getBranchAddress(), reductor->getBranchLeafList());
   }
-
   getObjectsManager()->startPublishing(mTrend.get());
+}
+void CTPTrendingTask::initialize(Trigger t, framework::ServiceRegistryRef services)
+{
+  // // read out the CTPConfiguration
+  // initCCTP(); - too eraly here ?
 }
 
 void CTPTrendingTask::update(Trigger t, framework::ServiceRegistryRef services)
 {
   auto& qcdb = services.get<repository::DatabaseInterface>();
-
+  static bool init = 1;
+  if(init) {
+    initCTP(t);
+    init = 0;
+  }
   trendValues(t, qcdb);
   generatePlots();
 }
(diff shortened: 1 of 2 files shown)


lietava:
@aferrero2707 @knopers8 : please, let me know if you think it can worked like this.
@lhusova : FYI.

knopers8 (approved):
It may work, depending on your post-processing update trigger. If it is "newobject:ccdb:path/to/ctp/config" it should work. If you use some periodic  trigger, such as "5 min", then I guess it is subject to some race conditions.

lietava:
@knopers8 , it looks like it is 
"initTrigger": [
          "userorcontrol"
        ],
        "updateTrigger": [
          "newobject:qcdb:CTP/MO/RawData/inputs"
        ],
        "stopTrigger": [
          "userorcontrol"
        ]
so ?
The json is at:
https://ali-consul-ui.cern.ch/ui/alice-o2-cluster/kv/o2/components/qc/ANY/any/ctp-raw-qc_v1/edit

knopers8:
So you are still not directly relying on the presence of the expected object in the CCDB, but will try to access it at the first appearance of the QC objects you are trending.

This is better than before, but if the QC object appears before the one `CTPRunManager` needs, it will fail anyway. Why not attempt to `initCTP` during `update()` (as you do in the PR), but whenever it is not initialized instead of only at the 1st call? If it fails to initialize, you could skip the rest of the processing and try next time. This would seem more robust to me.

aferrero2707:
Not directly related, but I was looking at the way this CCDB object is retrieved and initialized. This is the relevant code from [here](https://github.com/AliceO2Group/AliceO2/blob/c328a12b8288f51917b85322e2cc5b7099c68706/DataFormats/Detectors/CTP/src/RunManager.cxx#L301):
```
CTPConfiguration CTPRunManager::getConfigFromCCDB(long timestamp, std::string run)
{
  auto& mgr = o2::ccdb::BasicCCDBManager::instance();
  mgr.setURL(mCCDBHost);
  map<string, string> metadata; // can be empty
  metadata["runNumber"] = run;
  auto ctpconfigdb = mgr.getSpecific<CTPConfiguration>(CCDBPathCTPConfig, timestamp, metadata);
  if (ctpconfigdb == nullptr) {
    LOG(info) << "CTP config not in database, timestamp:" << timestamp;
  } else {
    // ctpconfigdb->printStream(std::cout);
    LOG(info) << "CTP config found. Run:" << run;
  }
  return *ctpconfigdb;
}
```

If I understand correctly, if no object is found the code is effectively returning `*(nullptr)`. Is that ok?

knopers8:
This is surely not OK, thanks for spotting it!

lietava:
Why is it not ok ? I can check against nullptr ?

aferrero2707:
@lietava @knopers8 In my opinion, the best would be to modify `CTPRunManager::getConfigFromCCDB()` such that it returns a pointer to the object, instead of the object itself. This would allow the caller code (like the QC post-processing task we are discussing here) to check if the retrieval was successful or not.

lietava:
@aferrero2707 : it is used only in my macros. I will fixed it. I am surprised that compiler does not issue error/warning.
I bypassed it in this code and implemented what @knopers8 suggested.

lietava:
Please, can this be approved or let me know your comments.

knopers8 (approved):

Modules/CTP/src/CTPTrendingTask.cxx
```diff
@@ -41,17 +41,23 @@ void CTPTrendingTask::configure(const boost::property_tree::ptree& config)
...
+  metadata["runNumber"] = run;
+  mCTPconfig = mgr.getSpecific<CTPConfiguration>(CCDBPathCTPConfig, t.timestamp, metadata);
+  if (mCTPconfig == nullptr) {
+    return;
+  }
   // get the indeces of the classes we want to trend
-  std::vector<ctp::CTPClass> ctpcls = mCTPconfig.getCTPClasses();
-  std::vector<int> clslist = mCTPconfig.getTriggerClassList();
+  std::vector<ctp::CTPClass> ctpcls = mCTPconfig->getCTPClasses();
+  std::vector<int> clslist = mCTPconfig->getTriggerClassList();
```
aferrero2707, line 60: In the loop below, you use `clslist.size()` but then you access `ctpcls[i]`. Is `clslist` really needed? Wouldn't be better to use `ctpcls.size()` in the loop?

lietava: @aferrero2707 thanks, we shall check, cc @lhusova .

PR #2220, opened 2024-04-03
QC-1137 EMC BCVisualization survives nullptrs from retrieveMO()

knopers8:


mfasDa (approved):
Thanks for the proposal! We were also thinking about it. I approve it, however I have a bit stomach ache with this as it might cover a lower level issue (see my reply on the ticket).

knopers8:
Well, we still react to the issue in a sane way - we produce an error, we do not hide it.
But of course the root cause will have to be addressed, there is no question.

PR #2222, opened 2024-04-04
two tests translated from boost test to catch2

knopers8:


Framework/test/testCustomParameters.cxx
```diff
@@ -15,73 +15,69 @@
...
+  CHECK(cp.at("key", "default") == "value");
+  CHECK(cp.at("key", "default", "beam1") == "value_beam1");
 
   cp.set("key", "value_run1_beam1", "run1", "beam1");
-  BOOST_CHECK_EQUAL(cp.at("key"), "value");
-  BOOST_CHECK_EQUAL(cp.at("key", "run1", "beam1"), "value_run1_beam1");
+  CHECK(cp.at("key") == "value");
+  CHECK(cp.at("key", "run1", "beam1") == "value_run1_beam1");
 
   cout << cp << endl;
```
justonedev1, line 50: I know that this is not part of this PR, but is the `cout` necessary in tests?

justonedev1 (approved):

Framework/test/testCustomParameters.cxx
knopers8: I doubt, probably some leftover of Barth trying to see what's going on. I agree it could be removed.

knopers8: Well, perhaps one use is that we can check that operator<< does not throw, I will put it around NO_THROW statement.

PR #2227, opened 2024-04-06
GLO: Add ITS-TPC/ITS Check for QC

f3sch:
This PR adds a check for the ITS-TPC/ITS matching efficiency. 

Configurable values are:
 - minPt ---> minimum Pt for checked range
 - maxPt ---> maximum Pt for checked range
 - threshold ---> conservative Threshold value

Below are two example plots for very low MC statistics, where the only difference is the set threshold.
[good.pdf](https://github.com/AliceO2Group/QualityControl/files/14981460/good.pdf)
[bad.pdf](https://github.com/AliceO2Group/QualityControl/files/14981468/bad.pdf)
P.S. Also the line extending past the canvas is an artefact of low statistics shown here.

### DIFF
--- a/Modules/GLO/src/ITSTPCmatchingCheck.cxx
+++ b/Modules/GLO/src/ITSTPCmatchingCheck.cxx
@@ -188,7 +188,8 @@ Quality ITSTPCmatchingCheck::check(std::map<std::string, std::shared_ptr<Monitor
   auto isWorse = [](Quality const& a, Quality const& b) {
     return a.isWorseThan(b) ? a : b;
   };
-  auto findWorst = [&]<typename T, typename... Args>(const T& first, Args... args) {
+  auto findWorst = [&]<typename T, typename... Args>(const T& first, Args... args)
+  {
     return isWorse(first, isWorse(args...));
   };
   result.set(findWorst(ptQual, etaQual, phiQual));
--- a/Modules/GLO/src/ITSTPCmatchingCheck.cxx
+++ b/Modules/GLO/src/ITSTPCmatchingCheck.cxx
@@ -293,9 +293,6 @@ void ITSTPCmatchingCheck::beautify(std::shared_ptr<MonitorObject> mo, Quality ch
       msg->AddText("Not-handled Quality flag, don't panic...");
     }
     eff->GetListOfFunctions()->Add(msg);
-
-    TFile f("testPt.root", "RECREATE");
-    f.WriteObject(eff, "qcplotPt");
   }
 
   if (mShowPhi && mo->getName() == "mFractionITSTPCmatchPhi_ITS") {
@@ -376,9 +373,6 @@ void ITSTPCmatchingCheck::beautify(std::shared_ptr<MonitorObject> mo, Quality ch
       msg->AddText("Not-handled Quality flag, don't panic...");
     }
     eff->GetListOfFunctions()->Add(msg);
-
-    TFile f("testPhi.root", "RECREATE");
-    f.WriteObject(eff, "qcplotPhi");
   }
 
   if (mShowEta && mo->getName() == "mFractionITSTPCmatchEta_ITS") {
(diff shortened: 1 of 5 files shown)


f3sch:
@knopers8 thanks for the feedback!

knopers8:
> If you have nothing else to add I would ping RC so that they can also give feedback before merging? Cheers.

Yes, of course! @aferrero2707 @rmunzer ?

rmunzer:
> > If you have nothing else to add I would ping RC so that they can also give feedback before merging? Cheers.
> 
> Yes, of course! @aferrero2707 @rmunzer ?

HI @knopers8 @f3sch ,

thank you very much. This checker is indeed very useful.
May i ask if it is possible to use the checker also for eta and phi plots?

Thanks 
Robert

f3sch:
Hi @rmunzer, thanks for the feedback. I would suggest to merge this, while we take a look how to define the check for eta/phi. Cheers Felix

knopers8 (approved):

Modules/GLO/src/ITSTPCmatchingCheck.cxx
```diff
@@ -0,0 +1,250 @@
...
+
+using namespace std;
+using namespace o2::quality_control;
+
+namespace o2::quality_control_modules::glo
+{
+
+void ITSTPCmatchingCheck::configure()
+{
+  if (auto param = mCustomParameters.find("minPt"); param != mCustomParameters.end()) {
```
aferrero2707, line 45: I would suggest to move the parameters initialization in the `startOfActivity()` function, and use the extended parameters to allow defining beam-type dependent thresholds, if needed.

Modules/GLO/src/ITSTPCmatchingCheck.cxx
```diff
@@ -0,0 +1,250 @@
...
+    (void)moName;
+
+    if (mo->getName() == "mFractionITSTPCmatch_ITS") {
+      auto* eff = dynamic_cast<TEfficiency*>(mo->getObject());
+      auto binLow = eff->FindFixBin(mMinPt), binUp = eff->FindFixBin(mMaxPt);
+
+      result = Quality::Good;
+
+      if (eff->GetTotalHistogram()->GetEntries() == 0) {
+        result = Quality::Bad;
```
aferrero2707, line 88: Why not set the quality to `Null` if the denominator is empty? In such case the efficiency cannot be determined.

Modules/GLO/src/ITSTPCmatchingCheck.cxx
```diff
@@ -0,0 +1,250 @@
...
+        result.addReason(FlagReasonFactory::Unknown(), "No ITS tracks in denumerator!");
```
aferrero2707, line 89: `denumerator` -> `denominator`

Modules/GLO/src/ITSTPCmatchingCheck.cxx
```diff
@@ -0,0 +1,250 @@
...
+    } else {
+      msg->SetFillColor(kRed);
+      for (const auto& [_, desc] : checkResult.getReasons()) {
+        msg->AddText(Form("%s", desc.c_str()));
+      }
+      msg->SetTextColor(kWhite);
+    }
+    eff->GetListOfFunctions()->Add(msg);
+
+    if constexpr (false) {
```
aferrero2707, line 219: This should be probably removed in the final version

rmunzer:
> Hi @aferrero2707, @rmunzer,
> 
> I addressed your suggestion. There are now more options moved to extendedCheckParameters:
> 
> * showPt : default true
> * thresholdPt
> * min/maxPt
> * showPhi: default true     <----- might be useful to spot TPC sector/ITS stave issues
> * thresholdPhi
> * showEta: default false    <----- it is there, but we did not come up with what insight it can provide
> * thresholdEta
> * min/maxEta
> 
> I attach two example bad quality plots. [testPhi.pdf](https://github.com/AliceO2Group/QualityControl/files/15024915/testPhi.pdf) [testEta.pdf](https://github.com/AliceO2Group/QualityControl/files/15024916/testEta.pdf)
> 
> Any feedback is appreciated :)

Very nice. 
Thank you very much for adding those two check. 
In think all three will be very useful.

PR #2228, opened 2024-04-08
QC-1142 fix object preservation for 1_per_run policy

knopers8:
In some cases (e.g. a postprocessing task trending moving windows), the "validFrom" of the created objects might evolve slightly to the past. This is what we may see at the beginning of run:
```
I	11	09:25:34.936	check/sink-PTOF_TOFTrendingO1e7a	TOF	2mZYw7qfc8t	549286	Validity of MO 'TOFTrendingOrbits' is (1711956219957, 1711956273223) 
I	11	09:26:34.055	check/sink-PTOF_TOFTrendingO1e7a	TOF	2mZYw7qfc8t	549286	Validity of MO 'TOFTrendingOrbits' is (1711956219854, 1711956333793) 
I	11	09:27:34.090	check/sink-PTOF_TOFTrendingO1e7a	TOF	2mZYw7qfc8t	549286	Validity of MO 'TOFTrendingOrbits' is (1711956219854, 1711956392709)
```
On the other hand, the 1_per_run policy tries to sort the objects by "validFrom" and only if they are equal, it sorts them by "createdAt". Thus, in the discussed case it will preserve the object which is the latest in terms of "validForm", thus the object from the beginning of the run.

This commit fixes 1_per_run policy by relying solely on "createdAt" to find the latest object for a run. I could not think of any cases when this could not be applied.

### DIFF
--- a/Framework/script/RepoCleaner/qcrepocleaner/rules/1_per_run.py
+++ b/Framework/script/RepoCleaner/qcrepocleaner/rules/1_per_run.py
@@ -15,7 +15,7 @@ def process(ccdb: Ccdb, object_path: str, delay: int,  from_timestamp: int, to_t
     '''
     Process this deletion rule on the object. We use the CCDB passed by argument.
     Objects which have been created recently are spared (delay is expressed in minutes).
-    This specific policy, 1_per_run, keeps only the most recent version for a given run based on the validity_from.
+    This specific policy, 1_per_run, keeps only the most recent version for a given run based on the createdAt.
 
     Config Parameters:
     - period_pass: Keep 1 version for a combination of run+pass+period if true.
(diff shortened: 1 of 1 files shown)


knopers8:
I (dry-)ran this on production data, it seems to behave as expected for:
- data in the grace period (preserve everything)
- objects where validFrom is always the same (preserve only the last objects in the run)
- objects where validFrom moves back (preserve only the last object in the run, NOT THE FIRST)
- objects where validFrom continuously moves forward (e.g. moving windows, preserve the last)

Framework/script/RepoCleaner/qcrepocleaner/rules/1_per_run.py
```diff
@@ -68,16 +68,17 @@ def process(ccdb: Ccdb, object_path: str, delay: int,  from_timestamp: int, to_t
...
         freshest: ObjectVersion = None
         for v in sorted_run_versions:
-            if freshest is None or freshest.validFromAsDt < v.validFromAsDt:
-                if freshest is not None:
-                    if policies_utils.in_grace_period(freshest, delay):
-                        preservation_list.append(freshest)
-                    else:
-                        deletion_list.append(freshest)
+            if freshest is None:
+                freshest = v
```
justonedev1, line 76: you are changing logic of the original loop. `freshest` was assigned only when it was valid, now you assign it every time when it is not `None`... so basically do you initialize it as the `sorted_run_versions[0]`? Is it correct? and if so, shouldn't it be done outside of the loop?

justonedev1: oh no, sorry it is the same in the original loop, but weirdly written. So the `freshest` is the first item in `sorted_run_versions ` dataset everytime?

knopers8: basically yes. I rephrased it, thinking it's clearer to write it this way

knopers8: i just find it confusing to check if a variable is `None` in an `if` statement and then check if it is **not** `None` in another `if` statement inside.

justonedev1 (approved):

PR #2233, opened 2024-04-11
TOF: add start publishing

njacazio:


njacazio:
@ercolessi

Modules/TOF/src/PostProcessHitMap.cxx
```diff
@@ -126,7 +126,23 @@ void PostProcessHitMap::update(Trigger t, framework::ServiceRegistryRef services
     ILOG(Warning, Support) << "mHistoHitMap undefined, can't finalize" << ENDM;
     return;
   }
-  TCanvas* canvas = new TCanvas(mHistoHitMap->GetName(), mHistoHitMap->GetName());
+  if (!mCanvasMo) {
+    mCanvasMo = std::make_shared<TCanvas>(mHistoHitMap->GetName(), mHistoHitMap->GetName());
+    getObjectsManager()->startPublishing(mCanvasMo.get());
```
knopers8, line 131: In case of START STOP START, at the 2nd start the object will be still there in `mCanvasMo`, but it will not be registered to publish again.

You could consider deleting it in `finalize()` to be in a fresh state after you enter `update()` first time after the 2nd start.

njacazio: Hi Piotr, that's right, many thanks!

knopers8: this is in principle ok, but the small problem is that at 2nd START for a brief moment there will be two instances of the `TCanvas` with the same, here:
```
mCanvasMo = std::make_shared<TCanvas>("defaultMap", "defaultMap");
```
one on the left, one on the right. unfortunately this makes ROOT crash sometimes.
It will be all good if you reset the sh. pointer before:
```
mCanvasMo.reset();
```

the same issue might happen with `mPhosPad`

njacazio: Top, many thanks

knopers8 (approved):

PR #2234, opened 2024-04-12
CTF size monitoring for synch. reco

martenole:
- Adding `CTFSizeTask` which picks up the CTF size in bytes per detector from the CTF writer and creates two histograms per detector, one with constant bin size set differently for each detector and one with a logarithmic axis which has the same binning for all detectors
- Missing is a trending task which should plot the mean values of each histogram multiplied by the TF rate in seconds to get to the CTF size in Bytes/second

### DIFF
--- a/Modules/GLO/src/CTFSizeTask.cxx
+++ b/Modules/GLO/src/CTFSizeTask.cxx
@@ -119,10 +119,14 @@ void CTFSize::endOfActivity(const Activity& /*activity*/)
 void CTFSize::reset()
 {
   for (auto h : mHistSizes) {
-    h->Reset();
+    if (h) {
+      h->Reset();
+    }
   }
   for (auto h : mHistSizesLog) {
-    h->Reset();
+    if (h) {
+      h->Reset();
+    }
   }
 }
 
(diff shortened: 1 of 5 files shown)


martenole:
I have removed the custom trending task, since it is very likely that we can simply use the TrendingTask provided by the framework. This is ready to be merged from my side @knopers8

Modules/GLO/src/CTFSizeTask.cxx
```diff
@@ -0,0 +1,129 @@
...
+
+void CTFSize::reset()
+{
+  for (auto h : mHistSizes) {
+    h->Reset();
+  }
+  for (auto h : mHistSizesLog) {
+    h->Reset();
+  }
+}
```
knopers8, line 127: This might crash if we invoke `reset()` before `startOfActivity()`, I would add nullptr checks to be on the safe side.

This being said, in the current state the QC framework should not do that, so I will mark it as green and I would propose to merge it before the next release in any case.

knopers8 (approved):

Modules/GLO/src/CTFSizeTask.cxx
martenole: Thanks for the comment. I have added a protection. Yes, please merge it before the next release so that we can already get the individual size plots in the QCDB

PR #2235, opened 2024-04-15
QC-1136 Publication Policy for ObjectsManager

knopers8:
This lets user ask for how long the registered object should be published. By that, the user is made more aware of the expected lifetime of the object that they register, which proved to be difficult for START STOP START in postprocessing. SkeletonPostProcessing has been adapted to document the three main scenarios that may happen.

User tasks were adapted to use reasonable parameters. On the occasion, a few problems leading to potentially faulty START STOP START were additionally fixed.


Modules/Skeleton/src/SkeletonPostProcessing.cxx
```diff
@@ -20,31 +20,55 @@
...
-  mHistogram = new TH1F("example", "example", 20, 0, 30000);
-  getObjectsManager()->startPublishing(mHistogram);
+  // Resetting the histogram A, so it is empty at each new START
+  mHistogramA->Reset();
+
+  // We want the histogram B to be recreated at each START, so first we delete the existing one from the possible
+  // previous run, then create the new one. ROOT may crash if there are two objects with the same name at the same time.
+  mHistogramB.reset();
+  mHistogramB = std::make_unique<TH1F>("exampleB", "exampleB", 20, 0, 30000);
+  // We request that the histogram A is published after each update(), up to and including finalize()
```
Barthelemy, line 49:
```suggestion
  // We request that the histogram B is published after each update(), up to and including finalize()
```

Modules/Skeleton/src/SkeletonPostProcessing.cxx
```diff
@@ -20,31 +20,55 @@
...
+  // The framework will stop publishing it after finalize()
+  getObjectsManager()->startPublishing(mHistogramB.get(), PublicationPolicy::ThroughStop);
 }
 
 void SkeletonPostProcessing::update(Trigger t, framework::ServiceRegistryRef)
 {
   ILOG(Info, Support) << "Trigger type is: " << t.triggerType << ", the timestamp is " << t.timestamp << ENDM;
-  mHistogram->Fill(t.timestamp % 30000);
+
+  // Histogram C is recreated at each update and is expected to be published just one. We delete the previous
```
Barthelemy, line 58:
```suggestion
  // Histogram C is recreated at each update and is expected to be published just once. We delete the previous
```

Modules/TOF/src/PostProcessDiagnosticPerCrate.cxx
```diff
@@ -54,6 +54,7 @@ void PostProcessDiagnosticPerCrate::initialize(Trigger, framework::ServiceRegist
 {
   int counter = 0;
   for (auto& i : mCrates) {
+    i.reset();
```
Barthelemy, line 57: why do you call `reset()` first ?

Barthelemy (changes requested):
Very nice, thank you. 

I have fixed a couple of typo and have one question. 
could you also add a paragraph in the documentation ? 

Thank you

Modules/TOF/src/PostProcessDiagnosticPerCrate.cxx
knopers8: as explained in the SkeletonPostProcessing task:
>   // We want the histogram B to be recreated at each START, so first we delete the existing one from the possible
>  // previous run, then create the new one. ROOT may crash if there are two objects with the same name at the same time.

Barthelemy (approved):

PR #2236, opened 2024-04-15
[EMCAL-792] Catch std::out_of_range if no checkerParam is defined

mfasDa:


mfasDa (commented):
Hotfix, urgent, needs patch release tag.

knopers8:
The relevant tests already pass, including two different OSes, merging to release the patch.

PR #2238, opened 2024-04-16
[QC-1144] 1_per_run : Consider RunNumber=0 as no run number

Barthelemy:


Barthelemy:
I took the liberty to reformat the file as well

PR #2240, opened 2024-04-16
o2 [EMCAL-525] 3 QC checkers for EMCAL Physics 

anabananana98:


Modules/EMCAL/include/EMCAL/CellOccupancyCheck.h
```diff
@@ -0,0 +1,53 @@
...
+  void configure() override;
+  Quality check(std::map<std::string, std::shared_ptr<MonitorObject>>* moMap) override;
+  void beautify(std::shared_ptr<MonitorObject> mo, Quality checkResult = Quality::Null) override;
+  std::string getAcceptedType() override;
+
+ private:
+  /************************************************
+   * threshold cuts                               *
+   ************************************************/
+  int cThresholdMaxCounts = 10000; ///< Max number of counts in supermodule
```
mfasDa, line 46: Please follow the naming rules: Class member variables must start with lower case "m".

Modules/EMCAL/include/EMCAL/CellTimeCalibCheck.h
```diff
@@ -0,0 +1,54 @@
...
+  Quality check(std::map<std::string, std::shared_ptr<MonitorObject>>* moMap) override;
+  void beautify(std::shared_ptr<MonitorObject> mo, Quality checkResult = Quality::Null) override;
+  std::string getAcceptedType() override;
+
+ private:
+  /************************************************
+   * threshold cuts                               *
+   ************************************************/
+  double cSigmaTSpectrum = 0.1;   ///< TSpectrum parameter sigma
+  double cThreshTSpectrum = 0.01; ///< TSpectrum parameter threshold
```
mfasDa, line 47: Again naming rules

Modules/EMCAL/src/CellAmpCheck.cxx
```diff
@@ -0,0 +1,85 @@
...
+{
+Quality CellAmpCheck::check(std::map<std::string, std::shared_ptr<MonitorObject>>* moMap)
+{
+    if(!mCellAmpReference) {
+        //Load reference histogram from the CCDB
+        mCellAmpReference = this->retrieveConditionAny<TH1>("/EMC/QCREF/CellAmpCalibCheck");
+    }
+    if(!mCellAmpReference) {
+        //If histogram not found print out error message
+        std::cout << "Could not find the reference histogram!" << std::endl;
```
mfasDa, line 34: Please use the infoLogger facility for logging instead
```c++
ILOG(Error, Support) << ... << ENDM;
```

Modules/EMCAL/src/CellOccupancyCheck.cxx
```diff
@@ -0,0 +1,251 @@
...
+    }
+  }
+}
+
+Quality CellOccupancyCheck::check(std::map<std::string, std::shared_ptr<MonitorObject>>* moMap)
+{
+  auto mo = moMap->begin()->second;
+  Quality result = Quality::Good;
+  if (mo->getName() == "cellOccupancyEMCwThr_PHYS") {
+    auto* h = dynamic_cast<TH2*>(mo->getObject());
```
mfasDa, line 41: You might have several objects being checked by the checker at the same time. This will only process the first object. Instead, please move common parts out (and if they don't change make them const), eventually to a different function, and iterate in a loop over the objects.

Modules/EMCAL/src/CellOccupancyCheck.cxx
```diff
@@ -0,0 +1,251 @@
...
+
+    // Define bin edge array of the form (BinEdgeX[n],BinEdgeX[n+1]) = (low bin edge of nth SM, up bin edge of nth SM)
+    int BinEdgesX[40];
+    int BinEdgesY[40];
```
mfasDa, line 45: Can be made const as they content is const.

Modules/EMCAL/src/CellOccupancyCheck.cxx
```diff
@@ -0,0 +1,251 @@
...
+          SMcounts[17] += counts;
+        }
+        if (BinEdgesX[36] <= bin_low_x && bin_up_x <= BinEdgesX[37] && BinEdgesY[36] <= bin_low_y && bin_up_y <= BinEdgesY[37]) {
+          SMcounts[18] += counts;
+        }
+        if (BinEdgesX[38] <= bin_low_x && bin_up_x <= BinEdgesX[39] && BinEdgesY[38] <= bin_low_y && bin_up_y <= BinEdgesY[39]) {
+          SMcounts[19] += counts;
+        }
+      }
+    }
```
mfasDa, line 208: This is a lot of checking. I wonder whether it wouldn't make more sense to iterate in the outer loop over the supermodule, and in the two inner ones about column and row, where with column and row you access the bin content.

PR #2242, opened 2024-04-17
TRD fix data matcher for CCDB object of sampled input

martenole:
It is a bit stupid, but I don't have a better idea at the moment. In case we are using direct data sampling the data spec has the expected name `TRD/FCHSTATUS/0`. But when we use data sampling (as it is done online) then the data spec receives a new name `DS` for data sampling, `trdall` for the sampling policy name and a number based on the position in the query string. The problem is of course that like this the chamber status always needs to be in the correct position of the query string, otherwise the task will segfault.

### DIFF
--- a/Modules/TRD/src/DigitsTask.cxx
+++ b/Modules/TRD/src/DigitsTask.cxx
@@ -335,7 +335,7 @@ void DigitsTask::endOfActivity(const Activity& /*activity*/)
 
 void DigitsTask::finaliseCCDB(o2::framework::ConcreteDataMatcher& matcher, void* obj)
 {
-  if (matcher == o2::framework::ConcreteDataMatcher("TRD", "FCHSTATUS", 0)) {
+  if (matcher == o2::framework::ConcreteDataMatcher("TRD", "FCHSTATUS", 0) || matcher == o2::framework::ConcreteDataMatcher("DS", "trdall4", 0)) {
     // LB: no half chamber distribution map for Digits, pass it as null pointer
     TRDHelpers::drawChamberStatusOnHistograms(static_cast<std::array<int, MAXCHAMBER>*>(obj), nullptr, mLayers, NCOLUMN);
   }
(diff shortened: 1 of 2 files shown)


knopers8 (approved):
Hi @martenole , this hack is fine, it should work, but FYI, since this PR you can also put the CCDB stuff as a direct source in combination with a data sampling policy: https://github.com/AliceO2Group/QualityControl/pull/2098

martenole:
Great, I did not no that one could specify multiple data sources. This is of course exactly what I needed. I tested it on staging and it is working, so I will close this one.
But can the multiple data sources have an effect on the cycle duration? I have configured 180 seconds, but the first objects of the tasks were published only after 7 minutes with one direct and one sampling source. When I use a single source only I see the expected 3 minutes delay

martenole:
Sure, the env with two separate sources was [2mt4uEy7zmS](http://alio2-cr1-hv-mvs00.cern.ch:4000/?page=env-details&environmentId=2mt4uEy7zmS) and the one with all in one data sampling policy was [2mt5dXXqtuJ](http://alio2-cr1-hv-mvs00.cern.ch:4000/?page=env-details&environmentId=2mt5dXXqtuJ)

knopers8:
Hm, it seems to me it was completely unrelated and expected. Your task has:
```
      "Digits": {
...
        "cycleDurationSeconds": "180",
...
        "mergerCycleMultiplier": "2",
...
        "mergersPerLayer": [
          "3",
          "1"
        ]
      },
```
thus the objects go through 3 cycles until they are stored. The first cycle is always randomly shorter than the full cycle, which will cause different latency at each run. This means [0, 180s] for QC tasks for the first cycle, then 180s for the later cycles. For Mergers it is [0, 360s], then 360s, but you get two layers. So the time you get your first is object is anything between 0 and `180s + 360s + 360s`, depending on the length of the first cycles and the order that they and among the three steps.

Yes, I would like to improve that ;)

martenole:
Ah OK, so my mistake was that I did not take into account the merger(s) cycle durations. And then I also did not know that the first cycle can be randomly shorter.
Sorry for the false alarm. I could have taken more runs of course yesterday evening and then would have seen the random delays, but I was not so motivated anymore :)

PR #2245, opened 2024-04-18
TPC Update CheckOfSlices with Masking

mlesch:


Modules/TPC/src/CheckOfSlices.cxx
```diff
@@ -98,6 +99,15 @@ void CheckOfSlices::configure()
...
   }
+
+  std::string maskingValues = common::getFromConfig<string>(mCustomParameters, "maskedPoints", "");
+  if (maskingValues != "") {
+    std::stringstream ss(maskingValues);
+    std::string point;
+    while (std::getline(ss, point, ',')) {
+      mMaskedPoints.push_back(std::stoi(point));
+    }
+  }
```
wiechula, line 110: You could
```
#include "Algorithm/RangeTokenizer.h"
```
and use
```
mMaskedPoints = RangeTokenizer::tokenize<int>(maskingValues);
```

wiechula (changes requested):
Just one minor suggestion.

Modules/TPC/src/CheckOfSlices.cxx
mlesch: Thanks for the nice suggestion! I implemented it

wiechula (approved):

PR #2254, opened 2024-04-22
[EMCAL-527] Added QC checker for Payload Size check

anabananana98:


mfasDa (changes requested):
Also here a few changes are needed. Furthermore. I am curious what range of chi2 we expect. This is crucial in order to tune the checker. We probably need some to propagate this information somewhere, i.e. in the quality itself.

Modules/EMCAL/src/PayloadPerEventDDLCheck.cxx
```diff
@@ -0,0 +1,109 @@
...
+
+namespace o2::quality_control_modules::emcal
+{
+// configure threshold-based checkers
+auto nChiSqMedPayloadThresh = mCustomParameters.find("ChiSqMedPayloadThresh");
+if (nChiSqMedPayloadThresh != mCustomParameters.end()) {
+  try {
+    mChiSqMedPayloadThresh = std::stod(nChiSqMedPayloadThresh->second);
+  } catch (std::exception& e) {
+    ILOG(Error, Support) << ("Value {} not a double", nChiSqMedPayloadThresh->second.data()) << ENDM;
```
mfasDa, line 29: Doesn't work, must be
``` c++
ILOG(Error, Support) << "Value " <<  nChiSqMedPayloadThresh->second << " not a double" << ENDM;
```
Did you try to build QualityControl?

In addition, when using c++ streams you can use strings directly, as well as other objects that implement operator<<, and don't need to access the C-string.

Modules/EMCAL/src/PayloadPerEventDDLCheck.cxx
```diff
@@ -0,0 +1,109 @@
...
+  }
+}
+
+auto nBadChiSqLowPayloadThresh = mCustomParameters.find("BadChiSqLowPayloadThresh");
+if (nBadChiSqLowPayloadThresh != mCustomParameters.end()) {
+  try {
+    mChiSqBadLowPayloadThresh = std::stod(nBadChiSqLowPayloadThresh->second);
+  } catch (std::exception& e) {
+    ILOG(Error, Support) << ("Value {} not a double", nBadChiSqLowPayloadThresh->second.data()) << ENDM;
```
mfasDa, line 38: Same as above

Modules/EMCAL/src/PayloadPerEventDDLCheck.cxx
```diff
@@ -0,0 +1,109 @@
...
+
+  auto mo = moMap->begin()->second;
+  Quality result = Quality::Good;
+
+  if (mo->getName() == "PayloadSizePerDDL") {
+
+    auto* h = dynamic_cast<TH2*>(mo->getObject());
+
+    for (int i = 1; i <= h->GetNbinsX(); i++) {
+      std::unique_ptr<TH1> h_y = (TH1*)h->ProjectionY(Form("h_y_%d", i), i, i);
```
mfasDa, line 58: Also this is not valid code, the operator= to pointers is not supported. The line must be changed to 
```c++
std::unique_ptr<TH1> h_y(h->ProjectionY(Form("h_y_%d", i), i, i));
```
To initialise a smart pointer this has to be done in the constructor or by setter functions, otherwise it cannot do the memory management.

Modules/EMCAL/src/PayloadPerEventDDLCheck.cxx
```diff
@@ -0,0 +1,109 @@
...
+      h_y->Scale(1. / h_y->Integral());
+      std::unique_ptr<TH1> h_y_ref = (TH1*)mCalibReference->ProjectionY(Form("h_y_ref_%d", i), i, i);
```
mfasDa, line 60: Same as above.

Modules/EMCAL/src/PayloadPerEventDDLCheck.cxx
```diff
@@ -0,0 +1,109 @@
...
+{
+  if (mo->getName() == "PayloadSizePerDDL") {
+    auto* h = dynamic_cast<TH1*>(mo->getObject());
+    TPaveText* msg = new TPaveText(0.5, 0.5, 0.9, 0.75, "NDC");
+    h->GetListOfFunctions()->Add(msg);
+    msg->SetName(Form("%s_msg", mo->GetName()));
+    if (checkResult == Quality::Good) {
+      //
+      msg->Clear();
+      msg->AddText("Agrees with reference histogram: OK!!!");
```
mfasDa, line 89: I would say something like "Data OK" in the message.

Modules/EMCAL/src/PayloadPerEventDDLCheck.cxx
```diff
@@ -0,0 +1,109 @@
...
+      msg->SetFillColor(kGreen);
+      //
+      h->SetFillColor(kGreen);
+    } else if (checkResult == Quality::Bad) {
+      ILOG(Debug, Devel) << "Quality::Bad, setting to red";
+      msg->Clear();
+      msg->AddText("ERROR: Disagrees with reference histogram, high Chi-Sq");
+      msg->AddText("If NOT a technical run,");
+      msg->AddText("call EMCAL on-call.");
```
mfasDa, line 98: I would say something like

"Suspicious payload size detected. If not in a technical run, call EMCAL on-call"

Modules/EMCAL/src/PayloadPerEventDDLCheck.cxx
```diff
@@ -0,0 +1,109 @@
...
+      h->SetFillColor(kRed);
+    } else if (checkResult == Quality::Medium) {
+      msg->Clear();
+      msg->AddText("WARNING: Chi-Sq greater than medium threshold");
```
mfasDa, line 102: I would say something like "Warning: Deviations from expectations. Keep monitoring"

mfasDa (changes requested):
Still not good. Few lines should raise compiler errors. Did you build and test the code locally?

Modules/EMCAL/src/PayloadPerEventDDLCheck.cxx
```diff
@@ -0,0 +1,112 @@
...
+    msg->SetName(Form("%s_msg", mo->GetName()));
+    if (checkResult == Quality::Good) {
+      //
+      msg->Clear();
+      msg->AddText("DATA OK!");
+      msg->SetFillColor(kGreen);
+      //
+      h->SetFillColor(kGreen);
+    } else if (checkResult == Quality::Bad) {
+      ILOG(Debug, Devel) << "Quality::Bad, setting to red";
```
jokonig, line 97: I think here we should also have an "<< ENDM;"
and the same for the one below

mfasDa (approved):
I think also this we can merge now

PR #2255, opened 2024-04-23
[O2B-1010] Hide protobuf internals

martinboulais:
Adapt to bookkeeping library hiding the internal dependency on protobuf
Depends on https://github.com/AliceO2Group/Bookkeeping/pull/1530


Barthelemy (approved):
I approve but this will be merged only when the new BK has been released and updated in alidist

PR #2258, opened 2024-04-23
[EMCAL-529] Added Checker for Number of Patches per FastOR Trigger histo

sweyh99:
Check for noisy FastORs and message in the quality object and infologger which TRUs are noisy


mfasDa (approved):
For the moment it can be merged for testing. The method is probably overcomplicated for the problem, I was thinking about something much simpler, however we need to see how it performs.

PR #2261, opened 2024-04-24
TOF: Use DPL CCDB fetcher rather than CCDBManager

ktf:
TOF: Use DPL CCDB fetcher rather than CCDBManager

### DIFF
--- a/Modules/TOF/include/TOF/TaskDigits.h
+++ b/Modules/TOF/include/TOF/TaskDigits.h
@@ -102,8 +102,6 @@ class TaskDigits final : public TaskInterface
 
   int mNoiseClassSelection = -1; /// Index to discard classes of noisy channels -1 no discarding, 0 first class are discarded, 1 second class, 2 third class
 
-  o2::dataformats::CalibTimeSlewingParamTOF* mCalChannel = nullptr;
-  o2::dataformats::CalibLHCphaseTOF* mLHCphase = nullptr;
   bool mApplyCalib = false;
 
   ////////////////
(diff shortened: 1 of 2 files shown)


ktf:
Notice you probably need to fix the json configuration as well for this to work correctly.

chiarazampolli:
Hello, 

Indeed, if this is not in sync with changes in the json, it will not work.

Chiara

chiarazampolli:
It should be the json in QC (if it exists), O2DPG for sync, async, MC, and then in consul.
One should add something like:

```
alpparITS:ITS/ALPIDEPARAM/0?lifetime=condition&ccdb-path=ITS/Config/AlpideParam
```
to the query, (example from https://github.com/AliceO2Group/O2DPG/blob/master/DATA/production/qc-async/itstpc.json, https://github.com/AliceO2Group/O2DPG/blob/7188d085958ba23d5a7cb647995a962d6d2d0090/DATA/production/qc-async/itstpc.json#L42C254-L42C333) with the correct CCDB entry.

ktf:
Yes, the correct string is: `channelCalib:TOF/CHCALIB/0?lifetime=condition&ccdb-path=TOF/Calib/ChannelCalib;lhcPhase:TOF/LHCP/0?lifetime=condition&ccdb-path=TOF/Calib/LHCphase`

ktf:
See also https://its.cern.ch/jira/browse/QC-1166

chiarazampolli:
Hello, 
This PR breaks async reco, probably the same is true for sync. We need to change the json files...
Chiara

Barthelemy:
For sync I had notified Pippo to make sure the json files are updated when we upgrade at P2

pippohub:
Ciao @ktf @chiarazampolli this commit  is breaking the run, TOF generate segfault ... we are considering to revert the commit to unlock the release ... would it be ok?

ktf:
I have personally have no objections, I just wanted to show off the point I was making in https://its.cern.ch/jira/browse/QC-1166. I did say it needed an updated configuration, but I have no idea which one actually applies for sync reco. In my local tests with the updated configuration as described above it was working fine, AFAICT.

PR #2266, opened 2024-04-28
[EMCAL-525] Added 3 checkers for EMCAL PHYS

anabananana98:
Number of physics triggers
Cell Occupancy check
Cell Time Calibration check


Modules/EMCAL/src/CellOccupancyCheck.cxx
```diff
@@ -0,0 +1,118 @@
...
+        1, 32, 65, 96, 1, 48, 49, 96
+      };
+
+      const int BinEdgesY[40] = {
+        1, 25, 1, 25, 26, 49, 26, 49,
+        50, 73, 50, 73, 74, 97, 74, 97,
+        98, 121, 98, 121, 122, 129, 122, 129,
+        130, 153, 130, 153, 154, 177, 154, 177,
+        178, 201, 178, 201, 201, 208, 201, 208
+      };
```
mfasDa, line 58: Constants, move before for-loop. Also prefer std::array with respect to traditional c-arrays

Modules/EMCAL/src/CellOccupancyCheck.cxx
```diff
@@ -0,0 +1,118 @@
...
+            }
+          }
+        }
+      }
+
+      // Check if the counts in any supermodule exceed the threshold
+      for (int k = 0; k < 20; k++) {
+        if (SMcounts[k] > mThresholdMaxCounts) {
+          result = Quality::Bad;
+        }
```
mfasDa, line 80: The counts depend on the statistics, there is in principle no problem if the count rate is above threshold, it will at some point be always true. 

What we are worried about is whether we suddenly get cold regions in the supermodule. Therefore I think we need to change the entire algorithm:
1) We need to split supermodules in rectangles, where I would do 4x8 tower squares (FECs)
2) We take the truncated mean of the FEC integrated count rates, excluding dead FECs (0 counts)
3) We count the number of FECs / supermodule which have a count rate < fraction of mean, excluding dead FECs
4) We check if the number of cold FECs / supermodule > threshold per supermodule where we make the threshold adjustable per supermodule.

mfasDa (changes requested):
Sorry to request changes again, but we need to work on the algorithm of the occupancy checker. It is supposed to find cold regions in the occupancy, not counts > threshold.

mfasDa (approved):
Approving the first iteration, however we need the beautify function as well.

knopers8:
Failing CI checks seem to be related.

PR #2270, opened 2024-04-30
MCH: fix crash when empty ROFs are encountered

aferrero2707:
The commit adds a protection when some of the processed ROFs are empty or has invalid digit indexes.
This is a pathological condition that should actually never occur.
In this case the ROF is skipped and a warning message is sent to the Infologger.


chiarazampolli (approved):

Barthelemy:
HI Chiara, 
Sorry, did not see your message.

Barthelemy (approved):

PR #2274, opened 2024-05-02
MFT: Fix for timestamp determination for noise scan

Edingrast:
*Addition of new timestamp parameter in the readNoiseMap() function.

*Adding timestamps given by validity of object or current time to correctly obtain noise maps for comparison inside task.


diana0x0f (approved):
thanks @Edingrast for preparing the fix!

PR #2275, opened 2024-05-03
Revert "[QC-1079] Run type as a string"

Barthelemy:
Reverts AliceO2Group/QualityControl#2077


Barthelemy:
We have to revert because we don't get the proper run type on the EPNs.

PR #2278, opened 2024-05-05
MCH: add per-station distributions

aferrero2707:
The code adds some 1-D histograms that provide distributions of relevant quantities (hit rates, efficiencies, cluster sizes and charge) separately for each tracking station.
The plots are useful for diagnostics and comparison with reference data.

It also introduces a utility function to create a log10-spaced binning that is suitable to initialize variable-bin-sized 1-D histograms for which the X-axis should be displayed in log scale.

### DIFF
--- a/Modules/MUON/Common/include/MUONCommon/Helpers.h
+++ b/Modules/MUON/Common/include/MUONCommon/Helpers.h
@@ -45,6 +45,10 @@ T getConfigurationParameter(o2::quality_control::core::CustomParameters customPa
   return getConfigurationParameter<T>(customParameters, parName, defaultValue);
 }
 
+// create an array of bins in a given range with log10 spacing
+std::vector<double> makeLogBinning(double xmin, double xmax, int nbins);
+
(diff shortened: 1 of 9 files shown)


Barthelemy:
Hi Andrea, 
How many plots are you adding ? 
thank you

aferrero2707:
In the current version we have the following additional plots:
* ClustersTask => 5 1-D plots
* DigitsTask => 20 1-D plots
* PreclustersTask => 15+15+10 = 40 1-D plots

I am preparing an update that reduces the number:
* ClustersTask => 5 1-D plots
* DigitsTask => 2 2-D plots
* PreclustersTask => 15+15 = 30 1-D plots

The last one could be eventually made optional if the number of new objects is too large, but those are also the most interesting to have (clusters size and charge, separately for each station and cathode combination)

Barthelemy:
Thanks, that's good !

Barthelemy (approved):

PR #2280, opened 2024-05-06
[EMCAL-711] Add monitoring of TRU decoding errors

mfasDa:
- Integrate TRU decoding errors into global error
  monitoring
- Add histograms for TRU error type vs. Link and index
  of the TRU raising the error (STU index scheme) vs.
  link


mfasDa:
Needs https://github.com/AliceO2Group/AliceO2/pull/13112

PR #2281, opened 2024-05-07
Cleanup CellTask

ktf:
Cleanup CellTask

### DIFF
--- a/Modules/EMCAL/include/EMCAL/CellTask.h
+++ b/Modules/EMCAL/include/EMCAL/CellTask.h
@@ -46,9 +46,7 @@ class GainCalibrationFactors;
 class Cell;
 } // namespace emcal
 
-namespace quality_control_modules
-{
-namespace emcal
+namespace quality_control_modules::emcal
 {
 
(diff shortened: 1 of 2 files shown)


mfasDa:
@ktf I guess this relates to c++20, right?

ktf:
What exactly? The main issue is that there are two arrays not initialised (mCellTimeSupermoduleEMCAL_Gain and mCellTimeSupermoduleDCAL_Gain).

Besides that, `delete nullptr` is actually guaranteed to be a no-op (since ever) and it does not reset the pointer to be null. From https://en.cppreference.com/w/cpp/memory/new/operator_delete:

> In all cases, if ptr is a null pointer, the standard library deallocation functions do nothing. If the pointer passed to the standard library deallocation function was not obtained from the corresponding standard library allocation function, the behavior is undefined.
After the standard library deallocation function returns, all pointers referring to any part of the deallocated storage become invalid.
Indirection through a pointer that became invalid in this manner and passing it to a deallocation function (double-delete) is undefined behavior. Any other use is implementation-defined.

PR #2283, opened 2024-05-07
Do not const_cast/modify sh.mem. objetcs from upstream procsesses

shahor02:
The tracks produced in the tracks and supposedly stored in the AOD as is were modified but the QC tasks after const_casting objects in the shared memory. It looks like only standalone TPC tracks were affected (were propagated to the DCA) so the damage was limited.


Barthelemy:
we have a release tomorrow. Is it urgent or do we wait for the approval of the detectors ?

shahor02:
Hi @Barthelemy , I've tested it locally. I would not wait for validation by detectors since this is a fix to respect basic aspect of O2 processing: the DPL input must treated as const.

Barthelemy (approved):
ok, then I approve and I will merge

PR #2287, opened 2024-05-14
[QC-1117] Handle case when no cycle duration is set

Barthelemy:


knopers8 (approved):
Yes, agreed!
However, I don't like the fact that we perform validation of this parameter here, but also in https://github.com/AliceO2Group/QualityControl/blob/0b697ea1920c2b317d5f89d10de6df6843be5242/Framework/src/TaskRunnerFactory.cxx#L183

Should we refactor this perhaps?

Barthelemy:
yes, indeed, I'll do that

knopers8:
Thanks!

I would be even better if the cycle durations were also sanitized for Mergers, since now they are not: https://github.com/AliceO2Group/QualityControl/blob/0b697ea1920c2b317d5f89d10de6df6843be5242/Framework/src/InfrastructureGenerator.cxx#L142 If you run a remote QC workflow, we will not enter the sanitization code path (it is done only when creating TaskRunners), thus we will not bail out. I let you decide whether it should be done in the scope of this ticket, but surely we should fix this.

Barthelemy:
Yes, I have called the sanitize method for the mergers as well.

PR #2290, opened 2024-05-16
[EMCAL-526, EMCAL-527, EMCAL-832] Update pedestal check, move per-SM histos to 2D histos

mfasDa:
- 2D histograms for min. bunch amp vs SM and FEC
- Support for 2D min. bunch amp histograms in raw
  check
- Settable cuts for min. bunch amp check

### DIFF
--- a/Modules/EMCAL/include/EMCAL/RawCheck.h
+++ b/Modules/EMCAL/include/EMCAL/RawCheck.h
@@ -18,8 +18,12 @@
 #define QC_MODULE_EMCAL_EMCALRAWCHECK_H
 
 #include "QualityControl/CheckInterface.h"
+#include <array>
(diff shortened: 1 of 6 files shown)


mfasDa:
Ready for merging

mfasDa:
@Barthelemy @knopers8 This PR moves all the histograms per-SM to 2D histograms, which removes 200 objects from the task and reduces the number of objects published by the RawTask to 38. In addition we will publish 10 objects form the SupermoduleProjectionTask publishing the per-supermodule projections as single canvas. Ideally would be good to optimise also on the trending plots, which we also need per-SM to publish them as a single canvas or even single pad.

jokonig (approved):

PR #2291, opened 2024-05-17
Adding K0s Qc to GLO matching QC

chiarazampolli:
This PR adds a first inv mass plot from K0s to the GLO matching QC (added here to not add another QC). The use of this is regulated by a flag.
@mfaggin , @iouribelikov , @f3sch , @alcaliva , @noferini , @shahor02
See also: https://its.cern.ch/jira/browse/QC-1174 and related PR in O2: https://github.com/AliceO2Group/AliceO2/pull/13144

### DIFF
--- a/Modules/GLO/src/ITSTPCMatchingTask.cxx
+++ b/Modules/GLO/src/ITSTPCMatchingTask.cxx
@@ -24,7 +24,7 @@
 #include <Framework/InputRecordWalker.h>
 #include <QualityControl/stringUtils.h>
 
-using matchType = o2::globaltracking::MatchITSTPCQC::matchType;
+using matchType = o2::gloqc::MatchITSTPCQC::matchType;
 
 namespace o2::quality_control_modules::glo
 {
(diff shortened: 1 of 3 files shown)


knopers8:
It seems the failing CI checks are related:
```
--
CMake Error at Modules/GLO/CMakeLists.txt:14 (target_link_libraries):
  Target "O2QcGLO" links to:

    O2::GLOQC

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.
 ```

**Comment by chiarazampolli** [comment]
The O2 PR is needed... 
I am now trying to make first O2 compatible with existing QC, see https://github.com/AliceO2Group/AliceO2/pull/13144. It might take a few days, due to meetings etc. Apologies. I can move this to WIP.

**Comment by chiarazampolli** [comment]
There are some fixes in https://github.com/AliceO2Group/AliceO2/pull/13240/files, but now this should compile, since the new structure of the GLOQC classes is in O2 dev.

**Comment by chiarazampolli** [comment]
Hello, 
So far the error is not related to this PR, see :

Screenshot 2024-06-21 at 10 18 53

Pinging @ktf  (I don't know Sergio's github account...).

Chiara

**Comment by ktf** [comment]
Sergio is @singiamtel. This happens because the checkout takes too long on those machines. LLVM is a huge repository. I have asked the upstream maintainer to repack it and we have already increased the timeout. I will check if something weird is happening on the machine.

File: Modules/GLO/src/ITSTPCMatchingTask.cxx [hunk 1]
```diff
@@ -116,6 +139,8 @@ void ITSTPCMatchingTask::startOfCycle()
 
 void ITSTPCMatchingTask::monitorData(o2::framework::ProcessingContext& ctx)
 {
+
```
ktf, line 142:
```suggestion
```

Modules/GLO/src/ITSTPCMatchingTask.cxx
```diff
@@ -116,6 +139,7 @@ void ITSTPCMatchingTask::startOfCycle()
 
 void ITSTPCMatchingTask::monitorData(o2::framework::ProcessingContext& ctx)
 {
+  ILOG(Info) << "********** Starting monitoring" << ENDM;
```
Barthelemy, line 142: We recommend never logging from `monitorData` because this method is usually called at very high rate.

Barthelemy (changes requested):

Modules/GLO/src/ITSTPCMatchingTask.cxx
chiarazampolli: You are right, it was a leftover of my tests. Changed to debug.

chiarazampolli:
Hello @Barthelemy ,

Anything else for this PR?

Cheers,

Chiara

Barthelemy:
no, it is perfect :-)

Barthelemy (approved):

PR #2294, opened 2024-05-20
[EMCAL-657, EMCAL-846] Update EMCAL reductors

mfasDa:
- Occupancy reductor exports also mean and RMS
  per supermodule
- Bad channel reductor exports also the observables of
  non-good channels (bad+dead)
- Added documentation in all 3 reductors
- Add basic doxygen module structure for EMCAL


mfasDa:
Ready to merge

jokonig (approved):

PR #2297, opened 2024-05-21
[EMCAL-1133] Added Pedestal Channel Checker

sweyh99:
For histograms: PedestalChannelFECHG, PedestalChannelFECLG, PedestalChannelLEDMONHG, PedestalChannelLEDMONLG 
Checks if the Average pedestal value in expected range and Fraction of channels with pedestal > nsigma * average smaller than threshold

### DIFF
--- a/Modules/EMCAL/include/EMCAL/PedestalChannelCheck.h
+++ b/Modules/EMCAL/include/EMCAL/PedestalChannelCheck.h
@@ -1,0 +1,62 @@
+// Copyright 2019-2020 CERN and copyright holders of ALICE O2.
+// See https://alice-o2.web.cern.ch/copyright for details of the copyright holders.
+// All rights not expressly granted are reserved.
+//
+// This software is distributed under the terms of the GNU General Public
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+///
+/// \file   PedestalChannelCheck.h
+/// \author Sierra Cantway
+///
+
+#ifndef QC_MODULE_EMCAL_EMCALPEDESTALCHANNELCHECK_H
+#define QC_MODULE_EMCAL_EMCALPEDESTALCHANNELCHECK_H
+
+#include "QualityControl/CheckInterface.h"
+
+namespace o2::quality_control_modules::emcal
+{
+
+/// \brief  Check whether a plot is good or not.
+///
+/// \author Sierra Cantway
+class PedestalChannelCheck : public o2::quality_control::checker::CheckInterface
+{
+ public:
+  /// Default constructor
+  PedestalChannelCheck() = default;
+  /// Destructor
+  ~PedestalChannelCheck() override = default;
+
+  // Override interface
+  void configure() override;
+  Quality check(std::map<std::string, std::shared_ptr<MonitorObject>>* moMap) override;
+  void beautify(std::shared_ptr<MonitorObject> mo, Quality checkResult = Quality::Null) override;
+  std::string getAcceptedType() override;
+
+ private:
+  /************************************************
+   * threshold cuts                               *
+   ************************************************/
+
+  float mBadPedestalChannelADCRangeLow = 20.;       ///< Bad Low Threshold for mean ADC used in the Pedestal Channels check
+  float mBadPedestalChannelADCRangeHigh = 100.;     ///< Bad High Threshold for mean ADC used in the Pedestal Channels check
+  float mBadPedestalChannelOverMeanThreshold = 0.2; ///< Bad Threshold used in the time Pedestal Channels points outside of mean check
+
+  float mMedPedestalChannelADCRangeLow = 40.;       ///< Med Low Threshold for mean ADC used in the Pedestal Channels check
+  float mMedPedestalChannelADCRangeHigh = 80.;      ///< Med High Threshold for mean ADC used in the Pedestal Channels check
+  float mMedPedestalChannelOverMeanThreshold = 0.1; ///< Med Threshold used in the time Pedestal Channels points outside of mean check
+
+  float mSigmaPedestalChannel = 2.; ///< Number of sigmas used in the Pedestal Channel points outside of mean check
+
+  ClassDefOverride(PedestalChannelCheck, 1);
+};
+
+} // namespace o2::quality_control_modules::emcal
+
+#endif // QC_MODULE_EMCAL_EMCALPEDESTALCHANNELCHECK_H
--- a/Modules/EMCAL/src/PedestalChannelCheck.cxx
+++ b/Modules/EMCAL/src/PedestalChannelCheck.cxx
@@ -1,0 +1,202 @@
+// Copyright 2019-2020 CERN and copyright holders of ALICE O2.
+// See https://alice-o2.web.cern.ch/copyright for details of the copyright holders.
+// All rights not expressly granted are reserved.
+//
+// This software is distributed under the terms of the GNU General Public
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+///
+/// \file   PedestalChannelCheck.cxx
+/// \author Sierra Cantway
+///
+
+#include "EMCAL/PedestalChannelCheck.h"
+#include "QualityControl/MonitorObject.h"
+#include "QualityControl/Quality.h"
+#include "QualityControl/QcInfoLogger.h"
+#include <fairlogger/Logger.h>
+#include <DataFormatsQualityControl/FlagType.h>
+#include <DataFormatsQualityControl/FlagTypeFactory.h>
+// ROOT
+#include <TH1.h>
+#include <TH2.h>
+#include <TPaveText.h>
+#include <TLatex.h>
+#include <TList.h>
+#include <TRobustEstimator.h>
+#include <TMath.h>
+#include <ROOT/TSeq.hxx>
+#include <iostream>
+#include <vector>
+
+using namespace std;
+
+namespace o2::quality_control_modules::emcal
+{
+
+void PedestalChannelCheck::configure()
+{
+  // configure threshold-based checkers for bad quality
+  auto nBadPedestalChannelADCRangeLow = mCustomParameters.find("BadPedestalChannelADCRangeLow");
+  if (nBadPedestalChannelADCRangeLow != mCustomParameters.end()) {
+    try {
+      mBadPedestalChannelADCRangeLow = std::stof(nBadPedestalChannelADCRangeLow->second);
+    } catch (std::exception& e) {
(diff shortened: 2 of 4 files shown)


Modules/EMCAL/src/PedestalChannelCheck.cxx
```diff
@@ -0,0 +1,202 @@
...
+
+void PedestalChannelCheck::configure()
+{
+  // configure threshold-based checkers for bad quality
+  auto nBadPedestalChannelADCRangeLow = mCustomParameters.find("BadPedestalChannelADCRangeLow");
+  if (nBadPedestalChannelADCRangeLow != mCustomParameters.end()) {
+    try {
+      mBadPedestalChannelADCRangeLow = std::stof(nBadPedestalChannelADCRangeLow->second);
+    } catch (std::exception& e) {
+      ILOG(Error, Support) << fmt::format("Value {} not a float", nBadPedestalChannelADCRangeLow->second.data()) << ENDM;
```
mfasDa, line 49: fmt library not needed, please change the next iteration. As example, the line should look like
```c++
ILOG(Error, Support) << "Value " << nBadPedestalChannelADCRangeLow->second << " not a float" << ENDM;
```
Also, in case of C++ streams, for all objects that implement operator<< (among them std::string) you can put the object directly on the stream, as shown above. There is no need to access the C-string from the std::string. This is only necessary if you use print- or logging functions.

Modules/EMCAL/src/PedestalChannelCheck.cxx
```diff
@@ -0,0 +1,202 @@
...
+    }
+  }
+
+  auto nBadPedestalChannelADCRangeHigh = mCustomParameters.find("BadPedestalChannelADCRangeHigh");
+  if (nBadPedestalChannelADCRangeHigh != mCustomParameters.end()) {
+    try {
+      mBadPedestalChannelADCRangeHigh = std::stof(nBadPedestalChannelADCRangeHigh->second);
+    } catch (std::exception& e) {
+      ILOG(Error, Support) << fmt::format("Value {} not a float", nBadPedestalChannelADCRangeHigh->second.data()) << ENDM;
```
mfasDa, line 58: See above

Modules/EMCAL/src/PedestalChannelCheck.cxx
```diff
@@ -0,0 +1,202 @@
...
+    }
+  }
+
+  auto nBadPedestalChannelOverMeanThreshold = mCustomParameters.find("BadPedestalChannelOverMeanThreshold");
+  if (nBadPedestalChannelOverMeanThreshold != mCustomParameters.end()) {
+    try {
+      mBadPedestalChannelOverMeanThreshold = std::stof(nBadPedestalChannelOverMeanThreshold->second);
+    } catch (std::exception& e) {
+      ILOG(Error, Support) << fmt::format("Value {} not a float", nBadPedestalChannelOverMeanThreshold->second.data()) << ENDM;
```
mfasDa, line 67: See above

Modules/EMCAL/src/PedestalChannelCheck.cxx
```diff
@@ -0,0 +1,202 @@
...
+    }
+  }
+
+  // configure threshold-based checkers for medium quality
+  auto nMedPedestalChannelADCRangeLow = mCustomParameters.find("MedPedestalChannelADCRangeLow");
+  if (nMedPedestalChannelADCRangeLow != mCustomParameters.end()) {
+    try {
+      mMedPedestalChannelADCRangeLow = std::stof(nMedPedestalChannelADCRangeLow->second);
+    } catch (std::exception& e) {
+      ILOG(Error, Support) << fmt::format("Value {} not a float", nMedPedestalChannelADCRangeLow->second.data()) << ENDM;
```
mfasDa, line 77: See above

Modules/EMCAL/src/PedestalChannelCheck.cxx
```diff
@@ -0,0 +1,202 @@
...
+    }
+  }
+
+  auto nMedPedestalChannelADCRangeHigh = mCustomParameters.find("MedPedestalChannelADCRangeHigh");
+  if (nMedPedestalChannelADCRangeHigh != mCustomParameters.end()) {
+    try {
+      mMedPedestalChannelADCRangeHigh = std::stof(nMedPedestalChannelADCRangeHigh->second);
+    } catch (std::exception& e) {
+      ILOG(Error, Support) << fmt::format("Value {} not a float", nMedPedestalChannelADCRangeHigh->second.data()) << ENDM;
```
mfasDa, line 86: See above

Modules/EMCAL/include/EMCAL/PedestalChannelCheck.h
```diff
@@ -0,0 +1,62 @@
...
+   * threshold cuts                               *
+   ************************************************/
+
+  float mBadPedestalChannelADCRangeLow = 20.;       ///< Bad Low Threshold for mean ADC used in the Pedestal Channels check
+  float mBadPedestalChannelADCRangeHigh = 100.;     ///< Bad High Threshold for mean ADC used in the Pedestal Channels check
+  float mBadPedestalChannelOverMeanThreshold = 0.2; ///< Bad Threshold used in the time Pedestal Channels points outside of mean check
+
+  float mMedPedestalChannelADCRangeLow = 40.;       ///< Med Low Threshold for mean ADC used in the Pedestal Channels check
+  float mMedPedestalChannelADCRangeHigh = 80.;      ///< Med High Threshold for mean ADC used in the Pedestal Channels check
+  float mMedPedestalChannelOverMeanThreshold = 0.1; ///< Med Threshold used in the time Pedestal Channels points outside of mean check
```
mfasDa, line 53: Expected pedestal values are in the range between 20 and 50 (that's where the min. bunch amp checker searches them). Suspicious it gets above 75 (warning), bad it becomes for pedestals > 100. Please change the default values in the next iteration. For the moment the configuration is sufficient.

Modules/EMCAL/include/EMCAL/PedestalChannelCheck.h
```diff
@@ -0,0 +1,62 @@
...
+#define QC_MODULE_EMCAL_EMCALPEDESTALCHANNELCHECK_H
+
+#include "QualityControl/CheckInterface.h"
+
+namespace o2::quality_control_modules::emcal
+{
+
+/// \brief  Check whether a plot is good or not.
+///
+/// \author Sierra Cantway
```
mfasDa, line 27: Doxygen comments should start with

```c++
/// \class PedestalChannelCheck
/// \brief Check for large amount of pedestal outliers in pedestal histograms 
/// \ingroup EMCALQCCheckers
/// \author Sierra Cantway
/// \since May 22nd, 2024
///
/// A bit more detailed description here ...
```

Also the comment `Check whether a plot is good or not.` is not really useful.

Obviously the parameters should be adapted for each case. Checkers are in the group `EMCALQCCheckers`, which is defined in `doxymodules.h`.

Modules/EMCAL/src/PedestalChannelCheck.cxx
```diff
@@ -0,0 +1,202 @@
...
+    }
+  }
+
+  auto nMedPedestalChannelOverMeanThreshold = mCustomParameters.find("MedPedestalChannelOverMeanThreshold");
+  if (nMedPedestalChannelOverMeanThreshold != mCustomParameters.end()) {
+    try {
+      mMedPedestalChannelOverMeanThreshold = std::stof(nMedPedestalChannelOverMeanThreshold->second);
+    } catch (std::exception& e) {
+      ILOG(Error, Support) << fmt::format("Value {} not a float", nMedPedestalChannelOverMeanThreshold->second.data()) << ENDM;
```
mfasDa, line 95: See above

Modules/EMCAL/src/PedestalChannelCheck.cxx
```diff
@@ -0,0 +1,202 @@
...
+    }
+  }
+
+  // configure nsigma-based checkers
+  auto nSigmaPedestalChannel = mCustomParameters.find("SigmaPedestalChannel");
+  if (nSigmaPedestalChannel != mCustomParameters.end()) {
+    try {
+      mSigmaPedestalChannel = std::stof(nSigmaPedestalChannel->second);
+    } catch (std::exception& e) {
+      ILOG(Error, Support) << fmt::format("Value {} not a float", nSigmaPedestalChannel->second.data()) << ENDM;
```
mfasDa, line 105: See above

Modules/EMCAL/src/PedestalChannelCheck.cxx
```diff
@@ -0,0 +1,202 @@
...
+      msg->Clear();
+      msg->AddText("Data OK: Average Pedestal & Fraction of Outliers Within Limits");
+      msg->SetFillColor(kGreen);
+      msg->Draw();
+    } else if (checkResult == Quality::Bad) {
+      ILOG(Debug, Devel) << "Quality::Bad, setting to red";
+      stringstream msg_bad;
+      msg_bad << "Error: Average Pedestal outside limits (" << mBadPedestalChannelADCRangeLow << "-" << mBadPedestalChannelADCRangeHigh << ") or Fraction of Outliers Above " << mBadPedestalChannelOverMeanThreshold << std::endl;
+      msg->Clear();
+      msg->AddText(msg_bad.str().c_str());
```
mfasDa, line 187: You should make use of the flag you defined in the check function. This allows to provide more meaningful messages. Please change in next iteration. As example see https://github.com/AliceO2Group/QualityControl/blob/master/Modules/EMCAL/src/RawCheck.cxx (beautify function)

Modules/EMCAL/src/PedestalChannelCheck.cxx
```diff
@@ -0,0 +1,202 @@
...
+      msg->AddText("Call EMCAL oncall");
+      msg->SetFillColor(kRed);
+      msg->Draw();
+    } else if (checkResult == Quality::Medium) {
+      ILOG(Debug, Devel) << "Quality::medium, setting to orange";
+      stringstream msg_med;
+      msg_med << "Warning: Average Pedestal outside limits (" << mMedPedestalChannelADCRangeLow << "-" << mMedPedestalChannelADCRangeHigh << ") or Fraction of Outliers Above " << mMedPedestalChannelOverMeanThreshold << std::endl;
```
mfasDa, line 194: Also here the message in the tag would be better to use.

Modules/EMCAL/include/EMCAL/PedestalChannelCheck.h
```diff
@@ -0,0 +1,62 @@
...
+  /// Default constructor
+  PedestalChannelCheck() = default;
+  /// Destructor
+  ~PedestalChannelCheck() override = default;
+
+  // Override interface
+  void configure() override;
+  Quality check(std::map<std::string, std::shared_ptr<MonitorObject>>* moMap) override;
+  void beautify(std::shared_ptr<MonitorObject> mo, Quality checkResult = Quality::Null) override;
+  std::string getAcceptedType() override;
```
mfasDa, line 40: Would be nice to add some function comments as doxygen comments. They don't need to be long, just to express what the functions are doing. They should be in the form

```c++
/// \brief ...
/// \param firstparam This is the information content of the first param
/// \param secondparam This is the information content of the second param
...
/// \return (for non-void functions)
```

mfasDa (approved):
There are a few homeworks for the next iteration, but for the moment the checker is good to go.

PR #2298, opened 2024-05-22
[EMCAL-1143] Provide reduction to FEC

mfasDa:
Reduce occupancy histogram to count rate, average
and RMS per FEC.

### DIFF
--- a/Modules/EMCAL/CMakeLists.txt
+++ b/Modules/EMCAL/CMakeLists.txt
@@ -2,7 +2,7 @@
 
 add_library(O2QcEMCAL)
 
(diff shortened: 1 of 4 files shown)


mfasDa (commented):
Will not be used very often, as it exports 2400 Datapoints (entries, mean and RMS per FEC, with 800 FECs), but if a supermodule shows instabilities in the trending we need to be able to determine - on request - which FEC is responsible for the instabilities. It doesn't help us just to know the supermodule. Recently we had such a case in one of the supermodules, and it took quite some time to find the FEC, with a lot of manual testing. If we would have had the trend points for each FEC at hand the problematic FEC would have been identified immediately.

@Barthelemy For such purpose a time series database would be ideal, because this is more debug info. Unlike the trend graph per supermodule we don't need this one persistently, this data can go after a day, it is debug information. However fort the moment it is a bit annoying, as if we find again instabilities in the supermodule we need to stop the run, enable the reductor, run for some time to get the data per supermodule, switch off the reductor, and go back to the global run. And I am afraid we cannot enable it for all runs as the data volume would be too large. I think most time series databases remove timestamps automatically after some time.

jokonig (approved):

jokonig (approved):

PR #2300, opened 2024-05-23
QC-1164 Modernize example code in the Skeleton module

knopers8:


Barthelemy (approved):
Excellent !
Thanks a lot

Barthelemy:
I see that a number of tests are failing though. They probably rely on some histograms of the Skeleton module

PR #2301, opened 2024-05-24
since too long data description is now safe, no need to warn

knopers8:
I think it's enough to have a debug log now, the warnings generate quite a lot of noise at the moment, while "too long" description is now safe to have.


knopers8:
See a typical global staging environment start to understand how many warnings we get.

Barthelemy:
@ktf there seems to be an issue with with the O2 build on MacOS-arm: 
```
## sw/BUILD/O2-latest/log
 22/501 Test  #19: framework:crash ......................................................................***Failed    3.48 sec
```

PR #2305, opened 2024-05-27
QC-1179 Fix TrendingTask test and allow it to be ran with CI

knopers8:
The test could be extended of course, but this already brings back what was tested before.


justonedev1 (approved):
I don't see anything wrong. However, to be sure, it might be better to wait for Barth.

Barthelemy (approved):
thanks, seems good !

PR #2312, opened 2024-05-29
QC-1153 Indicate that an object is a moving window in its title

knopers8:


knopers8:
Just a comment for @Barthelemy and @justonedev1 . The reported window duration is the merged object validity. Usually it will not be equal to the cycle duration, because while objects from different EPNs will be subject to the same cycle duration, the cycles will be out of phase. This means, that the merged validity is going to be within `[cycleDuration, 2*cycleDuration]`.

In case that someone asks why, this is why.

Framework/src/MonitorObjectCollection.cxx
```diff
@@ -120,6 +121,41 @@ const std::string& MonitorObjectCollection::getTaskName() const
...
+  if (hours > 0) {
+    result << hours << "h";
+  }
+  if (minutes > 0 || hours > 0) {
+    result << minutes << "m";
+  }
+  result << seconds << "s";
+  if (durationMs < 1000) {
+    result << durationMs << "ms";
+  }
```
Barthelemy, line 144: the ms are ignored if the duration is more than 1 second, correct ?

Barthelemy (approved):

Framework/src/MonitorObjectCollection.cxx
knopers8: Correct. I don't think there is need for such verbosity if the validity is longer. It's just for shifters to be aware what they are looking at.

Barthelemy: I agree

PR #2313, opened 2024-05-29
[QC-1182] Issue with CustomParameters in PP

Barthelemy:


Framework/include/QualityControl/CustomParameters.h
```diff
@@ -149,6 +149,8 @@ class CustomParameters
 
   std::unordered_map<std::string, std::string>::const_iterator end() const;
 
+  size_t size() const;
```
knopers8, line 152: Since `size()` quite non-intuitive in this case (at least to me), could you document that it returns the total count for all the kv pairs for all beam/type combinations?

Framework/src/PostProcessingConfig.cxx
```diff
@@ -50,6 +50,11 @@ PostProcessingConfig::PostProcessingConfig(const std::string& id, const boost::p
     stopTriggers.push_back(stopTrigger.second.get_value<std::string>());
   }
   auto ppTree = config.get_child("qc.postprocessing." + id);
+  if (ppTree.count("taskParameters") > 0) {
+    for (const auto& [key, value] : ppTree.get_child("taskParameters")) {
+      customParameters.set(key, value.get_value<std::string>());
+    }
+  }
```
knopers8, line 57: It's a different behaviour than for QC Tasks, where it's:
```
  if (taskTree.count("extendedTaskParameters") > 0) {
    ts.customParameters.populateCustomParameters(taskTree.get_child("extendedTaskParameters"));
  } else if (taskTree.count("taskParameters") > 0) {
    for (const auto& [key, value] : taskTree.get_child("taskParameters")) {
      ts.customParameters.set(key, value.get_value<std::string>());
    }
  }
```
In particular, the code for PP will merge `taskParameters` with `extendedTaskParameters`, while the code for QC tasks will take one or the other. Could you unify those for the sake of consistency?

knopers8 (changes requested):

Framework/include/QualityControl/CustomParameters.h
Barthelemy: sure, done

Framework/src/PostProcessingConfig.cxx
Barthelemy: Good catch, this is not what I intended. Fixed.

knopers8 (approved):

PR #2314, opened 2024-05-29
[EMCAL-767, EMCAL-846, EMCAL-1143] Additional protection trending

mfasDa:
- Protection against
  + Missing historam / wrong histogram type
  + Geometry and mapping errors
- Tasks / Reductors:
  + CalibMonitoringTask
  + TimeCalibParamReductor
  + BadChannelMapReductor
  + OccupancyToFECReductor


mfasDa (commented):
All possible error sources should be protected in these classes, I cannot think of what to protect more

PR #2315, opened 2024-05-29
ITS: Chip Status Task

IsakovAD:
Added new QC Task for ITS Module; 

This task will provide histograms of the Chips Status for the last N cycles in form of 3 histograms for IB/ML/OL

The new task expected to use in ENTIRE merging mode

### DIFF
--- a/Modules/ITS/include/ITS/ITSChipStatusTask.h
+++ b/Modules/ITS/include/ITS/ITSChipStatusTask.h
@@ -19,6 +19,7 @@
 
 #include "QualityControl/TaskInterface.h"
 #include "Common/Utils.h"
+#include "Common/TH2Ratio.h"
 
 #include <TH1D.h>
 #include <TH2D.h>
@@ -38,9 +39,9 @@ class Stack
  public:
   std::vector<std::vector<int>> stack;
   Stack() : size_max(0), current_element_id(0) {}
-  Stack(int nsizey, int nsizex) : size_max(nsizey), current_element_id(0)
+  Stack(int nsizex, int nsizey) : size_max(nsizex), current_element_id(0)
   {
-    stack.resize(nsizey, std::vector<int>(nsizex, 0));
+    stack.resize(nsizex, std::vector<int>(nsizey, 0));
   }
   void push(const std::vector<int>& element)
   {
@@ -73,19 +74,23 @@ class ITSChipStatusTask final : public TaskInterface
  private:
   void getParameters(); // get Task parameters from json file
   void setAxisTitle(TH1* object, const char* xTitle, const char* yTitle);
+  void Beautify();
   static constexpr int NLayer = 7;
   static constexpr int NLayerIB = 3;
   const int nChipsPerLayer[NLayer] = { 108, 144, 180, 2688, 3360, 8232, 9408 };
   const int StaveBoundary[NLayer + 1] = { 0, 12, 28, 48, 72, 102, 144, 192 };
   const int ChipBoundary[NLayer + 1] = { 0, 108, 252, 432, 3120, 6480, 14712, 24120 };
+  const int ChipsBoundaryLayers[8] = { 0, 108, 252, 432, 3120 - 432, 6480 - 432, 14712 - 6480, 24120 - 6480 }; // needed for drawing labels and lines for histogram
   const int ChipsBoundaryBarrels[4] = { 0, 432, 6480, 24120 };
   static constexpr int NFees = 48 * 3 + 144 * 2;
   TString LayerBinLabels[11] = { "L0", "L1", "L2", "L3B", "L3T", "L4B", "L4T", "L5B", "L5T", "L6B", "L6T" };
-  TH2D* DeadChips[3];
+  std::shared_ptr<TH2FRatio> DeadChips[3];
   Stack* ChipsStack[3];
+  Stack* TFsStack[3];
   int nQCCycleToMonitor = 50;
   TString BarrelNames[3] = { "IB", "ML", "OB" };
   std::vector<int> CurrentDeadChips[3]; // Vectors of Dead Chips in current QC cycle
+  std::vector<int> CurrentTFs[3];       // Vectors of Dead Chips in current QC cycle
 };
 
 } // namespace o2::quality_control_modules::its
--- a/Modules/ITS/include/ITS/ITSChipStatusTask.h
+++ b/Modules/ITS/include/ITS/ITSChipStatusTask.h
@@ -88,7 +88,6 @@ class ITSChipStatusTask final : public TaskInterface
   std::vector<int> CurrentDeadChips[3]; // Vectors of Dead Chips in current QC cycle
 };
 
-
 } // namespace o2::quality_control_modules::its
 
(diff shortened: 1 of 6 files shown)


IsakovAD:
@iravasen @nicolovalle 

Could you please have a look at this commit? 

The problem with merging should be fixed with ENTIRE mode, so I think we can use it at P2; In worse case, we always can disable this task.

Modules/ITS/include/ITS/ITSChipStatusTask.h
```diff
@@ -0,0 +1,93 @@
...
+class Stack
+{
+ private:
+  int size_max;
+  int current_element_id;
+
+ public:
+  std::vector<std::vector<int>> stack;
+  Stack() : size_max(0), current_element_id(0) {}
+  Stack(int nsizey, int nsizex) : size_max(nsizey), current_element_id(0)
```
iravasen, line 41: Not fundamental but probably you can invert `nsizey` with `nsizex` since when you define a pointer to this `class` you use `new Stack(nQCCycleToMonitor, ChipsBoundaryBarrels[i + 1] - ChipsBoundaryBarrels[i]);` but then the first element refers to the `x` axis of the TH2D while the second element refers to the `y` axis.

Modules/ITS/src/ITSChipStatusTask.cxx
```diff
@@ -0,0 +1,178 @@
...
+      iLayerEnd = 2;
+    } else if (i == 1) {
+      iLayerBegin = 3;
+      iLayerEnd = 4;
+    } else {
+      iLayerBegin = 5;
+      iLayerEnd = 6;
+    }
+    const int ChipBoundary[8] = { 0, 108, 252, 432, 3120 - 432, 6480 - 432, 14712 - 6480, 24120 - 6480 }; // needed for drawing labels and lines for histogram
+    const int StaveBoundary[NLayer] = { 0, 12, 28, 0, 24, 0, 42 };
```
iravasen, line 67: `ChipBoundary[8]` and `StaveBoundary[NLayer] are already in the header file. I would move these two to the header file if you really needed them (eventually renaming them properly).

Modules/ITS/src/ITSChipStatusTask.cxx
```diff
@@ -0,0 +1,178 @@
...
+      if (iLayer < iLayerEnd) {
+        auto l = new TLine(0, ChipBoundary[iLayer + 1] - 0.5, nQCCycleToMonitor, ChipBoundary[iLayer + 1] - 0.5);
+        DeadChips[i]->GetListOfFunctions()->Add(l);
+      }
+    }
+    int nChipsPerStave = i == 0 ? 9 : i == 1 ? 112
+                                             : 196;
+    for (int iy = 0; iy < DeadChips[i]->GetNbinsY(); iy++) {
+      if (iy % nChipsPerStave == 0) {
+        auto l = new TLine(0, iy - 0.5, nQCCycleToMonitor, iy - 0.5);
```
iravasen, line 83: I would remove them, I think they make confusion in the plot. I see two possibilities:
- add labels to the y axis of the TH2D which allow to identify the chips when hovering a bin in the GUI. Example of a label: L0_00_C0. But this becomes complex for OB. 
- keep the y axis as it is and have a table somewhere which allow us to quickly convert the chipID (chip counter) on the y axis with layer_stavenum_chipInStave.

Modules/ITS/src/ITSChipStatusTask.cxx
```diff
@@ -0,0 +1,178 @@
...
+        l->SetLineColor(40);
+        l->SetLineWidth(1);
+
+        DeadChips[i]->GetListOfFunctions()->Add(l);
+
+        if (iy >= ChipBoundary[Layer_Draw + 1]) {
+          Layer_Draw++;
+        }
+        TLatex* msg = new TLatex(-3, iy, Form("%d", (int)(iy) / nChipsPerStave - StaveBoundary[Layer_Draw]));
```
iravasen, line 92: do we need them?

Modules/ITS/src/ITSChipStatusTask.cxx
```diff
@@ -0,0 +1,178 @@
...
+
+  nQCCycleToMonitor = o2::quality_control_modules::common::getFromConfig<int>(mCustomParameters, "nQCCycleToMonitor", nQCCycleToMonitor);
+}
+
+void ITSChipStatusTask::endOfCycle()
+{
+
+  ILOG(Debug, Devel) << "endOfCycle" << ENDM;
+  for (int iBarrel = 0; iBarrel < 3; iBarrel++) {
+    ChipsStack[iBarrel]->push(CurrentDeadChips[iBarrel]);
```
iravasen, line 158: A question on `push`. If I understand well, if we reached a QC cycle number which is > `nQCCycleToMonitor`, you call `std::rotate` to shift all elements up by 1 position and then you store the new element at the end. But after that you fill the TH2D meaning that if before you had a vector of dead chips at QCcycle = i, now the same vector is at QCcycle = i-1 which means that the TH2D will have the same entry twice, once at x=i and then at x=i-1. Probably I'm missing something. Thanks.

Modules/ITS/src/ITSChipStatusTask.cxx
```diff
@@ -0,0 +1,178 @@
...
+  int id = 0;
+  for (auto chipID : aliveChips) {
+    int iBarrel = id < ChipsBoundaryBarrels[1] ? 0 : id < ChipsBoundaryBarrels[2] ? 1
+                                                                                  : 2;
+    if ((int)chipID != 1) {
+      int ChipID_fill = id - ChipsBoundaryBarrels[iBarrel];
+      if ((ChipID_fill < 0) || (ChipID_fill >= CurrentDeadChips[iBarrel].size())) {
+        ILOG(Warning, Devel) << " prolematic ID for chip: " << ChipID_fill << ENDM;
+      } else {
+        CurrentDeadChips[iBarrel][ChipID_fill]++;
```
nicolovalle, line 140: Hi @IsakovAD , do I well understand that this CurrentDeadChips is counting the number of TFs in the cycle for which the ChipID_fill was dead.. and this information will go to the TH2?
May I suggest to normalize it by the number of TFs (=monitorData cycles) processed in the QC cycle ? This would give absolute and understandable value non depending on the settings and on fluctuactions.

Another possibility is to use some absolute time unit on the x axis, e.g. orbit number instead of QC cycle.

Modules/ITS/src/ITSChipStatusTask.cxx
IsakovAD: You are right, but when I call rotate() function, I use not push_back() but just back() command, which fixes the problem you mentioned

Modules/ITS/src/ITSChipStatusTask.cxx
IsakovAD: Unfortunately, I can't use adaptive x-axis. 



But to show fraction, now we have TH2Fraction class which stores number of TFs per QCCycle in denominator

iravasen (approved):

PR #2320, opened 2024-05-31
Qc 978 Propagate QC flags to BK/RCT

justonedev1:
Bkp api still does not implement necessary proto, so I am putting this here as a draft, so @knopers8  can take a look

### DIFF
--- a/Framework/include/QualityControl/BookkeepingQualitySink.h
+++ b/Framework/include/QualityControl/BookkeepingQualitySink.h
@@ -18,6 +18,7 @@
 #define QUALITYCONTROL_BOOKKEEPINGQUALITYSINK_H
 
 #include <DataFormatsQualityControl/QualityControlFlagCollection.h>
+#include <Framework/CompletionPolicy.h>
 #include <Framework/DataProcessorSpec.h>
 #include <Framework/Task.h>
 #include "Provenance.h"
@@ -41,6 +42,8 @@ class BookkeepingQualitySink : public framework::Task
   void endOfStream(framework::EndOfStreamContext& context) override;
   void stop() override;
 
+  static void customizeInfrastructure(std::vector<framework::CompletionPolicy>& policies);
+  static framework::DataProcessorLabel getLabel() { return { "BookkeepingQualitySink" }; }
   static void send(const std::string& grpcUri, const FlagsMap&, Provenance);
 
  private:
--- a/Framework/include/QualityControl/BookkeepingQualitySink.h
+++ b/Framework/include/QualityControl/BookkeepingQualitySink.h
@@ -29,7 +29,8 @@ namespace o2::quality_control::core
 class BookkeepingQualitySink : public framework::Task
 {
  public:
-  using FlagsMap = std::unordered_map<std::string, std::unique_ptr<QualityControlFlagCollection>>;
+  // we are using map here instead of the set, because items in the map are changeable, however items of the set are not.
+  using FlagsMap = std::unordered_map<std::string /*detector*/, std::unique_ptr<QualityControlFlagCollection>>;
   using SendCallback = std::function<void(const std::string& grpcUri, const FlagsMap&, Provenance)>;
 
   // sendCallback is mainly used for testing without the necessity to do grpc calls
--- a/Framework/include/QualityControl/BookkeepingQualitySink.h
+++ b/Framework/include/QualityControl/BookkeepingQualitySink.h
@@ -48,7 +48,7 @@ class BookkeepingQualitySink : public framework::Task
 
  private:
   std::string mGrpcUri;
-  Provenance mType;
+  Provenance mProvenance;
   SendCallback mSendCallback;
   FlagsMap mQualityObjectsMap;
 
--- a/Framework/include/QualityControl/BookkeepingQualitySink.h
+++ b/Framework/include/QualityControl/BookkeepingQualitySink.h
@@ -20,7 +20,7 @@
 #include <DataFormatsQualityControl/QualityControlFlagCollection.h>
 #include <Framework/DataProcessorSpec.h>
 #include <Framework/Task.h>
-#include "ProvenanceToQCRunType.h"
+#include "Provenance.h"
 
 namespace o2::quality_control::core
 {
--- a/Framework/include/QualityControl/BookkeepingQualitySink.h
+++ b/Framework/include/QualityControl/BookkeepingQualitySink.h
@@ -1,0 +1,56 @@
+// Copyright 2024 CERN and copyright holders of ALICE O2.
+// See https://alice-o2.web.cern.ch/copyright for details of the copyright holders.
+// All rights not expressly granted are reserved.
+//
+// This software is distributed under the terms of the GNU General Public
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+///
+/// \file   BookkeepingQualitySink.h
+/// \author Michal Tichak
+///
+
+#ifndef QUALITYCONTROL_BOOKKEEPINGQUALITYSINK_H
+#define QUALITYCONTROL_BOOKKEEPINGQUALITYSINK_H
+
+#include <DataFormatsQualityControl/QualityControlFlagCollection.h>
+#include <Framework/DataProcessorSpec.h>
+#include <Framework/Task.h>
+#include "ProvenanceToQCRunType.h"
+
+namespace o2::quality_control::core
+{
+
+// This class gathers all QualityObjects from it's inputs, converting them to flags + sending them to grpc RCT/BKP when stop() is invoked
+class BookkeepingQualitySink : public framework::Task
+{
+ public:
+  using FlagsMap = std::unordered_map<std::string, std::unique_ptr<QualityControlFlagCollection>>;
+  using SendCallback = std::function<void(const std::string& grpcUri, const FlagsMap&, Provenance)>;
+
+  // sendCallback is mainly used for testing without the necessity to do grpc calls
+  BookkeepingQualitySink(const std::string& grpcUri, Provenance, SendCallback sendCallback = send);
+
+  void run(framework::ProcessingContext&) override;
+
+  void endOfStream(framework::EndOfStreamContext& context) override;
+  void stop() override;
+
+  static void send(const std::string& grpcUri, const FlagsMap&, Provenance);
+
+ private:
+  std::string mGrpcUri;
+  Provenance mType;
+  SendCallback mSendCallback;
+  FlagsMap mQualityObjectsMap;
+
+  void sendAndClear();
+};
+
+} // namespace o2::quality_control::core
+
+#endif
--- a/Framework/include/QualityControl/ProvenanceToQCRunType.h
+++ b/Framework/include/QualityControl/ProvenanceToQCRunType.h
@@ -1,0 +1,40 @@
+// Copyright 2024 CERN and copyright holders of ALICE O2.
+// See https://alice-o2.web.cern.ch/copyright for details of the copyright holders.
+// All rights not expressly granted are reserved.
+//
+// This software is distributed under the terms of the GNU General Public
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+///
+/// \file   ProvenanceToQCRunType.h
+/// \author Michal Tichak
+///
+
+#include <stdexcept>
+#include <string>
+namespace o2::quality_control::core
+{
+
+enum class Provenance {
+  SyncQc,
+  AsyncQc
+};
+
+inline Provenance toEnum(const std::string& provenance)
+{
+  if (provenance == "qc_mc") {
+    return Provenance::AsyncQc;
+  }
+
+  if (provenance == "qc" || provenance == "qc_async") {
+    return Provenance::SyncQc;
+  }
+
+  throw std::runtime_error{ "unknown provenance flag: " + provenance };
+}
+
+} // namespace o2::quality_control::core
--- a/Framework/include/QualityControl/ProvenanceToQCRunType.h
+++ b/Framework/include/QualityControl/ProvenanceToQCRunType.h
@@ -21,17 +21,22 @@ namespace o2::quality_control::core
 
 enum class Provenance {
   SyncQc,
-  AsyncQc
+  AsyncQc,
+  MCQC
 };
 
 inline Provenance toEnum(const std::string& provenance)
 {
   if (provenance == "qc_mc") {
-    return Provenance::AsyncQc;
+    return Provenance::MCQC;
   }
 
-  if (provenance == "qc" || provenance == "qc_async") {
+  if (provenance == "qc") {
     return Provenance::SyncQc;
+  }
+
+  if (provenance == "qc_async") {
+    return Provenance::AsyncQc;
   }
 
   throw std::runtime_error{ "unknown provenance flag: " + provenance };
--- a/Framework/src/BookkeepingQualitySink.cxx
+++ b/Framework/src/BookkeepingQualitySink.cxx
@@ -59,11 +59,11 @@ void BookkeepingQualitySink::send(const std::string& grpcUri, const BookkeepingQ
     try {
       switch (type) {
         case Provenance::SyncQc:
+        case Provenance::AsyncQc:
           qcClient->createForDataPass(flagCollection->getRunNumber(), flagCollection->getPassName(), flagCollection->getDetector(), bkpQcFlags);
           break;
-        case Provenance::AsyncQc:
-          // NOTE: There is productionName in Bookkeaping API instead of passName, but I don't seen anythin like production around here.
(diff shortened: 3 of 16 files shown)


Framework/src/QualitySender.cxx
```diff
@@ -0,0 +1,69 @@
...
+void send(const std::string grpcUri, const FlagsMap& flags)
+{
+  for (const auto& [_, flagCollection] : flags) {
+    for (const auto& flag : *flagCollection) {
+      //TODO: fill with bookkeeping api when ready
+    }
+  }
+}
+
+void QualitySender::stop()
```
knopers8, line 57: I would also do the same in case of an EndOfStream and ensure we do it just once. It's a matter of implementing also another callback, like this one: https://github.com/AliceO2Group/QualityControl/blob/cc46357fceebcf8f0e00289e8a96d06756122f00/Framework/src/TaskRunner.cxx#L320

EndOfStream is considered the right moment to wrap up knowing that no more data will come. `stop` can be the last resort, which in principle should always be called anyway, but I would rather not rely just on that.

Framework/test/testQualitySender.cxx
```diff
@@ -0,0 +1,72 @@
...
+#include "QualityControl/InfrastructureGenerator.h"
+#include "QualityControl/QualitySender.h"
+
+using namespace o2;
+using namespace o2::framework;
+using namespace o2::utilities;
+
+void customize(std::vector<CompletionPolicy>& policies)
+{
+  DataSampling::CustomizeInfrastructure(policies);
```
knopers8, line 27:
```suggestion
```
You are not using data sampling here, you can remove it.

Framework/src/QualitySender.cxx
```diff
@@ -0,0 +1,69 @@
...
+
+QualitySender::QualitySender(const std::string& grpcUri) : mGrpcUri{ grpcUri } {}
+
+void QualitySender::run(framework::ProcessingContext& context)
+{
+  for (auto const& ref : framework::InputRecordWalker(context.inputs())) {
+    try {
+      mQualityObjects.push_back(std::move(framework::DataRefUtils::as<core::QualityObject>(ref)));
+    } catch (...) {
+    }
```
knopers8, line 18:
```suggestion
    } catch (...) {
      ILOG(Warn, Support) << "Unexpected message received, QualityObject expected" << ENDM;
    }
```

Framework/test/testQualitySender.cxx
```diff
@@ -0,0 +1,72 @@
...
+#include <QualityControl/QualitySender.h>
+#include <QualityControl/QualityObject.h>
+
+using namespace o2::configuration;
+
+WorkflowSpec defineDataProcessing(ConfigContext const&)
+{
+  WorkflowSpec specs;
+
+  quality_control::flag_sender::QualitySender sender("grpcUri");
```
knopers8, line 45: For the sake of more precise testing later, you could consider `QualitySender` constructor to take a callback to send data to Bookkeeping, while for testing you could put there some testing logic instead, which could print out e.g. a fatal log if data is wrong.

Framework/src/QualitySender.cxx
```diff
@@ -0,0 +1,69 @@
...
+  converter(*qualityObject);
+  return converter.getResult();
+}
+
+auto collectionFromQualityObject(const core::QualityObject& qualityObject) -> std::unique_ptr<QualityControlFlagCollection>
+{
+  return std::make_unique<QualityControlFlagCollection>(
+    qualityObject.getName(),
+    qualityObject.getDetectorName(),
+    core::gFullValidityInterval, // we don't need exact run duration because each flag have it's own range.
```
knopers8, line 34: Yes and no. If a part of run does not get any flag, such timespan is marked flagged `UnknownQuality`: https://github.com/AliceO2Group/QualityControl/blob/cc46357fceebcf8f0e00289e8a96d06756122f00/Framework/src/QualitiesToFlagCollectionConverter.cxx#L82

Thus we would need to make sure with BKP devs that e.g. sending them a flag `UnknownQuality` spanning from time `0` to an arbitrary time during the run will be handled by BKP correctly.

knopers8 (changes requested):
Thanks! It's a nice proof of concept!

Some next steps that you can already do without needing the Bookkeeping API updated are:
- add an config parameter for Check and Aggregator to enable sending Qualities to Bookkeeping.
- if there is at least one of such parameters enabled, create QualitySender in the QC workflow. you can do that by modifying `InfrastructureGenerator` in the parts that concern full workflows and remote ones. We can sit together to show you in detail.
- ensure that `QualitySender` behaves correctly in sequences with multiple runs (having a clean state at each new `start()`)

Framework/src/QualitySender.cxx
```diff
@@ -0,0 +1,69 @@
...
+namespace o2::quality_control::flag_sender
+{
+
+QualitySender::QualitySender(const std::string& grpcUri) : mGrpcUri{ grpcUri } {}
+
+void QualitySender::run(framework::ProcessingContext& context)
+{
+  for (auto const& ref : framework::InputRecordWalker(context.inputs())) {
+    try {
+      mQualityObjects.push_back(std::move(framework::DataRefUtils::as<core::QualityObject>(ref)));
```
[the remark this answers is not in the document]
justonedev1: The reason why I am doing it this way is that I don't know the validity range of given run.

Framework/include/QualityControl/ProvenanceToQCRunType.h
```diff
@@ -0,0 +1,40 @@
...
+#include <string>
+namespace o2::quality_control::core
+{
+
+enum class QCRunType {
+  Data,
+  Simulation
+};
+
+inline QCRunType toRunType(const std::string& provenance)
```
knopers8, line 27: We call run type those: https://github.com/AliceO2Group/Control/blob/f894c3c8b1b7597fed5bc6281d0b195bc7e088cd/common/runtype/runtype.go#L37

Perhaps we could just have a converter from string to enumerated Provenance? Then you could handle `Provenance::SyncQC` and `Provenance::AsyncQC` the same way when sending flags to BKP.

knopers8 (changes requested):

Framework/src/BookkeepingQualitySink.cxx
```diff
@@ -0,0 +1,139 @@
...
+        .flagTypeId = flag.getFlag().getID(),
+        .from = flag.getStart() == gFullValidityInterval.getMin() ? std::nullopt : std::optional<uint64_t>{ flag.getStart() },
+        .to = flag.getEnd() == gFullValidityInterval.getMax() ? std::nullopt : std::optional<uint64_t>{ flag.getEnd() },
+        .origin = flag.getSource(),
+        .comment = flag.getComment() });
+    }
+
+    if (bkpQcFlags.empty()) {
+      continue;
+    }
```
knopers8, line 55: Since `bkpQcFlags` will basically have the same number of elements as `*flagCollection`, could you do input validation earlier, i.e. continue if `flagCollection->empty()` at the beginning of the loop?

Framework/src/BookkeepingQualitySink.cxx
```diff
@@ -0,0 +1,139 @@
...
+
+    try {
+      switch (type) {
+        case Provenance::SyncQC:
+        case Provenance::AsyncQC:
+          qcClient->createForDataPass(flagCollection->getRunNumber(), flagCollection->getPassName(), flagCollection->getDetector(), bkpQcFlags);
```
knopers8, line 61: @martinboulais , @xsalonx can you confirm or deny that we should indeed use `createForDataPass` when uploading flags for synchronous QC? I remember you were considering creating a separate interface function for that.

If we should use this one, what is the pass name that you recognize as synchronous QC?

Framework/src/BookkeepingQualitySink.cxx
```diff
@@ -0,0 +1,139 @@
...
+      }
+    } catch (const std::runtime_error& err) {
+      ILOG(Error, Support) << "Failed to send flags for detector: " << flagCollection->getDetector()
+                           << " and pass: " << flagCollection->getPassName()
+                           << " with error: " << err.what() << ENDM;
+    }
+  }
+}
+
+BookkeepingQualitySink::BookkeepingQualitySink(const std::string& grpcUri, Provenance type, SendCallback sendCallback)
```
knopers8, line 75:
```suggestion
BookkeepingQualitySink::BookkeepingQualitySink(const std::string& grpcUri, Provenance provenance, SendCallback sendCallback)
```
please update the var name

Framework/src/BookkeepingQualitySink.cxx
```diff
@@ -0,0 +1,139 @@
...
+  : mGrpcUri{ grpcUri }, mType{ type }, mSendCallback{ sendCallback } {}
```
knopers8, line 76:
```suggestion
  : mGrpcUri{ grpcUri }, mProvenance{ provenance }, mSendCallback{ sendCallback } {}
```

Framework/src/BookkeepingQualitySink.cxx
```diff
@@ -0,0 +1,139 @@
...
+
+auto merge(std::unique_ptr<QualityControlFlagCollection>&& collection, const std::unique_ptr<QualityObject>& qualityObject) -> std::unique_ptr<QualityControlFlagCollection>
+{
+  QualitiesToFlagCollectionConverter converter(std::move(collection), qualityObject->getPath());
+  converter(*qualityObject);
+  return converter.getResult();
+}
+
+auto collectionFromQualityObject(const QualityObject& qualityObject) -> std::unique_ptr<QualityControlFlagCollection>
```
knopers8, line 85: Given it's name, it's slightly unclear to me that this helper creates the collection without putting the QO itself in it.

Perhaps it could be renamed to `collectionForQualityObject`?

Framework/src/BookkeepingQualitySink.cxx
```diff
@@ -0,0 +1,139 @@
...
+  key.reserve(qualityObject.getDetectorName().size() + std::size("||") + qualityObject.getCheckName().size());
+  key.append(qualityObject.getDetectorName());
+  key.append("||");
+  key.append(qualityObject.getCheckName());
+  return key;
+}
+
+template <typename t>
+struct tt;
+
```
knopers8, line 109: leftover?

Framework/src/BookkeepingQualitySink.cxx
```diff
@@ -0,0 +1,139 @@
...
+void BookkeepingQualitySink::run(framework::ProcessingContext& context)
+{
+  for (auto const& ref : framework::InputRecordWalker(context.inputs())) {
+    try {
+      auto qualityObject = framework::DataRefUtils::as<QualityObject>(ref);
+      auto [emplacedIt, _] = mQualityObjectsMap.emplace(to_key(*qualityObject), collectionFromQualityObject(*qualityObject));
+      emplacedIt->second = merge(std::move(emplacedIt->second), qualityObject);
```
knopers8, line 116: Please store and send to BKP one FlagCollection per detector, not one per QO. Indeed it must have been lost in the requirements I passed to you - we agreed some time ago with Martin to have one upload call per detector, to minimize the overhead on BKP.

Framework/src/InfrastructureGenerator.cxx
```diff
@@ -800,4 +808,48 @@ void InfrastructureGenerator::generatePostProcessing(WorkflowSpec& workflow, con
...
 }
 
+template <typename Type>
+auto createSinkInput(const std::string& detectorName, const std::string& name) -> framework::InputSpec
+{
+  const auto outputSpec = Type::createOutputSpec(detectorName, name);
+  auto input = DataSpecUtils::matchingInput(outputSpec);
+  input.binding = name;
+  return input;
+}
```
knopers8, line 818: cool, this should also lead us to some unification of the different Data Processors we have in QC.

Framework/src/InfrastructureGenerator.cxx
```diff
@@ -800,4 +808,48 @@ void InfrastructureGenerator::generatePostProcessing(WorkflowSpec& workflow, con
...
+
+  for (const auto& aggregatorSpec : infrastructureSpec.aggregators) {
+    if (aggregatorSpec.active && aggregatorSpec.exportToBookkeeping) {
+      ILOG(Debug, Support) << "Adding input to BookkeepingSink from aggregator " << aggregatorSpec.aggregatorName << " and detector: " << aggregatorSpec.detectorName << ENDM;
+      sinkInputs.emplace_back(createSinkInput<Aggregator>(aggregatorSpec.detectorName, aggregatorSpec.aggregatorName));
+    }
+  }
+
+  if (sinkInputs.empty()) {
+    ILOG(Debug, Support) << "BookkeepingSink is not being created because there any suitable inputs with quality objects were not found" << ENDM;
```
knopers8, line 839: I don't quite understand the part `because there any suitable input`, could you rephrase please?

Framework/test/testBookkeepingQualitySink.cxx
```diff
@@ -0,0 +1,101 @@
...
+void customize(std::vector<CompletionPolicy>& policies)
+{
+  quality_control::customizeInfrastructure(policies);
+}
+
+#include <Framework/runDataProcessing.h>
+#include <Framework/ControlService.h>
+#include <Configuration/ConfigurationFactory.h>
+#include <Configuration/ConfigurationInterface.h>
+#include <TH1F.h>
```
knopers8, line 35: Please review the header inclusion, `TH1F` is surely not used.

knopers8 (changes requested):
Please extend `testInfrastructureGenerator` to see if `BookkeepingQualitySink` is created with expected Inputs in the expected workflow types.

knopers8:
Also, please put name of the ticket or some description of what is done in the name of the PR.

Framework/src/BookkeepingQualitySink.cxx
justonedev1: I don't really understand how to do that. The BKP API has passname/productionname and detector as parameters. So I cannot send everything per one detector. only I need to separate by these two parameters. Which is also the reason why I am using the `unordered_map`: to keep flag collections separated by their names and detectors. So this is already an optimization to reduce the number of objects sent into the BKP.

Framework/include/QualityControl/BookkeepingQualitySink.h
```diff
@@ -0,0 +1,56 @@
...
+#include "ProvenanceToQCRunType.h"
+
+namespace o2::quality_control::core
+{
+
+// This class gathers all QualityObjects from it's inputs, converting them to flags + sending them to grpc RCT/BKP when stop() is invoked
+class BookkeepingQualitySink : public framework::Task
+{
+ public:
+  using FlagsMap = std::unordered_map<std::string, std::unique_ptr<QualityControlFlagCollection>>;
```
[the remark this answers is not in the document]
justonedev1: reason why I am using `unordered_map` over `unordered_set` is quite simple: items in map can change, but they cannot change inside the set. So in order to transform code to use set over map, I would need to remove and add flag collection every time it has to be modified. Or create my own implementation. 

So it is good observation, that I basically do not use the key anywhere and it is even possible to change the code to use set, but I wanted to avoid big amount of insert and remove operations in the code.

Framework/src/BookkeepingQualitySink.cxx
justonedev1: Solved in person. The issue was my misunderstanding of the nomenclature. I will rewrite the code, so everything is grouped by detector only.

Framework/src/InfrastructureGenerator.cxx
```diff
@@ -800,4 +808,48 @@ void InfrastructureGenerator::generatePostProcessing(WorkflowSpec& workflow, con
...
+template <typename Type>
+auto createSinkInput(const std::string& detectorName, const std::string& name) -> framework::InputSpec
+{
+  const auto outputSpec = Type::createOutputSpec(detectorName, name);
+  auto input = DataSpecUtils::matchingInput(outputSpec);
+  input.binding = name;
+  return input;
+}
+
+void InfrastructureGenerator::generateBookkeepingQualitySink(WorkflowSpec& workflow, const InfrastructureSpec& infrastructureSpec)
```
knopers8, line 820: aside from that, you will also have to customize how this Data Processor treats inputs. By default, DPL will try to match the timeslice for all expected inputs, which makes sense for processing timeframe-based data, but since our messages are usually created based on timers, DPL will not be able to match them.

You can get inspired by `AggregatorRunnerFactory::customizeInfrastructure` and add basically the same thing for `BookkeepingQualitySink`. It will let it consume any incoming input immediately.

Framework/src/BookkeepingQualitySink.cxx
xsalonx: For synchronous QC there will be separate method, we will update proto file

justonedev1: okay, but we will go ahead with this ticket and when you have it, I will add the function call here.

knopers8 (approved):
Thanks, it seems good now!

In the separate PR, please update the documentation by adding the new option in:
- https://github.com/AliceO2Group/QualityControl/blob/master/doc/Advanced.md#qc-checks-configuration
- https://github.com/AliceO2Group/QualityControl/blob/master/doc/Advanced.md#qc-aggregators-configuration

...and by adding a new section in Advanced.md about the possibility to export flags to BKP.

knopers8:
the mac issue seems to be relevant:
```
/Volumes/build/alice-ci-workdir/qualitycontrol-o2/sw/SOURCES/QualityControl/2320/0/Framework/src/InfrastructureGenerator.cxx:851:17: error: no matching conversion for functional-style cast from 'const char[10]' to 'framework::DataProcessorLabel'
    .labels = { framework::DataProcessorLabel("resilient") }
```

PR #2321, opened 2024-06-01
MUON: use log binning for tracks-per-TF histogram x-axis

aferrero2707:
The log binning allows to draw the histogram with the x-axis set to log scale while keeping the bin width constant in log representation.


aferrero2707:
@knopers8 @Barthelemy would it be possible to have this PR included in the tag of tomorrow? The changes are quite small, and the code for making the log binning is already tested and used in another task in production. Thanks!

knopers8 (approved):

PR #2322, opened 2024-06-05
Common: properly handle Sumw2 in THxRatio classes

aferrero2707:
In the original implementation the creation of the sum-of-weights structure was forcibly enabled in the `init()` function, for both the ratio histogram and the internal numerato/denominator ones, without the possibility to disable it.
This doubles the size of the produced objects, because of the extra memory needed to store the sum-of-weights structures.

The new implementation allows to disable the sum-of-weights on a per-histogram basis, by calling `Sumw2(kFALSE)` immediately after the instantiation of the histogram ratio. This applies to both the ratio histogram and the internal numerator/denominator ones.

The sum-of-weights is still enabled by default in the `init()` function of the histogram ratio, to match the previous behavior and guarantee that errors are correctly computed and propagated at each step.

The relevant changes are the following:
- added a `Bool_t mSumw2Enabled` member variable to keep track of wether or not the sum-of-weights structure should be updated in the histogram computations
- overridden the `void Sumw2(Bool_t flag)` virtual function to propagate the flag to the internal numerator/denominator histograms
- modified the `update()` function to call `T::Scale(norm, "nosw2")` when the sum-of-weights structure is not enabled (otherwise the `Scale()` function automatically creates the structure regardless of the user settings, see [here](https://root.cern.ch/doc/master/classTH1.html#add929909dcb3745f6a52e9ae0860bfbd))

### DIFF
--- a/Modules/Common/include/Common/TH1Ratio.h
+++ b/Modules/Common/include/Common/TH1Ratio.h
@@ -70,11 +70,13 @@ class TH1Ratio : public T, public o2::mergers::MergeInterface
   Bool_t Add(const TH1* h1, const TH1* h2, Double_t c1 = 1, Double_t c2 = 1) override;
   Bool_t Add(const TH1* h1, Double_t c1 = 1) override;
   void SetBins(Int_t nx, Double_t xmin, Double_t xmax) override;
+  void Sumw2(Bool_t flag = kTRUE) override;
 
  private:
   T* mHistoNum{ nullptr };
   T* mHistoDen{ nullptr };
   bool mUniformScaling{ true };
+  Bool_t mSumw2Enabled{ kFALSE };
   std::string mTreatMeAs{ T::Class_Name() };
 
   ClassDefOverride(TH1Ratio, 1);
--- a/Modules/Common/include/Common/TH1Ratio.h
+++ b/Modules/Common/include/Common/TH1Ratio.h
@@ -76,7 +76,7 @@ class TH1Ratio : public T, public o2::mergers::MergeInterface
   T* mHistoNum{ nullptr };
   T* mHistoDen{ nullptr };
   bool mUniformScaling{ true };
-  Bool_t mSumw2Enabled{ kFALSE };
+  Bool_t mSumw2Enabled{ kTRUE };
   std::string mTreatMeAs{ T::Class_Name() };
 
   ClassDefOverride(TH1Ratio, 1);
--- a/Modules/Common/include/Common/TH2Ratio.h
+++ b/Modules/Common/include/Common/TH2Ratio.h
@@ -76,7 +76,7 @@ class TH2Ratio : public T, public o2::mergers::MergeInterface
   T* mHistoNum{ nullptr };
   T* mHistoDen{ nullptr };
   bool mUniformScaling{ true };
-  Bool_t mSumw2Enabled{ kFALSE };
+  Bool_t mSumw2Enabled{ kTRUE };
   std::string mTreatMeAs{ T::Class_Name() };
 
   ClassDefOverride(TH2Ratio, 1);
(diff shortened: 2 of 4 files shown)


aferrero2707:
@chiarazampolli @noferini @alcaliva @Barthelemy @knopers8 @JianLIUhep this PR will allow to reduce the size of the THxRatio plots, in all cases where accurate errors are not needed.

An example is given here: https://github.com/AliceO2Group/QualityControl/pull/2326

Some possible candidates:
- [ITSClusterTask.cxx](https://github.com/AliceO2Group/QualityControl/blob/master/Modules/ITS/src/ITSClusterTask.cxx)
- [ITSChipStatusTask.cxx](https://github.com/AliceO2Group/QualityControl/blob/master/Modules/ITS/src/ITSChipStatusTask.cxx)
- [QcMFTClusterTask.cxx](https://github.com/AliceO2Group/QualityControl/blob/master/Modules/MFT/src/QcMFTClusterTask.cxx)

Modules/Common/include/Common/TH2Ratio.h
```diff
@@ -70,11 +70,13 @@ class TH2Ratio : public T, public o2::mergers::MergeInterface
   Bool_t Add(const TH1* h1, const TH1* h2, Double_t c1 = 1, Double_t c2 = 1) override;
   Bool_t Add(const TH1* h1, Double_t c1 = 1) override;
   void SetBins(Int_t nx, Double_t xmin, Double_t xmax, Int_t ny, Double_t ymin, Double_t ymax) override;
+  void Sumw2(Bool_t flag = kTRUE) override;
```
knopers8, line 73:
```suggestion
  void SetSumw2(Bool_t flag = kTRUE) override;
```

Modules/Common/include/Common/TH1Ratio.h
```diff
@@ -70,11 +70,13 @@ class TH1Ratio : public T, public o2::mergers::MergeInterface
   Bool_t Add(const TH1* h1, const TH1* h2, Double_t c1 = 1, Double_t c2 = 1) override;
   Bool_t Add(const TH1* h1, Double_t c1 = 1) override;
   void SetBins(Int_t nx, Double_t xmin, Double_t xmax) override;
+  void Sumw2(Bool_t flag = kTRUE) override;
 
  private:
   T* mHistoNum{ nullptr };
   T* mHistoDen{ nullptr };
   bool mUniformScaling{ true };
+  Bool_t mSumw2Enabled{ kFALSE };
```
knopers8, line 79:
```suggestion
  Bool_t mSumw2Enabled{ kTRUE };
```
Should it be also `true` here for consistency with `init()`? Otherwise, someone may look at the header and assume that it is disabled by default, while it is anyway set to true in `init()`.

Modules/Common/include/Common/TH2Ratio.h
knopers8: For consistency with other setters, should we make is `SetSumw2`?

knopers8 (changes requested):
Thanks @aferrero2707 , we will all appreciate it greatly!

I have only minor corrections for consideration, once we agree on those, it's good with me.

Modules/Common/include/Common/TH2Ratio.h
aferrero2707: @knopers8 here I opted for overriding the virtual [TH1::Sumw2()](https://root.cern.ch/doc/master/classTH1.html#aefa4ee94f053ec3d217f3223b01fa014) function, since thre THxRatio classes are designed to behave as much as possible like standard histograms.

Modules/Common/include/Common/TH1Ratio.h
aferrero2707: Yes, good point, I'll push the change a bit later.

Modules/Common/include/Common/TH2Ratio.h
knopers8: ah, thanks, i didn't realise this is an override. all clear!

knopers8 (approved):

PR #2323, opened 2024-06-05
QC-1193 AggregatorRunner now sends quality objects

justonedev1:


Framework/src/AggregatorRunner.cxx
```diff
@@ -247,6 +260,14 @@ void AggregatorRunner::store(QualityObjectsType& qualityObjects)
   }
 }
 
+void AggregatorRunner::send(const core::QualityObjectsType& qualityObjects, framework::DataAllocator& allocator)
+{
+  ILOG(Debug, Devel) << "Sending" << qualityObjects.size() << " quality objects" << ENDM;
+  for (const auto& qualityObject : qualityObjects) {
+    allocator.snapshot(framework::Output{ createDataOrigin(), createAggregatorRunnerDataDescription(createAggregatorRunnerName()), 0 }, *qualityObject);
```
knopers8, line 267: I would propose to stay in line with how `QualityObjects` are published by `CheckRunner`. In particular, I mean that each `Aggregator` should have it's own output to send to. This will be consistent with `CheckRunner`, but most importantly, it will allow `BookkeepingQualitySink` to subscribe only to the inputs it needs.

knopers8 (changes requested):

Framework/src/AggregatorRunner.cxx
justonedev1: okay, will do

knopers8:
You can add [WIP] to the title of PR, to stop CI from running and make it clear that we should not consider merging it yet.

PR #2324, opened 2024-06-06
QC-1193: Aggregators are sendind quality objects

justonedev1:
I am sorry for closing the previous PR, I am not even sure how I did that. Maybe by reverting previous changes and pushing new ones.


knopers8 (approved):
Thanks, it seems good!

I am not very happy about keeping also `mAggregatorsMap`, but I do see the need for it. At some point we could see if this can be refactored into something more elegant.

PR #2325, opened 2024-06-07
CTP: small adjustment

lhusova:
add configurable database + fix message in check

### DIFF
--- a/Modules/CTP/src/RawDataQcTask.cxx
+++ b/Modules/CTP/src/RawDataQcTask.cxx
@@ -69,18 +69,18 @@ void CTPRawDataReaderTask::startOfActivity(const Activity& activity)
   mHistoClassRatios->Reset();
   mHistoInputRatios->Reset();
   mHistoMTVXBC->Reset();
-  //
+
   mRunNumber = activity.mId;
   mTimestamp = activity.mValidity.getMin();
-  //
-  // mTimestamp = 1714315086649;
-  //
+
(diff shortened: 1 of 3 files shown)


Modules/CTP/src/RawDataQcTask.cxx
```diff
@@ -69,18 +69,18 @@ void CTPRawDataReaderTask::startOfActivity(const Activity& activity)
...
+
   std::string MBclassName = mCustomParameters["MBclassName"];
   if (MBclassName.empty()) {
     MBclassName = "CMTVX-B-NOPF";
   }
   std::string run = std::to_string(mRunNumber);
-  o2::ctp::CTPRunManager::setCCDBHost("https://alice-ccdb.cern.ch");
+  std::string ccdbName = mCustomParameters["ccdbName"];
+
+  o2::ctp::CTPRunManager::setCCDBHost(ccdbName);
```
aferrero2707, line 83: I suggest to check if `ccdbName` is empty and use a default value (like `"https://alice-ccdb.cern.ch"`) in this case.
Like it is done for `MBclassName`

Barthelemy: I should probably provide a way to get it from the "official" item in the config.

Barthelemy: https://its.cern.ch/jira/browse/QC-1194

Barthelemy (approved):

PR #2326, opened 2024-06-08
MCH: disabled sum-of-weights for histogram ratios

aferrero2707:
The histogram ratio plots do not require accurate errors, disabling the sum-of-weights allows to reduce by a factor of two the size of the objects stored in the QCDB.

For proper size reduction it requires https://github.com/AliceO2Group/QualityControl/pull/2322


aferrero2707:
@chiarazampolli @noferini @alcaliva @Barthelemy @knopers8 @JianLIUhep this PR will reduce the output size of the MCH Digits and Preclusters tasks by about a factor of two.

As stated in the description, the reduction will be effective only when https://github.com/AliceO2Group/QualityControl/pull/2322 is also merged.

Barthelemy (approved):

PR #2327, opened 2024-06-10
added missing output to aggregator generator

justonedev1:
Issue QC-1193 was missing assigning outputs to Aggregators. This small PR is meant to fix that.

However there is one more issue. When @knopers8 was merging mentioned PR all tests were green, but when I run `o2-qc-run-producer | o2-qc --config json://.....basic-aggregator.json` it showed missing outputs for Aggregator (actor does not have privilege to write to XXX). This means that there is missing test that simply runs the workflow created by generator (eg. similar to one described `basic-aggregator.json`)

### DIFF
--- a/Framework/src/AggregatorRunnerFactory.cxx
+++ b/Framework/src/AggregatorRunnerFactory.cxx
@@ -43,7 +43,7 @@ DataProcessorSpec AggregatorRunnerFactory::create(const core::CommonSpec& common
   DataProcessorSpec newAggregatorRunner{
     aggregatorRunner.getDeviceName(),
     aggregatorRunner.getInputs(),
-    Outputs{},
+    aggregatorRunner.getOutputs(),
     AlgorithmSpec{},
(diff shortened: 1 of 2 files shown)


knopers8:
>  it showed missing outputs for Aggregator (actor does not have privilege to write to XXX)

Perhaps wrong bindings used when sending compared to the ones declared in `OutputSpecs` of the processor?

justonedev1:
No, the reason for the error was the simple missing line in `AggregatorRunnerFactory`.  I created all the outputSpecs in AggregatorRunner, but forgot to add them. And I also relied on tests, that everything is working. But when I simply run the testing workflow for task QC-978 I found this mistake

knopers8:
Ah, I understood there was still another problem even with this fix.

> And I also relied on tests, that everything is working.

Well, you added this feature, so it could not have been tested before :D

justonedev1:
Everything was cleared personally. We will extend testing suite

PR #2330, opened 2024-06-10
[EMCAL-529] Change NumPatchesPerFastORCheck to nsigma over mean checker

sweyh99:
Determine if a FastOR is noisy by how many sigma it is over the mean FastOR noise level 
Remove from the list of noisy FastORs any neighboring FastORs to the noisiest ones

### DIFF
--- a/Modules/EMCAL/src/NumPatchesPerFastORCheck.cxx
+++ b/Modules/EMCAL/src/NumPatchesPerFastORCheck.cxx
@@ -31,7 +31,6 @@
 #include <TRobustEstimator.h>
 #include <TMath.h>
 #include <ROOT/TSeq.hxx>
-#include <TSpectrum.h>
 #include <iostream>
 #include <vector>
 #include <sstream>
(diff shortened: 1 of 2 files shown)


knopers8:
@mfasDa this seems waiting for your review

Modules/EMCAL/src/NumPatchesPerFastORCheck.cxx
```diff
@@ -117,43 +112,121 @@ Quality NumPatchesPerFastORCheck::check(std::map<std::string, std::shared_ptr<Mo
...
-                if (mLogLevelIL > 2) {
-                  ILOG(Warning, Support) << errorMessage << ENDM;
-                }
-                mHighCountTRUPositions.insert(obj);
+            }
+          }
+
+          // Save the final bad FastORs
+          FastORNoiseLevel final{ i->mCounts, i->mFastORID, i->mPosGlobalPhi, i->mPosGlobalEta, i->mRejected };
+          finalBadFastORs.push_back(final);
```
mfasDa, line 173: `final` is a reserved keyword, please replace.

mfasDa (changes requested):
One small issue, otherwise it is in line with what we discussed and could be merged.

mfasDa (approved):
PR is approved, the overwriting of the keyword should be done in a separate commit.

PR #2331, opened 2024-06-11
test if aggregator outputs are stored in QCDB

knopers8:


knopers8:
if we add the Aggregator to functional test (https://github.com/AliceO2Group/QualityControl/pull/2329), let's also test if the output appears in QCDB

Barthelemy (approved):
thank you

PR #2333, opened 2024-06-11
TPC: Reductor for TimeGainCalib

mlesch:


Modules/TPC/src/TimeGainCalibReductor.cxx
```diff
@@ -0,0 +1,101 @@
...
+          mStats.meanEntries[iCharge][iStack] = 0.;
+        }
+        if (counter[iStack] > 1) {
+          mStats.stddevEntries[iCharge][iStack] = (sumSquare[iStack] - (sum[iStack] * sum[iStack]) / counter[iStack]) / (counter[iStack] - 1.);
+        } else {
+          mStats.stddevEntries[iCharge][iStack] = 0.;
+        }
+      }
+
+      double tgl = 1.;
```
wiechula, line 84:
```suggestion
      const double tgl = 1.;
```

wiechula (changes requested):
Look good. One non-critical thing.

Modules/TPC/src/TimeGainCalibReductor.cxx
mlesch: Done, thanks!

wiechula (approved):

PR #2337, opened 2024-06-13
GLO: fix memory leak

f3sch:
Hi @knopers8, sorry for missing this initially. I am currently traveling, so I have not tested if this fixes the leak issue, in principle it should. Can do this properly next week.

### DIFF
--- a/Modules/GLO/src/ITSTPCMatchingTask.cxx
+++ b/Modules/GLO/src/ITSTPCMatchingTask.cxx
@@ -128,6 +128,7 @@ void ITSTPCMatchingTask::endOfCycle()
   if (common::getFromConfig(mCustomParameters, "isSync", false)) {
     { // Pt
       auto hEffPt = mMatchITSTPCQC.getFractionITSTPCmatch(globaltracking::MatchITSTPCQC::ITS);
+      mHEffPt.reset();
       mHEffPt.reset(dynamic_cast<TH1*>(hEffPt->GetPassedHistogram()->Clone("mFractionITSTPCmatch_ITS_Hist")));
       if (mHEffPt) {
         mHEffPt->Divide(hEffPt->GetPassedHistogram(), hEffPt->GetTotalHistogram(), 1.0, 1.0, "B");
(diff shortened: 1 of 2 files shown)


Modules/GLO/src/ITSTPCMatchingTask.cxx
```diff
@@ -132,33 +128,36 @@ void ITSTPCMatchingTask::endOfCycle()
   if (common::getFromConfig(mCustomParameters, "isSync", false)) {
     { // Pt
       auto hEffPt = mMatchITSTPCQC.getFractionITSTPCmatch(globaltracking::MatchITSTPCQC::ITS);
-      if (auto hEffPtHist = dynamic_cast<TH1*>(hEffPt->GetPassedHistogram()->Clone("mFractionITSTPCmatch_ITS_Hist")); hEffPtHist != nullptr) {
-        hEffPtHist->Divide(hEffPt->GetPassedHistogram(), hEffPt->GetTotalHistogram(), 1.0, 1.0, "B");
-        hEffPtHist->SetBit(TH1::EStatusBits::kNoStats);
-        getObjectsManager()->startPublishing(hEffPtHist);
-        getObjectsManager()->setDefaultDrawOptions(hEffPtHist->GetName(), "logx");
+      mHEffPt.reset(dynamic_cast<TH1*>(hEffPt->GetPassedHistogram()->Clone("mFractionITSTPCmatch_ITS_Hist")));
```
knopers8, line 131: Doing it this way, for a moment we will have two histograms there with the same name, one in the unique_ptr, one on rhs as an argument. With any other library it would be fine, but ROOT sometimes segfaults in such case. Please do it in two steps instead.
```suggestion
      mHEffPt.reset();
      mHEffPt.reset(dynamic_cast<TH1*>(hEffPt->GetPassedHistogram()->Clone("mFractionITSTPCmatch_ITS_Hist")));
```

f3sch: Thanks, yes delete is called last.

knopers8 (approved):
thank you!

PR #2342, opened 2024-06-14
[EMCAL-767, QC-1198] Fix memory leak for Start-Stop-Start

mfasDa:
Initialize the histograms only if they are nullptr in
initialize. In case of Start-Stop-Start this means
using the existing histograms.


mfasDa:
@knopers8 Please review!

knopers8 (approved):
Many thanks for quick reaction, seems good!

PR #2348, opened 2024-06-18
Common: bumped ClassDefOverride version of THxRatio objects

aferrero2707:
The bump is needed due to the addition of the `mSumw2Enabled` member.


knopers8 (approved):
Indeed, thanks!

PR #2354, opened 2024-06-21
Add info about templating config files

Barthelemy:


doc/Advanced.md
```diff
@@ -1161,9 +1168,49 @@ Make sure that you are able to run with the new workflow before actually templat
...
 ### Include a config file
 
-To include a config file (e.g. named `mch_digits`) add this line: `{% include "mch_digits" %}`
+To include a config file (e.g. named `mch_digits`) add this line : 
+```
+{% include "MCH/mch_digits" %}
+```
+The content of the file `mch_digits` is then copied into the config file. Thus make sure that you include all the commas and stuff. 
+
+There is an example in Staging : components/qc/ANY/any/tpc-pulser-calib-qcmn
```
knopers8, line 1177: If you document it as an example, perhaps it should be enforced with ansible that it is there?

knopers8 (approved):
You could consider mentioning that one can test the config file looks right by calling apricot through a browser, such as:
```
{{ aprcitot_endpoint }}/components/qc/ANY/any/tpc-pulser-calib-qcmn?process=true
```
Otherwise, understanding issues such as typos might be quite difficult, since QC will only report an error saying "could not parse JSON".

Barthelemy:
> You could consider mentioning that one can test the config file looks right by calling apricot through a browser, such as:
> 
> ```
> {{ aprcitot_endpoint }}/components/qc/ANY/any/tpc-pulser-calib-qcmn?process=true
> ```
> 
> Otherwise, understanding issues such as typos might be quite difficult, since QC will only report an error saying "could not parse JSON".

Excellent idea, I have added it !

doc/Advanced.md
Barthelemy: You are correct but then we need a simpler example because I don't want to copy all the TPC files (in the subfolder).

doc/Advanced.md
```diff
@@ -1212,6 +1212,11 @@ or like this respectively:
 {% endif %}
 ```
 
+### Test and debug 
+
+To see how a config file will look like once templated, simply open a browser at this address: `{{apricot_endpoint}}:32188/components/qc/ANY/any/tpc-pulser-calib-qcmn?process=true`
+Replace `{{apricot_endpoint}}` by the value you can find in Consul under `o2/runtime/aliecs/vars/apricot_endpoint` (it is different on staging and prod). 
+
```
knopers8: probably everyone will realise, but what you will get by replacing `apricot_endpoint` will be:
```
myhost.cern.ch:32188:32188/components/qc/ANY/any/tpc-pulser-calib-qcmn?process=true
```
notice the double port

PR #2356, opened 2024-06-25
QC-271: Implement SOR/EOR triggers in post-processing

justonedev1:
For now this is just a draft to show the mechanism, that I created for propagating SOR (EOR to be implemented). Plus I did not make a PR to alidist with moderncppkafka, so this will not even build.

edit: this message is not valid anymore


Framework/src/Triggers.cxx
```diff
@@ -119,9 +120,19 @@ TriggerFcn Never(const Activity& activity)
   };
 }
 
-TriggerFcn EndOfRun(const Activity&)
+TriggerFcn EndOfRun(const std::string& kafkaBrokers, const std::string& topic, const Activity& activity)
 {
-  return NotImplemented("EndOfRun");
+  auto returnedActivity = activity;
+  auto poller = std::make_shared<core::KafkaPoller>(kafkaBrokers, "EOR_postprocessing_group");
```
knopers8, line 126: Forgive my lack of familiarity with kafka here...
What are the implications of using a common group ID for all postprocessing tasks who want to trigger on EOR? Does it mean that just one will get the event?

Framework/src/Triggers.cxx
```diff
@@ -119,9 +120,19 @@ TriggerFcn Never(const Activity& activity)
...
+  poller->subscribe(topic);
+  return [poller, returnedActivity]() mutable -> Trigger {
+    for (const auto& record : poller->poll()) {
+      if (proto_parser::isEOR(record.value())) {
+        return { TriggerType::StartOfRun, true, returnedActivity, static_cast<uint64_t>(record.timestamp().msSinceEpoch), "eor" };
```
knopers8, line 131:
```suggestion
        return { TriggerType::EndOfRun, false, returnedActivity, static_cast<uint64_t>(record.timestamp().msSinceEpoch), "eor" };
```
You used `TriggerType::StartOfRun` by accident here.
Otherwise, the comments below apply to both:
- I would use `last : false`, since it may not be the last time this trigger fires. If `last` is `true`, `tryTrigger` removes the trigger from the list of currently used triggers, which is probably not what we want.
- please fill `Activity` provided to the `Trigger` with the best knowledge of the run that starts or ends.
- please use  the SOSOR and SOEOR times as `timestamp` in the two triggers. To be precise, I mean the same values that ECS calls `"run_start_time_ms"` and `"run_end_time_ms"`. Is `record.timestamp()` exactly that?

Framework/test/testKafkaPoller.cxx
```diff
@@ -22,13 +22,32 @@
 TEST_CASE("test_kafka_poller")
```
knopers8, line 22: indeed we would need a kafka broker to run this test automatically... should we mark it as manual in CMake meanwhile?

Framework/src/Triggers.cxx
knopers8: Having a quick look at the ECS, I assume it's the good timestamp: https://github.com/AliceO2Group/Control/blob/1e83c517a22e5c4d27199f8dac9301e6d18a97cd/core/environment/environment.go#L205

knopers8: hmm, i'm not sure anymore. I think `record.timestamp()` is different than `event.Timestamp`, while the timestamp from the link mentioned earlier goes to the latter.

justonedev1: I am sorry I overlooked the event.timestamp. My mistake

Framework/src/KafkaPoller.cxx
```diff
@@ -0,0 +1,166 @@
...
+  for (size_t retryNumber = 0; retryNumber != numberOfRetries; ++retryNumber) {
+    try {
+      mConsumer.subscribe({ topic });
+      return;
+    } catch (const kafka::KafkaException& ex) {
+      // it sometimes happen that subscibe timeouts but another retry succeeds
+      if (ex.error().value() != RD_KAFKA_RESP_ERR__TIMED_OUT) {
+        throw;
+      } else {
+        ILOG(Warning, Ops) << "Failed to subscribe to kafka due to timeout " << retryNumber + 1 << "/" << numberOfRetries + 1 << " times, retrying...";
```
knopers8, line 148:
```suggestion
        ILOG(Warning, Ops) << "Failed to subscribe to kafka due to timeout " << retryNumber + 1 << "/" << numberOfRetries + 1 << " times, retrying..." << ENDM;
```

CMakeLists.txt
```diff
@@ -13,7 +13,7 @@ include(AddQCWorkflow)
 # ---- Project ----
```
knopers8, line 13: this seems like a weird git artifact, can you rebase with master or avoid modifying this file in any other way?

knopers8 (commented):
thanks! small technical issues to be corrected and we are good to go

justonedev1:
@ktf can you take a look on why, some of these builds are failing? I don't understand why every build is missing different symbols. It seems that only this branch is failing, so it might signal that librdkafka recipe is not correct, but I don't really understand why would `libssh.so` na libO2CCDB.so complain, when I don't think that I touched these. Thank you

justonedev1:
@ktf Can I get some pointers, please?

ktf:
I was on holidays last week and I am still not at 100%, sorry for the delayed reply.

Apparently the moderncppkafka is linking against the system libssh (why?) which uses a different openssl. Maybe the best solution is to actually check the recipe again and remove the dependency?

ktf:
I gave a look at the librdkafka.sh recipe. You need to either disable SASL / OpenSSL support, or add the needed `requires` for them.

justonedev1:
Thank you for the answer and sorry that I was pinging you multiple times when you were at the vacation.  Do you know how to test whether my changes to recipes worked without access to build machines?

justonedev1:
Thank you! I am able to reproduce it with you recommended cli arguments.

Framework/src/KafkaPoller.cxx
```diff
@@ -0,0 +1,166 @@
...
+void fillActivityWithoutTimestamp(const events::Event& event, Activity& activity)
+{
+  if (event.has_runevent()) {
+    auto& runEvent = event.runevent();
+
+    if (!isRunNumberSet(activity.mId)) {
+      activity.mId = runEvent.runnumber();
+    }
+
+    if (!isEnvironmentIdSet(activity.mProvenance)) {
```
knopers8, line 74: I think you mistook provenance with partition ID (which is another, somewhat obsolete name for environment ID). Provenance says whether the QC objects were created in sync, async or monte carlo QC.

Framework/src/Triggers.cxx
```diff
@@ -109,9 +115,23 @@ TriggerFcn Never(const Activity& activity)
...
-  return NotImplemented("EndOfRun");
+  auto copiedActivity = activity;
+  auto poller = std::make_shared<core::KafkaPoller>(kafkaBrokers, createKafkaGroupId("EOR_postprocessing", detector, taskName));
+  poller->subscribe(topic);
+  return [poller, copiedActivity]() mutable -> Trigger {
+    for (const auto& record : poller->poll()) {
+      if (auto event = proto::recordToEvent(record.value())) {
+        if (proto::end_of_run::check(*event, copiedActivity.mProvenance, copiedActivity.mId)) {
+          auto newActivityForTrigger = copiedActivity;
+          proto::end_of_run::fillActivity(*event, newActivityForTrigger);
```
knopers8, line 128: I don't think we should take any fields from the Activity of the postprocessing task. After all, we want to know about other ECS environment starting/finishing and any metadata we can extract about it.

Framework/src/KafkaPoller.cxx
```diff
@@ -0,0 +1,166 @@
...
+}
+
+} // namespace proto
+
+kafka::Properties createProperties(const std::string& brokers, const std::string& groupId)
+{
+  return { { { "bootstrap.servers", { brokers } },
+             { "group.id", { groupId } },
+             { "enable.auto.commit", { "true" } },
+             { "auto.offset.reset", { "latest" } } } };
```
knopers8, line 129: does `auto.offset.reset` == `latest` mean that we will consume only the latest event of a kind? What if two SORs happen in between two consecutive postprocessing cycles (two `tryTrigger` calls)?

Framework/test/testKafkaTests.cxx
```diff
@@ -0,0 +1,484 @@
...
+#include "kafka/ProducerRecord.h"
+#include "kafka/Types.h"
+
+#include <QualityControl/KafkaPoller.h>
+#include <QualityControl/Triggers.h>
+#include <kafka/KafkaProducer.h>
+#include <proto/events.pb.h>
+
+// change this url if you are testing on different kafka cluster
+constexpr auto kafkaCluster = "mtichak-flp-1-27.cern.ch:9092";
```
knopers8, line 42: perhaps put there some invalid value, you will avoid someone else testing on your instance by accident and making you wonder what happened. Also, it is somewhat more secure to avoid exposing internal endpoints.

knopers8 (changes requested):
Please update the documentation at: https://github.com/AliceO2Group/QualityControl/blob/master/doc/PostProcessing.md#post-processing-interface . Please specify there what we consider as SOR and EOR (among SOSOR, EOSOR, SOEOR, EOEOR).

Framework/src/KafkaPoller.cxx
```diff
@@ -0,0 +1,166 @@
...
+
+  throw std::runtime_error(std::string{ "Kafka Poller failed to subscribe after " }.append(std::to_string(numberOfRetries)).append(" retries"));
+}
+
+auto KafkaPoller::poll() -> KafkaRecords
+{
+  const auto records = mConsumer.poll(std::chrono::milliseconds{ 100 });
+  auto filtered = records | std::ranges::views::filter([](const auto& record) { return !record.error(); });
+  KafkaRecords result{};
+  result.reserve(records.size());
```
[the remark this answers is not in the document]
knopers8: that is unless a `view` does not provide you `size()`...

Framework/include/QualityControl/KafkaPoller.h
```diff
@@ -0,0 +1,61 @@
...
+
+namespace proto
+{
+
+auto recordToEvent(const kafka::Value&) -> std::optional<events::Event>;
+
+namespace start_of_run
+{
+void fillActivity(const events::Event& event, Activity& activity);
+bool check(const events::Event& event, const std::string& environmentID = "", int runNumber = 0);
```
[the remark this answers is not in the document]
justonedev1: I will use `isValid` as I agree that it sounds a bit clearer. But the purpose of namespaces is to distinguish which type of event is valid, eg. `start_of_run::isValid` , so it is overkill to repeat the name in the function.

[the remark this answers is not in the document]
knopers8: agreed

Framework/src/KafkaPoller.cxx
justonedev1: it means, that we when new consumer with new group id is connected to cluster it will consume only latest messages, if some consumer connects to already existing group, than it will resume from the offset in the group.

knopers8: ah, all good then, I think this is what we need

Framework/src/Triggers.cxx
```diff
@@ -57,28 +58,33 @@ TriggerFcn NotImplemented(std::string triggerName)
...
+
+TriggerFcn StartOfRun(const std::string& kafkaBrokers, const std::string& topic, const std::string& detector, const std::string& taskName, const core::Activity& activity)
+{
+  auto copiedActivity = activity;
+  auto poller = std::make_shared<core::KafkaPoller>(kafkaBrokers, createKafkaGroupId("SOR_postprocessing", detector, taskName));
+  poller->subscribe(topic);
+  return [poller, copiedActivity]() mutable -> Trigger {
+    for (const auto& record : poller->poll()) {
+      if (auto event = proto::recordToEvent(record.value())) {
+        if (proto::start_of_run::check(*event, copiedActivity.mProvenance, copiedActivity.mId)) {
```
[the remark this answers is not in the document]
justonedev1: How I understood the requirements: if Provenance/ID is filled in config activity, than match against it otherwise don't.

Framework/src/Triggers.cxx
justonedev1: Right now it is setup that if Activity has empty fields, it is not filled and trigger matches against everything. And if you want to provide some filter over runnumber/envid, than it is possible.

knopers8: uhm, but `Activity` received by argument of this function is set as the current Activity (i.e. current run, current env ID if it's ran with ECS), so user won't always have control over that. I think we can safely assume that the user should get any available SOR/EOR except of their own.

justonedev1: How? When I look into triggerFactory, than this activity is from config

knopers8: And this is the config that is passed: https://github.com/AliceO2Group/QualityControl/blob/f6820b593ec6c69eff455bb2151d19501f65ec79/Framework/src/PostProcessingRunner.cxx#L196

You can see that the `activityFromDriver` is used.

justonedev1: or is the config always templated to the current activity? And if so, than what should I do, just invert the filtering? return anything except id if current run?

knopers8: yes, return a valid trigger if any other run starts or ends, but not this one.

Framework/CMakeLists.txt
```diff
@@ -100,14 +118,19 @@ add_library(O2QualityControl
...
 target_include_directories(
   O2QualityControl
   PUBLIC $<INSTALL_INTERFACE:include>
   $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
+  ${MODERNCPPKAFKA_ROOT}
+  $<BUILD_INTERFACE:${RDFKAFKA_INCLUDE}>
   PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/src
-  $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}/include>)
+  $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}/include>
+  ${MODERNCPPKAFKA_ROOT}
```
Barthelemy, line 132: It seems that you have twice `${MODERNCPPKAFKA_ROOT}`

Barthelemy (approved):
For the cmake part it is ok for me. Just one small comment but I guess it is just a typo.

Framework/CMakeLists.txt
justonedev1: you are correct, thanks

Framework/src/KafkaPoller.cxx
```diff
@@ -0,0 +1,166 @@
...
+      }
+    }
+  }
+
+  throw std::runtime_error(std::string{ "Kafka Poller failed to subscribe after " }.append(std::to_string(numberOfRetries)).append(" retries"));
+}
+
+auto KafkaPoller::poll() -> KafkaRecords
+{
+  const auto records = mConsumer.poll(std::chrono::milliseconds{ 100 });
```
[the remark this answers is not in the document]
justonedev1: it is not waiting, it is just timeout for kafka operations. `poll` is function from `moderncppkafka` that just tries to get all available messages.

Framework/src/KafkaPoller.cxx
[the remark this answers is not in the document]
justonedev1: it does not

justonedev1:
I will fill in the requested information, when I get it from Teo

Framework/src/KafkaPoller.cxx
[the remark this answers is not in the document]
knopers8: Let me understand then. Is it a timeout for how long the function should take, assuming there is data at the waiting to be received (as I understand you now) or is it a timeout to wait if there is no data (as I understood before)?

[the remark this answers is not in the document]
justonedev1: I am sorry... I misremembered what timeout means, I thought that it is for network only stuff, but that settings is in global config. You are right that this timeout is used to wait for new messages. [see](https://docs.confluent.io/platform/current/clients/librdkafka/html/rdkafka_8h.html#a53511739a2cf498b8d88287fef6873ce) ... So in our case it is fine to use 0

justonedev1:
So in kafka topic that is being consumed we are getting SOSOR, EOSOR, SOEOR, EOEOR, which one should I use? EOSOR and EOEOR?

knopers8:
I would say we should use whatever ECS considers as SOR and EOR, which are SOSOR and SOEOR correspondingly as far as i remember. We are not bound by any requirements yet, so I think this is a good start and we can consider changing it once we have some feedback.

Framework/src/KafkaPoller.cxx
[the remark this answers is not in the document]
justonedev1: After putting 0 as timeout, I had problems with the test, so I used 10ms instead, which works well with the tests and it seems to me, that it should be fine to wait 10ms every x minutes...

knopers8 (approved):
thank you!

Framework/src/KafkaPoller.cxx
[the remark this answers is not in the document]
knopers8: sounds good

PR #2358, opened 2024-06-26
[QC-885] Modify reference check to also accept a TH1 …

Barthelemy:
… instead of a Canvas

- also pass the database object from CheckRunner to CheckInterface and use it to implement a new method `retrieveReference`.

### DIFF
--- a/Framework/include/QualityControl/CheckInterface.h
+++ b/Framework/include/QualityControl/CheckInterface.h
@@ -94,7 +94,12 @@ class CheckInterface : public core::UserCodeInterface
 
   /// \brief Retrieve a reference plot at the provided path, matching the give activity and for the provided run.
   /// the activity is the current one, while the run number is the reference run.
-  std::shared_ptr<MonitorObject> retrieveReference(std::string path, int referenceRun, Activity activity);
+  ///
+  /// \param path path to the object (no provenance)
+  /// \param referenceRun Run number of the reference data
+  /// \param activity Current activity (necessary for the provenance and the pass)
+  /// \return
+  std::shared_ptr<MonitorObject> retrieveReference(std::string path, size_t referenceRun, Activity activity);
 
  private:
   std::shared_ptr<o2::quality_control::repository::DatabaseInterface> mDatabase;
--- a/Framework/include/QualityControl/CheckInterface.h
+++ b/Framework/include/QualityControl/CheckInterface.h
@@ -83,7 +83,8 @@ class CheckInterface : public core::UserCodeInterface
   virtual void startOfActivity(const core::Activity& activity); // not fully abstract because we don't want to change all the existing subclasses
   virtual void endOfActivity(const core::Activity& activity);   // not fully abstract because we don't want to change all the existing subclasses
 
-  void setDatabase(std::shared_ptr<o2::quality_control::repository::DatabaseInterface> database) {
+  void setDatabase(std::shared_ptr<o2::quality_control::repository::DatabaseInterface> database)
+  {
     mDatabase = database;
   }
 
--- a/Framework/include/QualityControl/CheckInterface.h
+++ b/Framework/include/QualityControl/CheckInterface.h
@@ -96,10 +96,9 @@ class CheckInterface : public core::UserCodeInterface
   /// the activity is the current one, while the run number is the reference run.
   ///
   /// \param path path to the object (no provenance)
-  /// \param referenceRun Run number of the reference data
-  /// \param activity Current activity (necessary for the provenance and the pass)
+  /// \param referenceActivity Reference activity (usually a copy of the current activity with a different run number)
   /// \return
-  std::shared_ptr<MonitorObject> retrieveReference(std::string path, size_t referenceRun, Activity activity);
+  std::shared_ptr<MonitorObject> retrieveReference(std::string path, Activity referenceActivity);
 
  private:
   std::shared_ptr<o2::quality_control::repository::DatabaseInterface> mDatabase;
--- a/Framework/include/QualityControl/CheckInterface.h
+++ b/Framework/include/QualityControl/CheckInterface.h
@@ -17,6 +17,7 @@
 #ifndef QC_CHECKER_CHECKINTERFACE_H
 #define QC_CHECKER_CHECKINTERFACE_H
 
+#include "QualityControl/DatabaseInterface.h"
 #include "QualityControl/Quality.h"
 #include "QualityControl/UserCodeInterface.h"
 #include "QualityControl/Activity.h"
@@ -82,9 +83,20 @@ class CheckInterface : public core::UserCodeInterface
   virtual void startOfActivity(const core::Activity& activity); // not fully abstract because we don't want to change all the existing subclasses
   virtual void endOfActivity(const core::Activity& activity);   // not fully abstract because we don't want to change all the existing subclasses
 
+  void setDatabase(std::shared_ptr<o2::quality_control::repository::DatabaseInterface> database) {
+    mDatabase = database;
+  }
+
  protected:
   /// \brief Called each time mCustomParameters is updated.
   virtual void configure() override;
+
+  /// \brief Retrieve a reference plot at the provided path, matching the give activity and for the provided run.
+  /// the activity is the current one, while the run number is the reference run.
+  std::shared_ptr<MonitorObject> retrieveReference(std::string path, int referenceRun, Activity activity);
+
+ private:
+  std::shared_ptr<o2::quality_control::repository::DatabaseInterface> mDatabase;
 
   ClassDef(CheckInterface, 6)
 };
--- a/Framework/include/QualityControl/ReferenceUtils.h
+++ b/Framework/include/QualityControl/ReferenceUtils.h
@@ -1,0 +1,61 @@
+// Copyright 2019-2020 CERN and copyright holders of ALICE O2.
+// See https://alice-o2.web.cern.ch/copyright for details of the copyright holders.
+// All rights not expressly granted are reserved.
+//
+// This software is distributed under the terms of the GNU General Public
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+///
+/// \file   ReferenceComparatorUtils.h
+/// \author Andrea Ferrero and Barthelemy von Haller
+///
+
+#ifndef QUALITYCONTROL_ReferenceComparatorUtils_H
+#define QUALITYCONTROL_ReferenceComparatorUtils_H
+
+#include <memory>
+#include "QualityControl/MonitorObject.h"
+#include "QualityControl/DatabaseInterface.h"
+#include "QualityControl/Activity.h"
+#include "QualityControl/ActivityHelpers.h"
+#include "QualityControl/QcInfoLogger.h"
+#include "QualityControl/RepoPathUtils.h"
+
+namespace o2::quality_control_modules::common
+{
+
+//_________________________________________________________________________________________
+//
+// Get the reference plot for a given MonitorObject path
+
+static std::shared_ptr<quality_control::core::MonitorObject> getReferencePlot(quality_control::repository::DatabaseInterface* qcdb, std::string& fullPath, uint32_t referenceRun, quality_control::core::Activity activity)
+{
+  uint64_t timeStamp = 0;
+  activity.mId = referenceRun;
+  const auto filterMetadata = quality_control::core::activity_helpers::asDatabaseMetadata(activity, false);
+  const auto objectValidity = qcdb->getLatestObjectValidity(activity.mProvenance + "/" + fullPath, filterMetadata);
+  if (objectValidity.isValid()) {
+    timeStamp = objectValidity.getMax() - 1;
+  } else {
+    ILOG(Warning, Devel) << "Could not find the object '" << fullPath << "' for run " << activity.mId << ENDM;
+    return nullptr;
+  }
+
+  std::string path;
+  std::string name;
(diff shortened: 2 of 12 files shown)


Barthelemy:
@aferrero2707

aferrero2707:
@Barthelemy shall I test it locally and report back?

Barthelemy:
@aferrero2707 good idea ! thanks
Also if you have any comment on the code as you wrote it

aferrero2707:
@Barthelemy I have loaclly compared the direct check you implemented with the check result from the ReferenceComparator output, and they perfectly match, so as far as I can tell the code is working as expected.

Here is a screenshot showing the result of the two check methods:
image

Barthelemy:
@aferrero2707 Thanks a lot for checking. I have resolved the conflict as well. 

@knopers8 I think that you can review it if you want

Framework/include/QualityControl/CheckInterface.h
```diff
@@ -82,10 +83,22 @@ class CheckInterface : public core::UserCodeInterface
...
+    mDatabase = database;
+  }
+
  protected:
   /// \brief Called each time mCustomParameters is updated.
   virtual void configure() override;
 
+  /// \brief Retrieve a reference plot at the provided path, matching the give activity and for the provided run.
+  /// the activity is the current one, while the run number is the reference run.
+  std::shared_ptr<MonitorObject> retrieveReference(std::string path, int referenceRun, Activity activity);
```
knopers8, line 97: If there is only one correct/expected Activity that the user shall provide and the framework has the knowledge of it, why even ask the user to do it? Shouldn't the framework handle it?

Also, since the run numbers are unique, why even constrain the selection with the `Activity`? I don't understand what's the benefit, except that someone would want to fail if the reference run is from an earlier period than the current one. However, for such cases we should anyway produce more specific errors than `could not find object`, so we should handle it explicitly.

Framework/include/QualityControl/ReferenceUtils.h
```diff
@@ -0,0 +1,58 @@
...
+
+#include <memory>
+#include "QualityControl/MonitorObject.h"
+#include "QualityControl/DatabaseInterface.h"
+#include "QualityControl/Activity.h"
+#include "QualityControl/ActivityHelpers.h"
+#include "QualityControl/QcInfoLogger.h"
+#include "QualityControl/RepoPathUtils.h"
+
+namespace o2::quality_control_modules::common
```
knopers8, line 28:
```suggestion
namespace o2::quality_control::checker
```
since it is in the framework, putting it in `quality_control_modules` is somewhat confusing.

Framework/include/QualityControl/ReferenceUtils.h
```diff
@@ -0,0 +1,58 @@
...
+    ILOG(Warning, Devel) << "Could not find the object '" << fullPath << "' for run " << activity.mId << ENDM;
+    return nullptr;
+  }
+
+  std::string path;
+  std::string name;
+  if (!o2::quality_control::core::RepoPathUtils::splitObjectPath(fullPath, path, name)) {
+    return nullptr;
+  }
+  return qcdb->retrieveMO(path, name, timeStamp, activity);
```
knopers8, line 53:
```suggestion
  return qcdb->retrieveMO(path, name, repository::DatabaseInterface::Timestamp::Latest, activity);
```
why not that instead?

Framework/include/QualityControl/RepoPathUtils.h
```diff
@@ -128,6 +128,19 @@ class RepoPathUtils
 
   static constexpr auto allowedProvenancesMessage = R"(Allowed provenances are "qc" (real data processed synchronously), "qc_async" (real data processed asynchronously) and "qc_mc" (simulated data).)";
   static bool isProvenanceAllowed(const std::string& provenance);
+
+  static bool splitObjectPath(const std::string& fullPath, std::string& path, std::string& name)
```
knopers8, line 132: could you document that the 2nd and the 3rd arguments are outputs?

Otherwise, one could modify the function to return a tuple with `path` and `name`.

Framework/include/QualityControl/ReferenceUtils.h
knopers8: but please add nullptr check for the QCDB and perhaps also complain if `referenceRun` is zero, which is likely a user error.

Modules/Common/src/ReferenceComparatorCheck.cxx
```diff
@@ -129,6 +133,32 @@ static Quality compare(TCanvas* canvas, ObjectComparatorInterface* comparator, s
   return comparator->compare(plots.first, plots.second, message);
 }
 
+Quality ReferenceComparatorCheck::getSinglePlotQuality(std::shared_ptr<MonitorObject> mo, std::string& message)
+{
+  // retrieve the reference plot and compare
+  auto* th1 = dynamic_cast<TH1*>(mo->getObject());
```
knopers8, line 139: please return Quality::Null if `th1` is nullptr.

Modules/Common/src/ReferenceComparatorCheck.cxx
```diff
@@ -129,6 +133,32 @@ static Quality compare(TCanvas* canvas, ObjectComparatorInterface* comparator, s
...
+  auto referenceRun = std::stoi(mCustomParameters.atOptional("referenceRun").value_or("0"));
```
knopers8, line 140: is default value "0" ever worth continuing with? Perhaps we should return `Quality::Null` instead?

Modules/Common/src/ReferenceComparatorCheck.cxx
```diff
@@ -129,6 +133,32 @@ static Quality compare(TCanvas* canvas, ObjectComparatorInterface* comparator, s
...
+
+  // get path of mo and ref (we have to remove the provenance)
+  string path = mo->getPath();
+  size_t pos = path.find('/');
+  if (pos != std::string::npos) {
+    path = path.substr(pos + 1);
+  }
```
knopers8, line 147: could be extracted to a helper.

Modules/Common/src/ReferenceComparatorCheck.cxx
```diff
@@ -129,6 +133,32 @@ static Quality compare(TCanvas* canvas, ObjectComparatorInterface* comparator, s
...
+
+  auto referencePlot = retrieveReference(path, referenceRun, mActivity);
+  if (!referencePlot) {
+    message = "The reference plot is empty";
```
knopers8, line 151: this message might be misleading. It could be understood as "the reference plot is there, but there are no entries". Please rephrase to convey the meaning "could not retrieve the reference Monitor Object".

Modules/Common/src/ReferenceComparatorCheck.cxx
```diff
@@ -139,17 +169,22 @@ Quality ReferenceComparatorCheck::check(std::map<std::string, std::shared_ptr<Mo
   Quality result = Quality::Good;
   for (auto& [key, mo] : *moMap) {
     auto moName = mo->getName();
+    Quality quality;
+    std::string message;
 
-    auto* canvas = dynamic_cast<TCanvas*>(mo->getObject());
-    if (!canvas) {
+    // run the comparison algorithm
+    if (mo->getObject()->IsA() == TClass::GetClass<TCanvas>()) {
```
knopers8, line 176: please add `#include <TClass.h>` if you are going for this approach.

Modules/Common/src/ReferenceComparatorCheck.cxx
```diff
@@ -139,17 +169,22 @@ Quality ReferenceComparatorCheck::check(std::map<std::string, std::shared_ptr<Mo
...
+      // We got a canvas. It contains the plot and its reference.
+      auto* canvas = dynamic_cast<TCanvas*>(mo->getObject());
+      quality = compare(canvas, mComparator.get(), message);
+    } else if (mo->getObject()->InheritsFrom(TH1::Class())) {
+      // We got a plot, we have to find the reference before calling the comparator
+      quality = getSinglePlotQuality(mo, message);
+    } else {
+      ILOG(Warning, Ops) << "mo is not a TCanvas or a TH1" << ENDM;
```
knopers8, line 184: If it's a message for an operator, it could be more helpful and precise perhaps?
```suggestion
      ILOG(Warning, Ops) << "Compared Monitor Object '" << mo->getName() << "` is not a TCanvas or a TH1, the detector QC responsible should review the configuration" << ENDM;
```

doc/Advanced.md
```diff
@@ -1398,6 +1398,75 @@ void customize(std::vector<ConfigParamSpec>& workflowOptions)
...
 
+## Reference data
+
+A reference object is an object from a previous run. It is usually used as a point of comparison. 
+
+### Get a reference plot in a check
+
+To retrieve a reference plot in your Check, use 
+```
+  std::shared_ptr<MonitorObject> retrieveReference(std::string path, int referenceRun, Activity activity);
```
knopers8, line 1409:
```suggestion
  std::shared_ptr<MonitorObject> CheckInterface::retrieveReference(std::string path, int referenceRun, Activity activity);
```
perhaps it should be clearer that it is a member of CheckInterface.

doc/Advanced.md
```diff
@@ -1398,6 +1398,75 @@ void customize(std::vector<ConfigParamSpec>& workflowOptions)
...
+          }
+        }
+      }
+```
+The check needs the following parameters
+- `referenceRun` to specify what is the run of reference and retrieve the reference data. 
+- `comparatorName` to decide how to compare, see below for their descriptions.
+- `threshold` to specifie the value used to discriminate between good and bad matches between the histograms.
+
+Three comparator are provided:
```
knopers8, line 1456:
```suggestion
Three comparators are provided:
```

Framework/src/CheckInterface.cxx
```diff
@@ -59,4 +60,9 @@ void CheckInterface::endOfActivity(const Activity& activity)
   // noop, override it if you want.
 }
 
+shared_ptr<MonitorObject> CheckInterface::retrieveReference(std::string path, int referenceRun, Activity activity)
```
knopers8, line 63: here reference run uses `int`, elsewhere `getReferencePlot` uses `uint32_t`. Should it be unified for better error handling (e.g. negative run numbers)?

knopers8 (changes requested):
Thank you! I have some suggestions to be discussed.

Framework/include/QualityControl/ReferenceUtils.h
Barthelemy: I must shamefully admit that I moved this piece of code from `ReferenceComparatorTask`. @aferrero2707 perhaps you want to comment on the approach ? 

@knopers8 I will add the check

Framework/include/QualityControl/RepoPathUtils.h
Barthelemy: This method must not be needed anymore if the change above is merged in

Framework/src/CheckInterface.cxx
Barthelemy: agreed

Framework/include/QualityControl/CheckInterface.h
Barthelemy: @aferrero2707 I picked these parameters based on your code. Do you have an opinion?  
I would tend to agree with Piotr.

Modules/Common/src/ReferenceComparatorCheck.cxx
Barthelemy: agreed

aferrero2707:
@Barthelemy @knopers8 I basically agree on all suggestion from Piotr. I have also added a comment d=regarding the way the reference run number is retrieved in the checker, suggesting to make use of the extended check parameters for that.

For the rest, looks good to me.

aferrero2707 (changes requested):

Framework/include/QualityControl/CheckInterface.h
Barthelemy: After discussing with Andrea as well, I came to the conclusion that we should keep the parameter `activity`, at least for the time being. Replacing it with `provenance` and `pass` opens the door to having more and more parameters that will match what the Activity contains in the end. Moreover, we cannot let the framework feed the activity because the CheckInterface does not store the activity. We could modify all the *Interface to start storing the activity but I believe that it would go beyond the scope of this PR. 

What I did was to document what the activity parameter is.

Framework/include/QualityControl/ReferenceUtils.h
Barthelemy: done

Modules/Common/src/ReferenceComparatorCheck.cxx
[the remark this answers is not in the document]
Barthelemy: done

Modules/Common/src/ReferenceComparatorCheck.cxx
```diff
@@ -143,7 +143,11 @@ Quality ReferenceComparatorCheck::getSinglePlotQuality(std::shared_ptr<MonitorOb
     message = "The MonitorObject is not a TH1";
     return Quality::Null;
   }
-  auto referenceRun = std::stoi(mCustomParameters.atOptional("referenceRun").value_or("0"));
+  if (mCustomParameters.count("referenceRun") == 0) {
+    message = "No reference run provided";
+    return Quality::Null;
+  }
+  auto referenceRun = std::stoi(mCustomParameters.at("referenceRun"));
```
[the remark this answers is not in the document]
Barthelemy: done. I actually changed all the calls to the custom parameters to pass the activity.

Barthelemy:
@knopers8 as discussed on Friday, I have replaced the `runNumber` + `Activity` with just `Activity` which should be the activity matching the reference plot. I agree with you that it makes more sense.

knopers8 (approved):

PR #2359, opened 2024-06-27
Common: use extended parameters in ReferenceComparatorTask

aferrero2707:
The access to the custom parameters is modified to make use of the extended version, such that the reference run number can be configured separately for different run and beam types.


aferrero2707:
@Barthelemy @knopers8 this is a small update of the ReferenceComparatorTask to allow using the `extendedTaskParameters`. This is actually quite needed to properly set reference run numbers according to different run types...

Barthelemy (approved):

PR #2362, opened 2024-06-28
CTP: changes requested by review

lhusova:
Dear @Barthelemy, dear @knopers8 , 

please have a look on the changes. We didn't touch the CountersTask yet, because it is not used currently and we need to develop the way how to monitor without run, as discussed in the emails, so of this we will create a separate PR later. For the ccdb readout, there was some discussion and it is ongoing, because there were some problems, but we will test also the recommended way. This is what we have now, I wanted to put it here as I'm leaving for holidays next week. 

Moreover, this PR contains also a commit changing the TH1F to TH1D, as spotted by Andrea, by integration mode, there are too many entries for TH1F.


knopers8 (approved):
thank you!

PR #2363, opened 2024-07-02
[EMCAL-767] added FEEDCS object histograms into the calibration task

sweyh99:
The histograms added are:

Link active: 1 Histogram with 46 bins
SRU Firmware version: 1 Histogram with 20 bins
TRU threshold: 1 histograms with 46 bins
L0 algorithm: 1 histogram with 46 bins
Rollback Buffer: 1 histogram with 46 bins
TRU mask: 2D histogram with 48 x 104 bins, showing the position of a mask.


mfasDa (approved):
Ready to merge, in a second iteration we also need the TRU grid to be drawn for the FastOR mask histogram.

PR #2364, opened 2024-07-03
QC-1202 Improve naming of canvas elements created by TrendingTask

knopers8:
As it turned out, some TPC Checks were relying on the default naming used by TTree::Draw when drawing the trends, i.e. "Graph" and "htemp". When implementing QC-1155, we could not keep them default to allow for drawing a legend in case we had multiple trends on one canvas.

This commit changes the following:
- It adds an optional "name" field to graphs. If it's absent, "title" is used. If that is absent as well, the main "name" of the plot is used.
- In case that "htemp" is used by TTree::Draw only to draw the axes, title, etc, it is renamed to "background". In case that the plot is 1D, "htemp" is renamed to the plot "name"
- In case that errors are drawn, the object is renamed to "name" suffixed with "_errors"

### DIFF
--- a/Framework/include/QualityControl/TrendingTaskConfig.h
+++ b/Framework/include/QualityControl/TrendingTaskConfig.h
@@ -33,6 +33,7 @@ struct TrendingTaskConfig : PostProcessingConfig {
 
   // this corresponds to one TTree::Draw() call, i.e. one graph or histogram drawing
   struct Graph {
+    std::string name;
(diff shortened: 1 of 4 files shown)


knopers8:
@mlesch is that OK with you?

mlesch:
@knopers8 perfect, thanks! I can then adapt the TPC Checkers for this case

PR #2368, opened 2024-07-04
[EMCAL-767] Added the TRU gridlines to the TRU Mask Position MO

sweyh99:


Barthelemy (approved):

mfasDa (changes requested):
I think you could modify the code in a way that no extra canvas is needed. It is done in a same way for the supermodule grid in the acceptance plots by the CellCheck. In order to do so you make use of 
```c++
histo->GetListOfFunctions()->Add(line);
```
for each line to add. In that case you pass a 2D histogram to the function.

Ideally, I would like to move the function -  as well as the code for the supermodule grid - to a helper class in order to avoid code duplication. We need this in more places, therefore it would be helpful to generalise.

Last, with such a restructuring I would suggest the following sequence:
- Supermodule grid
- TRU grid (1/3rd supermodue)
- FastOR grid
in separate functions. Like this we can combine as needed.

Modules/EMCAL/include/EMCAL/CalibMonitoringTask.h
```diff
@@ -94,7 +94,8 @@ class CalibMonitoringTask final : public quality_control::postprocessing::PostPr
   TH1* mTRUThresholds = nullptr;                      ///< The L0 threshold vs TRU ID PHYS
   TH1* mL0Algorithm = nullptr;                        ///< The L0 algorithm vs TRU ID
   TH1* mRollbackSTU = nullptr;                        ///< The Rollback buffer vs TRU ID
-  TH1* mTRUMaskPosition = nullptr;                    ///< The FastOR Mask Position in Eta, Phi
+  TCanvas* mTRUMaskPositionCanvas = nullptr;          ///< The FastOR Mask Position in Eta, Phi Canvas
```
mfasDa, line 97: There is still an unused TCanvas in

mfasDa (changes requested):
Only a minor change needed

mfasDa (approved):
Now ready to merge

PR #2369, opened 2024-07-09
ITS: NCluster plot moved to Double

IsakovAD:
Changed type of the nClustersPerTrack histogram from float to double. 

This change should fix problem for a-QC where we have overflow of float limits


IsakovAD:
Hello @iravasen 

Could you please approve this request? This is trivial solution to a-QC problem

iravasen (approved):

iravasen (changes requested):
@IsakovAD thanks, you need also to modify the checkers. Example, this line: https://github.com/AliceO2Group/QualityControl/blob/635ab2651c8a4f01b316aed4b5d02d31ac8dd414/Modules/ITS/src/ITSTrackCheck.cxx#L42

iravasen:
No wait sorry you did it! I did not see it. 

Then I think you can also modify the cluster task. In particular the 1D cluster size distributions.

IsakovAD:
Hello @iravasen 

I moved all the plots in Cluster/Track tasks to the Double. FEE task and Decoding error are already fine;

iravasen (approved):
Thanks a lot! Approving

PR #2371, opened 2024-07-11
[QC-1026] Separate source / target QCDB for postprocessing

Barthelemy:


Framework/src/PostProcessingRunner.cxx
```diff
@@ -82,16 +91,14 @@ void PostProcessingRunner::init(const PostProcessingRunnerConfig& runnerConfig,
...
+  mSourceDatabase = configureDatabase(mRunnerConfig.sourceDatabase, "Source");
+  mDestinationDatabase = configureDatabase(mRunnerConfig.destinationDatabase, "Destination");
 
   mObjectManager = std::make_shared<ObjectsManager>(mTaskConfig.taskName, mTaskConfig.className, mTaskConfig.detectorName, mRunnerConfig.consulUrl);
   mObjectManager->setActivity(mActivity);
-  mServices.registerService<DatabaseInterface>(mDatabase.get());
+  mServices.registerService<DatabaseInterface>(mSourceDatabase.get());
   if (mPublicationCallback == nullptr) {
-    mPublicationCallback = publishToRepository(*mDatabase);
+    mPublicationCallback = publishToRepository(*mDestinationDatabase);
```
knopers8, line 101: If done like this, it will only work in cases when a postprocessing task is ran with `o2-qc-run-postprocessing`, that is without DPL. When we run with DPL, this callback is set by `PostProcessingDevice` to publish objects into DPL, then CheckRunners receive them, apply Checks and store them to the database they use.

I think it will be very misleading if it works only for some cases.

How about getting rid of the `destinationRepo` parameter and using always the main repo as the destination?

Barthelemy: yes, we can do that for the destinatino repo. I did not think of this issue and thought that why not add both source and destination. I'll amend the PR.

Barthelemy: I have pushed a commit

knopers8 (approved):

PR #2374, opened 2024-07-23
ITS: Included new Reset plots to FEE Checks

IsakovAD:
Changed code for FEE Checks to include reset versions of TriggerVsFEEid and TrailerCount

### DIFF
--- a/Modules/ITS/itsDecoding.json
+++ b/Modules/ITS/itsDecoding.json
@@ -28,14 +28,14 @@
                 "className": "o2::quality_control_modules::its::ITSDecodingErrorTask",
                 "moduleName": "QcITS",
                 "detectorName": "ITS",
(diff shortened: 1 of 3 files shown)


Modules/ITS/itsDecoding.json
```diff
@@ -28,14 +28,14 @@
                 "className": "o2::quality_control_modules::its::ITSDecodingErrorTask",
                 "moduleName": "QcITS",
                 "detectorName": "ITS",
-                "cycleDurationSeconds": "30",
+                "cycleDurationSeconds": "10",
```
iravasen, line 31: Minor comment: if this is planned also for P2 json, one just need to check whether 10s is not too short. Thinking about possible backpressure issues.

iravasen (approved):

PR #2375, opened 2024-07-25
ITS: New shifter plot in Chip Status Task

IsakovAD:
- Added TH2Polly plot with the Chip Status information: rises BAD flag if all chips of stave areempty for N QC cycles
- Added QC check for this plot
- Added option to fill ChipStatus plot from right to left

### DIFF
--- a/Modules/ITS/CMakeLists.txt
+++ b/Modules/ITS/CMakeLists.txt
@@ -28,7 +28,7 @@ add_library(O2QcITS
             src/ITSDecodingErrorCheck.cxx 
             )
 
-target_sources(O2QcITS PRIVATE src/ITSChipStatusTask.cxx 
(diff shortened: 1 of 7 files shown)


iravasen (approved):

IsakovAD:
Hello @knopers8 @Barthelemy 

Could you please help us and check why this commit fails on macOS-arm?

Barthelemy:
@IsakovAD I am trying to compile

Barthelemy:
@IsakovAD I think that it is just exactly what the log says : 
```
/Volumes/build/alice-ci-workdir/qualitycontrol-o2/sw/SOURCES/QualityControl/2375/0/Modules/ITS/src/ITSChipStatusCheck.cxx:45:76: error: data argument not used by format string [-Werror,-Wformat-extra-args]
```
Have you tried removing the data argument ?

PR #2376, opened 2024-07-29
Common: improve reference run handling in checker

aferrero2707:
In the current code, a valid reference run is needed in all cases, even when only TCanvas objects were checked, which already have the reference plot embedded.

In order to avoid redundantly setting the reference run number both in the post-processing task and the checker when the `ReferenceComparatorTask` is used, the code is modified to only require a valid reference run number when TH1 objects are checked.

In addition, the code adds the caching of the reference plots, such that they are only loaded from QCDB once and re-used until a new activity is started.


aferrero2707:
@Barthelemy a couple of small improvements to the `ReferenceComparatorCheck`. Could you have a look and let me know if you agree? Thanks!

PR #2378, opened 2024-08-01
MUON: add plots with global kinematics

aferrero2707:
Some additional kinematics plots are introduced for eta, phi and pT distributions that are filled using the parameters of the global MFT+MCH(+MID) forward tracks.
The additional plots are only instantiated when MFT is included in the required detectors.
The already existing plots are now always filled using the MCH track parameters.

Five other plots have been disabled by default, to compensate the newly introduced plots (which are also five by default).


aferrero2707:
@Barthelemy @knopers8 this PR contains some changes needed to assess the efficiency of the matching between the muons and MFT. The net balance of added/removed plots is zero.

knopers8 (approved):

PR #2380, opened 2024-08-08
CcdbInspector: run number specified in CCDB queries

aferrero2707:
In the case of queries to the CCDB the metadata was completely empty, so the latest version of the queried object was retrieved regardless of the run number.
This leads to failures in retrieving the appropriate object in cases where the same type of object is updated by multiple parallel runs, as it is the case for example for the CTP configuration object. In such cases the object from the wrong run might be retrieved, if it was created shortly after the good one.

To avoid that this kind of corner cases leads to wrongly reported failures, the run number is now explicitly set in the metadata used for the CCDB queries, as it was already done for the QCDB ones.


aferrero2707:
@Barthelemy @knopers8 @justonedev1 would it be possible to have this PR reviewed and eventually merged in time for the next release? Thanks a lot!

Barthelemy (approved):

PR #2381, opened 2024-08-08
ReferenceComparator: allow ignoring period and pass when retrieving reference plots

aferrero2707:
The period and pass names were always included in the QCDB queries for retrieving the reference plots. This means that a run from a given period and pass could not be used as reference for a different period and/or pass, which is a strong limitation that needs to be made configurable.

The new code adds two configuration parameters that control whether the period and/or pass names should be ignored or not in the queries (see documentation for details).

### DIFF
--- a/Modules/Common/src/ReferenceComparatorCheck.cxx
+++ b/Modules/Common/src/ReferenceComparatorCheck.cxx
@@ -50,6 +50,8 @@ void ReferenceComparatorCheck::startOfActivity(const Activity& activity)
   auto comparatorName = mCustomParameters.atOptional("comparatorName", activity).value_or("");
   double threshold = std::stof(mCustomParameters.atOptional("threshold", activity).value_or("0"));
   mReferenceRun = std::stoi(mCustomParameters.atOptional("referenceRun", activity).value_or("0"));
+  mIgnorePeriodForReference = std::stoi(mCustomParameters.atOptional("ignorePeriodForReference", activity).value_or("1")) != 0;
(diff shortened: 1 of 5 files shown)


aferrero2707:
@Barthelemy @knopers8 @justonedev1 would it be possible to have this PR reviewed and eventually merged in time for the next release? Thanks a lot!

Modules/Common/src/ReferenceComparatorCheck.cxx
```diff
@@ -60,9 +62,14 @@ void ReferenceComparatorCheck::startOfActivity(const Activity& activity)
     mComparator->setThreshold(threshold);
   }
 
-  mActivity = activity;
```
Barthelemy, line 63: if `mActivity` is not used nor set I would simply remove it from this class

Barthelemy (commented):
Hi, 

It makes sense. I have only a small comment that I let you assess. 

Cheers

Modules/Common/src/ReferenceComparatorCheck.cxx
aferrero2707: @Barthelemy It is actually already removed, see https://github.com/AliceMCH/QualityControl/blob/4fd830b5a5e870d612ff2c42f0def61037bc9861/Modules/Common/include/Common/ReferenceComparatorCheck.h#L58

Or did I miss something? Thanks!

Barthelemy: sorry, indeed, I was checking the wrong branch when I saw it. 
All good !

Barthelemy (approved):

PR #2383, opened 2024-08-12
Common: add generic checker for trending plots

aferrero2707:
The checker is designed to verify that the values of trend graphs are within given minimum and maximum limits.
The limits can be specified as fixed values, or values relative to either the mean or the standard deviation of a given set of graph points.

See https://its.cern.ch/jira/browse/QC-1216 for more details.

### DIFF
--- a/Modules/Common/src/TrendCheck.cxx
+++ b/Modules/Common/src/TrendCheck.cxx
@@ -1,0 +1,466 @@
+// Copyright 2019-2020 CERN and copyright holders of ALICE O2.
+// See https://alice-o2.web.cern.ch/copyright for details of the copyright holders.
+// All rights not expressly granted are reserved.
+//
+// This software is distributed under the terms of the GNU General Public
+// License v3 (GPL Version 3), copied verbatim in the file "COPYING".
+//
+// In applying this license CERN does not waive the privileges and immunities
+// granted to it by virtue of its status as an Intergovernmental Organization
+// or submit itself to any jurisdiction.
+
+///
+/// \file   TrendCheck.cxx
+/// \author Andrea Ferrero
+///
+
+#include "Common/TrendCheck.h"
+#include "Common/CheckerThresholdsConfig.h"
+#include "QualityControl/MonitorObject.h"
+#include "QualityControl/Quality.h"
+#include "QualityControl/QcInfoLogger.h"
+#include <CommonUtils/StringUtils.h>
+#include <TMath.h>
+#include <TGraph.h>
+#include <TH1.h>
+#include <TCanvas.h>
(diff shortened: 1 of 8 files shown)


aferrero2707:
@Barthelemy @knopers8 this PR is rather big, but it does not touch existing code, it only introduces a new checker + some utility class to handle the configuration parameters.
Once it will be possible to retrieve the current interaction rate from CTP, the code is already designed to take advantage of it by specifying rate-dependent checker thresholds in the configuration.

aferrero2707:
@Barthelemy would you have the possibility to review this PR in the next days? Apart from the support for rate-dependent checker thresholds (which might only be useful at a later stage), the code could be already used to check the trends of the raw and CTF sizes, for example, and to replace the custom checker for the vertex positions and sigmas.
Thanks!

Barthelemy:
@aferrero2707 yes, apologies, this slipped through some hole and disappeared from my radar.

Modules/Common/src/TrendCheck.cxx
```diff
@@ -0,0 +1,466 @@
...
+  if (nPoints < 3) {
+    return result;
+  }
+
+  // index of the last-but-one point
+  int pointIndexMax = nPoints - 2;
+  // index of the starting point for the mean and stadrad deviation computation
+  int pointIndexMin = (nPointsForAverage > 0 && pointIndexMax >= nPointsForAverage) ? pointIndexMax - nPointsForAverage + 1 : 0;
+  // number of points used for the computation, must be greater than one
+  int N = pointIndexMax - pointIndexMin + 1;
```
Barthelemy, line 152: lower case and if possible descriptive name

Barthelemy (commented):
Very good, thanks. 

Could you perhaps add a bit of doc ? 
I am thinking that we should split the doc further as it becomes very large. We could have a separate page for the "commons". 
I'll do it once this is merged.

aferrero2707:
@Barthelemy thanks for reviewing!

> Could you perhaps add a bit of doc ? 

Yes, I was indeed planning to do that, I just wanted to wait for the first feedback in case it would required significant re-work.
What it the appropriate documentation file for putting the documentation for checkers?

Barthelemy:
@aferrero2707 You can put it in Advanced.md for now. I will refactor the doc and move all the info about common in a separate file.

Modules/Common/src/TrendCheck.cxx
aferrero2707: Done.

aferrero2707:
@Barthelemy I did the suggested change and added the documentation.

Barthelemy (approved):
Great ! thank you

PR #2385, opened 2024-08-14
ITS: Unified EPN QC Checks to work with MW

IsakovAD:
- Check on object name in QC Check not rely anymore on exact match -> should be work with 'mw' object names as well


iravasen:
Thanks @IsakovAD, I guess the previous code (with `==` in the `if` conditions) does not work anymore because now in the object name you have `int` for integrated objects (i.e. without MW) and `mw` for the objects included in the MW logic. Anyway, I approve it!

iravasen (approved):

PR #2386, opened 2024-08-14
Use int8_t rather than char

ktf:
Sign of char is architecture dependent


Barthelemy:
I think that it would require changes in PHOS code in O2 too...

ktf:
Yes, I need to think how to decouple them...

justonedev1:
@ktf @Barthelemy @knopers8 what is the progress of this PR? IT looks stale to me and might be better to close it without merging.

Barthelemy:
I would close it if @ktf agrees

PR #2389, opened 2024-08-16
Common: fix computation of output quality for ReferenceComparatorCheck

aferrero2707:
The new code fixes a bug that forces the output quality to be always Null, because the initial Null value is always not better than any other quality from the plots.


aferrero2707:
@Barthelemy @knopers8 @justonedev1 this is a rather trivial fix of a bug in the ReferenceComparatorCheck.

Barthelemy (approved):

PR #2390, opened 2024-08-16
ITS: reset of calibration plots in FEE

IsakovAD:
- Added reset of calibration plots in FEE for correct operations in `delta` mode


IsakovAD:
@iravasen this PR should improve results of FEE QC for calibration runs

iravasen (approved):

iravasen:
THanks a lot @IsakovAD ! 

@Barthelemy / @knopers8 since it is trivial and this fixes a bug in calibration, can we merge it by today? Just wanted to be sure it enters the SW upgrade of next week (if we are in time).

PR #2393, opened 2024-08-21
GLO: Change to TH1Ratio plot

f3sch:
Changes from TH1 to TH1Ratio to get a correct efficiency after the mergers. One question I have if after each endOfCycle also reset is called?

Tested with:
```bash
#!/usr/bin/env sh

GLOSET=" -b --shm-segment-size 4000000000 --hbfutils-config o2_tfidinfo.root,upstream --timeframes-rate-limit 1 --timeframes-rate-limit-ipcid 0 "
CONFIG="$ALIBUILD_WORK_DIR/../QualityControl/Modules/GLO/glo-itstpc-mtch-qcmn-test.json"

export IGNORE_VALIDITYCHECK_OF_CCDB_LOCALCACHE=1
export ALICEO2_CCDB_LOCALCACHE=$PWD/ccdb

o2-reader-driver-workflow --max-tf 2 $GLOSET | \
    o2-global-track-cluster-reader $GLOSET --disable-mc --track-types "ITS,TPC,ITS-TPC,ITS-TPC-TOF,TPC-TOF,TPC-TRD,ITS-TPC-TRD,TPC-TRD-TOF,ITS-TPC-TOF,ITS-TPC-TRD-TOF" --cluster-types "TPC,ITS" --primary-vertices --secondary-vertices | \
    o2-qc --full-chain --config json:///"$CONFIG" --timeframes-rate-limit 1 --timeframes-rate-limit-ipcid 0 -b --run | tee qc.log
```

Output on qcg-test:
image

Hopefully everything works now :)

Nota Bene: I added a manual calculation of binomial errors for the efficiency errors (which we might not care about) since TH1Ratio does not support it.

### DIFF
--- a/Modules/GLO/src/ITSTPCMatchingTask.cxx
+++ b/Modules/GLO/src/ITSTPCMatchingTask.cxx
@@ -167,26 +167,6 @@ void ITSTPCMatchingTask::endOfCycle()
       }
       ratio->Sumw2();
       ratio->update();
-
-      // Calculate binominal errors for efficiency, taken from ROOT
-      for (Int_t iBin = 0; iBin < ratio->GetNcells(); ++iBin) {
-        const Double_t b1 = ratio->getNum()->GetBinContent(iBin);
-        const Double_t b2 = ratio->getDen()->GetBinContent(iBin);
-        if (b2 == 0) {
-          ratio->SetBinError(iBin, 0);
-          continue;
-        }
-        const Double_t b1sq = b1 * b1;
-        const Double_t b2sq = b2 * b2;
-        const Double_t e1 = ratio->getNum()->GetBinError(iBin), e1sq = e1 * e1;
-        const Double_t e2 = ratio->getDen()->GetBinError(iBin), e2sq = e2 * e2;
-        if (b1 != b2) {
-          ratio->SetBinError(iBin, TMath::Sqrt(TMath::Abs(((1. - 2. * b1 / b2) * e1sq + b1sq * e2sq / b2sq) / b2sq)));
-        } else {
-          ratio->SetBinError(iBin, 0);
-        }
-      }
-
       return ratio;
     };
 
--- a/Modules/GLO/src/ITSTPCMatchingTask.cxx
+++ b/Modules/GLO/src/ITSTPCMatchingTask.cxx
@@ -151,45 +151,57 @@ void ITSTPCMatchingTask::endOfCycle()
 
   // Sync Mode
   if (common::getFromConfig(mCustomParameters, "isSync", false)) {
-    { // Pt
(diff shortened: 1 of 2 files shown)


chiarazampolli:
Hello @f3sch , 

I think that yes, after an `endOfCycle`, the `reset` is called, see https://github.com/AliceO2Group/QualityControl/blob/ad246d7ed5af3c2817bb8b22ac21938fe2054e05/Framework/src/TaskRunner.cxx#L362-L378, but better that @knopers8  and/or @Barthelemy confirm.

Chiara

f3sch:
@chiarazampolli super, thanks!

knopers8:
> One question I have if after each endOfCycle also reset is called?

Short answer - yes.
Long answer - it is called when running with Mergers in "delta" mode, which is the only reasonable to use with EPN.

The code that Chiara cited shows `TaskRunner::stop()`, this happens only at run stop, so it's irrelevant. This is where you can see the `reset()` after each cycle: https://github.com/AliceO2Group/QualityControl/blob/befd90539257fc292939044fd65bdc797bfeeab6/Framework/src/TaskRunner.cxx#L201

Modules/GLO/src/ITSTPCMatchingTask.cxx
```diff
@@ -151,45 +151,57 @@ void ITSTPCMatchingTask::endOfCycle()
...
+        if (b2 == 0) {
+          ratio->SetBinError(iBin, 0);
+          continue;
+        }
+        const Double_t b1sq = b1 * b1;
+        const Double_t b2sq = b2 * b2;
+        const Double_t e1 = ratio->getNum()->GetBinError(iBin), e1sq = e1 * e1;
+        const Double_t e2 = ratio->getDen()->GetBinError(iBin), e2sq = e2 * e2;
+        if (b1 != b2) {
+          ratio->SetBinError(iBin, TMath::Sqrt(TMath::Abs(((1. - 2. * b1 / b2) * e1sq + b1sq * e2sq / b2sq) / b2sq)));
```
knopers8, line 184: Do you know what happens to the errors when two objects are merged?

f3sch: Ah your right, when merged this would be done incorrectly... This should probably be done in the TH1Ratio::update under some flag. Which I guess I can do but do not know if I should?

knopers8: I think it's fine, but please do it in a separate PR and I will also ask Andrea to have a look.

f3sch: sure, dropped it for now.

knopers8 (approved):

PR #2394, opened 2024-08-22
TH1Ratio: Add weighted opt. binomial error calculation

f3sch:
This adds the correct weighted error (iff Sumw2) calculation with an optionally binomial treatment modelled what is done in ROOT to the TH1Ratio class. The errors are re-calculated in the update step thus also correctly treated during merging. 
Nota Bene: I do not know if we care about the error bars at all? I did not test the TH2Ratio but the code is the same.

With only the weighted errors:
image

plus the binomial treatment:
image

One benefit is that for efficiencies besides correct calculation of error-bars, is that the errors do not go beyond 1 anymore, which might spare some confusion.
I also added the use case to the GLO matching check.

### DIFF
--- a/Modules/Common/include/Common/TH1Ratio.inl
+++ b/Modules/Common/include/Common/TH1Ratio.inl
@@ -149,9 +149,8 @@ void TH1Ratio<T>::update()
   T::GetXaxis()->Set(mHistoNum->GetXaxis()->GetNbins(), mHistoNum->GetXaxis()->GetXmin(), mHistoNum->GetXaxis()->GetXmax());
   T::SetBinsLength();
 
-  T::Add(mHistoNum);
-
   if (mUniformScaling) {
+    T::Add(mHistoNum);
     double entries = mHistoDen->GetBinContent(1);
     double norm = (entries > 0) ? 1.0 / entries : 0;
     // make sure the sum-of-weights structure is not initialized if not required
@@ -167,7 +166,7 @@ void TH1Ratio<T>::update()
         mHistoDen->GetXaxis()->SetBinLabel(bin, T::GetXaxis()->GetBinLabel(bin));
       }
     }
-    T::Divide(mHistoDen);
+    T::Divide(mHistoNum, mHistoDen, 1.0, 1.0, mBinominalErrors ? "B" : "");
   }
 }
 
--- a/Modules/Common/include/Common/TH1Ratio.h
+++ b/Modules/Common/include/Common/TH1Ratio.h
@@ -60,6 +60,10 @@ class TH1Ratio : public T, public o2::mergers::MergeInterface
     return mUniformScaling;
   }
 
+  // Use Binominal errors; only works if Sumw2 was called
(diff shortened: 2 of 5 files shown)


knopers8:
@aferrero2707 could you please also cross-check the modifications to `THRatio`?

Modules/Common/include/Common/TH1Ratio.inl
```diff
@@ -168,6 +168,29 @@ void TH1Ratio<T>::update()
       }
     }
     T::Divide(mHistoDen);
+    if (mSumw2Enabled) { // Re-calculate weighted errors; taken from https://root.cern.ch/doc/master/classTH1.html#ae7c34d74d043ff9d6816e5b026f85133
```
aferrero2707, line 171: @f3sch in I understand correctly, you found that `T::Divide()` does not compute correctly the errors, even if `Sumw2()` was called beforehand for all the histograms? Do you understand why?

f3sch: Hi @aferrero2707, you are right Divide already does the right thing for the weighted errors, but there is no option to instead do binomial error propagation. I changed it accordingly.

aferrero2707: Thanks @f3sch! Have you already double-checked that `Divide()` really computes the errors correctly? I have to admit I just assumed it was ok...

f3sch: @aferrero2707, yes I did.

aferrero2707: @f3sch then I am wondering if it would be better to use the ROOT functions for the binomial errors as well.
Instead of `T::Divide(mHistoDen);` one could call `T::Divide(mHistoNum, mHistoDen, 1, 1, option);`, and set `option` to `"b"` if binomial errors are requested. 

In this case `T::Add(mHistoNum);` would be needed only in the uniform scaling case.

f3sch: Done.

aferrero2707: @f3sch if you have a test set-up available, I'd recomment to check that ROOT is really doing what you expect for the binomial errors, just to be 100% sure that your custom code is really not needed. Thanks a lot!

f3sch: @aferrero2707, the plots look identical to ones I have posted above.

aferrero2707: Very good! I have no more comments.

aferrero2707:
@knopers8 @Barthelemy for me the latest version is ok.

Modules/Common/include/Common/TH1Ratio.inl
f3sch: thanks for your comments!

knopers8 (approved):
Thanks to both of you!

PR #2397, opened 2024-08-26
[QC-803] added config for trendingTask that enable trending only if all source…

justonedev1:
…s are available


Framework/include/QualityControl/TrendingTaskConfig.h
```diff
@@ -61,6 +61,7 @@ struct TrendingTaskConfig : PostProcessingConfig {
 
   bool producePlotsOnUpdate{};
   bool resumeTrend{};
+  bool trendOnlyIfAllInputsAreAvailable{ false };
```
knopers8, line 64: I think you also have to read this config value from the provided configuration, see the .cxx

knopers8:
> Is there a way, how to test this?

Yes. There is `testTrendingTask` that you could modify accordingly. Otherwise, you could also take the example from here: https://github.com/AliceO2Group/QualityControl/blob/master/doc/QuickStart.md#post-processing-example and for example switch off the Check, so one of the two data sources in https://github.com/AliceO2Group/QualityControl/blob/master/Framework/postprocessing.json#L64 would be missing.

justonedev1:
Thank you! I was not sure how to turn off one of the datasources, thank you

doc/PostProcessing.md
```diff
@@ -389,6 +389,8 @@ use the boolean flag `"producePlotsOnUpdate"`.
 
 To pick up the last existing trend which matches the specified Activity, set `"resumeTrend"` to `"true"`.
 
+To generate plots only when all input objects are available, set `"trendOnlyIfAllInputsAreAvailable"`.
```
knopers8, line 392: I am wondering if we could make the parameter shorter... Perhaps "trendIfAllInputs"?

PR #2400, opened 2024-08-28
[QC-1176] Display a warning for developers if a registered object is not mergeable

justonedev1:
this is just a draft with an idea, how to do it compile time, I am not sure, that you like it. It does not build, it would need a bit more work. Otherwise I don't know how to do this compile time and it would need to be done run time via `dynamic_cast`

### DIFF
--- a/Framework/src/ObjectsManager.cxx
+++ b/Framework/src/ObjectsManager.cxx
@@ -60,19 +60,22 @@ ObjectsManager::~ObjectsManager()
   ILOG(Debug, Devel) << "ObjectsManager destructor" << ENDM;
 }
 
-void ObjectsManager::startPublishingImpl(TObject* object, PublicationPolicy publicationPolicy)
+void ObjectsManager::startPublishingImpl(TObject* object, PublicationPolicy publicationPolicy, bool ignoreMergeableWarning)
 {
   if (!object) {
     ILOG(Warning, Support) << "A nullptr provided to ObjectManager::startPublishing" << ENDM;
     return;
   }
+
   if (mMonitorObjects->FindObject(object->GetName()) != nullptr) {
     ILOG(Warning, Support) << "Object is already being published (" << object->GetName() << "), will remove it and add the new one" << ENDM;
     stopPublishing(object->GetName());
   }
-  if (mergers::isMergeable(object)) {
-    ILOG(Warning, Support) << "Object '" + std::string(object->GetName()) + "' with type '" + std::string(object->ClassName()) + "' is not one of the mergeable types, it might cause issues during publishing";
+
+  if (!ignoreMergeableWarning && !mergers::isMergeable(object)) {
+    ILOG(Warning, Support) << "Object '" + std::string(object->GetName()) + "' with type '" + std::string(object->ClassName()) + "' is not one of the mergeable types, it will not be correctly merged in distributed setups, such as P2 and Grid" << ENDM;
   }
+
   auto* newObject = new MonitorObject(object, mTaskName, mTaskClass, mDetectorName);
   newObject->setIsOwner(false);
   newObject->setActivity(mActivity);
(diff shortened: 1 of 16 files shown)


knopers8:
For the record, we discussed in person, the agreed improvements are:
- move the type handling logic to Mergers
- find a way to avoid including all of the complete types that are checked (typically, the user will include what they need)
- fail the compilation if the type is not mergeable, but add another method like `"startPublishingNotMergeable"` to allow expert users to do what they want.

justonedev1:
I think that I addressed everything that we talked about, but a bit different way. If you would prefer something else, I can do that.

Framework/src/ObjectsManager.cxx
```diff
@@ -70,6 +70,9 @@ void ObjectsManager::startPublishing(TObject* object, PublicationPolicy publicat
     ILOG(Warning, Support) << "Object is already being published (" << object->GetName() << "), will remove it and add the new one" << ENDM;
     stopPublishing(object->GetName());
   }
+  if (mergers::isMergeable(object)) {
```
knopers8, line 73: I think this should be also disabled by the presence of `QUALITYCONTROL_POSTPROCESSINTERFACE_H` and `IgnoreMergable==true`, otherwise we will see a lot of spurious warnings.

Framework/src/ObjectsManager.cxx
```diff
@@ -70,6 +70,9 @@ void ObjectsManager::startPublishing(TObject* object, PublicationPolicy publicat
...
+    ILOG(Warning, Support) << "Object '" + std::string(object->GetName()) + "' with type '" + std::string(object->ClassName()) + "' is not one of the mergeable types, it might cause issues during publishing";
```
knopers8, line 74:
```suggestion
    ILOG(Warning, Support) << "Object '" + std::string(object->GetName()) + "' with type '" + std::string(object->ClassName()) + "' is not one of the mergeable types, it will not be correctly merged in distributed setups, such as P2 and Grid" << ENDM;
```
missing ENDM and rephrasing

knopers8 (changes requested):
Thanks, just some details to sort out, but in general the idea is good.

Do I understand correctly that the inclusion order of `ObjectManager.h` and `PostProcessingInterface.h` should not matter because the template is put in place after both?

Framework/src/ObjectsManager.cxx
knopers8: @justonedev1 can you explain how your most recent changes make `QUALITYCONTROL_POSTPROCESSINTERFACE_H` disable this warning? I'm convinced they do.

Framework/src/TrendingTask.cxx
```diff
@@ -257,7 +257,7 @@ void TrendingTask::generatePlots()
     }
     auto c = drawPlot(plotConfig);
     mPlots[plotConfig.name].reset(c);
-    getObjectsManager()->startPublishing(c, PublicationPolicy::Once);
+    getObjectsManager()->startPublishing<true>(c, PublicationPolicy::Once);
```
knopers8, line 260: you should not need `<true>` in TrendingTask and SliceTrendingTask anymore, after you disabled the check for postprocessing tasks.

Framework/src/ObjectsManager.cxx
justonedev1: They do not. Somehow my changes got deleted, sorry about that

justonedev1: it should be done now

justonedev1:
@wiechula Can you please take a look at this PR?

Modules/TPC/src/PID.cxx
```diff
@@ -69,6 +69,11 @@ void PID::initialize(o2::framework::InitContext& /*ctx*/)
       getObjectsManager()->startPublishing(hist.get());
     }
   }
+  for (auto const& pair : mQCPID.getMapOfCanvas()) {
+    for (auto& canv : pair.second) {
+      getObjectsManager()->startPublishing<true>(canv.get());
+    }
+  }
```
knopers8, line 76: why publish those objects if they were not published before?

justonedev1: It must have been changed since I did the PR, good catch

PR #2403, opened 2024-08-30
QC-604 removed analysis integration except of QCDB upload

knopers8:
The possibility to consume AODs and generating DPL jsons to run in the train system proved not to be used and needed in the end. `o2-qc-upload-root-objects` remains the only allowed interface between Analysis and QC.


Barthelemy (approved):
Always nice to see things go away and get simplified.

PR #2407, opened 2024-09-03
[QC-1229] Repocleaner policy for the moving windows

Barthelemy:
modify `multiple_per_run` to have an option for the moving windows. 
I did not want to duplicate the policy thus I added an option that will trigger the deletion after a certain amount of time. 
As I was afraid that it would for some reason delete "normal" data, I added a check and it only deletes if it is in a path containing `mw`.

### DIFF
--- a/Framework/script/RepoCleaner/qcrepocleaner/rules/multiple_per_run.py
+++ b/Framework/script/RepoCleaner/qcrepocleaner/rules/multiple_per_run.py
@@ -109,7 +109,7 @@ def process(ccdb: Ccdb, object_path: str, delay: int,  from_timestamp: int, to_t
         elif first_object.createdAtDt < datetime.now() - timedelta(minutes=mw_deletion_delay):
             logger.debug(f"     after mw_deletion_delay period, delete this bucket")
             for v in run_versions:
-                if "mw" in v.path: # this is because we really don't want to take the risk of batch deleting non moving windows
+                if "/mw/" in v.path: # this is because we really don't want to take the risk of batch deleting non moving windows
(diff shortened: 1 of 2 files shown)


Framework/script/RepoCleaner/qcrepocleaner/rules/multiple_per_run.py
```diff
@@ -95,9 +103,19 @@ def process(ccdb: Ccdb, object_path: str, delay: int,  from_timestamp: int, to_t
...
             logger.debug(f"     in grace period, skip this bucket")
             preservation_list.extend(run_versions)
-        elif not (from_timestamp < first_object.createdAt < to_timestamp):  # in the allowed period
+        elif not (from_timestamp < first_object.createdAt < to_timestamp):  # not in the allowed period
             logger.debug(f"     not in the allowed period, skip this bucket")
             preservation_list.extend(run_versions)
+        elif first_object.createdAtDt < datetime.now() - timedelta(minutes=mw_deletion_delay):
+            logger.debug(f"     after mw_deletion_delay period, delete this bucket")
+            for v in run_versions:
+                if "mw" in v.path: # this is because we really don't want to take the risk of batch deleting non moving windows
```
knopers8, line 112:
```suggestion
                if "/mw/" in v.path: # this is because we really don't want to take the risk of batch deleting non moving windows
```
I would even go for "/mw/"`, otherwise you would include names such as `clusters_sumw2`.

Barthelemy: good point, thank you

knopers8 (approved):

PR #2408, opened 2024-09-03
Avoid calling Sumw2 if it was already called

shahor02:
To suppress errors like https://ali-bookkeeping.cern.ch/?page=log-detail&id=107930


knopers8 (approved):
thank you

PR #2410, opened 2024-09-04
QC-1214 Updates and improvements to QO -> Flag conversion

knopers8:
The converter has been adapted to the rules below:
- **Good QOs with no Flags are not onverted to any Flags**
- **Bad and Medium QOs with no Flags are converted to Flag 14 (Unknown)**
- **Null QOs with no Flags are converted to Flag 1**
- **All QOs with Flags are converted to Flags, while Quality is ignored**. A warning should be printed if a Check associates a GOOD flag to BAD quality, or a BAD flag to GOOD quality.
- **Timespans not covered by a given QO are filled with Flag 1 (Unknown Quality)**
- **Overlapping or adjacent Flags with the same ID, comment and source (QO) are merged, even if they were associated to different Qualities.**
- **Flag 1 (UnknownQuality) gets overwritten by any other Flag**.
- **Good and Bad flags do not affect each other, they may coexist**.
- **Flags for different QOs do not affect each other, Flag 1 is added separately for each.**
- **The order of QO arrival should not matter**. Today, `QualitiesToFlagCollectionConverter` assumes it should receive quality flags in chronological order. However, this may not always be the case in sync QC, where validity start might slightly move to the past due unaligned validities of objects coming from EPNs, being merged in the merger nodes.

Additionally the tests were extended and updated.

### DIFF
--- a/Framework/CMakeLists.txt
+++ b/Framework/CMakeLists.txt
@@ -119,7 +119,8 @@ add_library(O2QualityControl
   src/TimekeeperFactory.cxx
   src/RootFileStorage.cxx
   src/ReductorHelpers.cxx
-  src/KafkaPoller.cxx)
+  src/KafkaPoller.cxx
+  src/FlagHelpers.cxx)
 
 target_include_directories(
   O2QualityControl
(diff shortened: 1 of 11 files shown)


knopers8:
@justonedev1 @Barthelemy ready for review

Barthelemy (approved):

PR #2411, opened 2024-09-08
ITS - rearrangement of plots and checks for shifters

nicolovalle:
In this PR:

- The Deconding Errors vs FeeID can be checked, and it is possible to ignore multiple occurrences of the same error on the same FeeID
- The ChipStatus task publishes an object with the fraction chips that are missing in the last N cycles, versus FeeID
   - This is passed through a checker. Different thresholds can be set for IB, ML and OL


IsakovAD (approved):
all is good!

PR #2416, opened 2024-09-11
Common: allow retrieving latest QOs in QualityTask

aferrero2707:
Instead of retrieving the QOs based on the time stamp of the update trigger, the latest version of the QOs is retrieved, even if their validities are ending before the update time stamp.

A `maxObjectAgeSeconds` configuration parameter allows to consider only QOs that were created not more than a given number of seconds before the time at which the `update()` function was called.

This logic only works for the online QualityControl, where the QOs creation time follows is aligned with the time of the QC cycles and of the post-processing updates.

### DIFF
--- a/Modules/Common/include/Common/QualityTask.h
+++ b/Modules/Common/include/Common/QualityTask.h
@@ -83,12 +83,14 @@ class QualityTask : public quality_control::postprocessing::PostProcessingInterf
   void finalize(quality_control::postprocessing::Trigger, framework::ServiceRegistryRef) override;
 
  private:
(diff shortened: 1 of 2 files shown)


aferrero2707:
@knopers8 @Barthelemy this is the PR associated to the changes in the QOs validities discussed in https://its.cern.ch/jira/browse/QC-1223?focusedId=6519202&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-6519202

There is one CI failure, but seems to be completely unrelated. CI was green before I added the commit for the documentation.

knopers8 (approved):
thank you

PR #2164, opened 2024-03-03
Use lightweight its::GeometryTGeo instead of full geometry

shahor02:

PR #2418, opened 2024-09-12
ITS alignment monitoring (check with QCv20240903), revised 20240919

delico18:
New 14 plots(7, layer x 2, axis[z \~ alpide col, phi \~ alpide row] for ITS alignment monitoring are added.

- Circle Fitter (in XY plane)
- Line Fitter (in DZ plane), D is almost same as the path length of particle.

Based on Least Square Fit, we can observe the residual of particles' track, which we believe they will show the alignment quality of ITS.

### DIFF
--- a/Modules/ITS/include/ITS/ITSTrackTask.h
+++ b/Modules/ITS/include/ITS/ITSTrackTask.h
@@ -127,15 +127,15 @@ class ITSTrackTask : public TaskInterface
   o2::itsmft::TopologyDictionary* mDict;
 
  private:
-  //analysis for its-only residual
+  // analysis for its-only residual
   o2::its::GeometryTGeo* mGeom;
 
-  std::vector<double> FitStepSize{0.3, 1.0e-5, 1.0e-5, 1.0e-5};
+  std::vector<double> FitStepSize{ 0.3, 1.0e-5, 1.0e-5, 1.0e-5 };
 
   double FitTolerance = 1.0e-8;
-  double ITS_AbsBz = 0.5; //T
+  double ITS_AbsBz = 0.5; // T
 
-  double inputG_C[3*NLayer];
+  double inputG_C[3 * NLayer];
   double FitparXY[6];
   double FitparDZ[2];
   ROOT::Fit::Fitter fitterA;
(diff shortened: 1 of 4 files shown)


Modules/ITS/src/ITSTrackTask.cxx
```diff
@@ -92,6 +111,10 @@ void ITSTrackTask::monitorData(o2::framework::ProcessingContext& ctx)
 
   ILOG(Debug, Devel) << "START DOING QC General" << ENDM;
 
+  //o2::base::GeometryManager::loadGeometry();
+  mGeom = o2::its::GeometryTGeo::Instance();
+  mGeom->fillMatrixCache(o2::math_utils::bit2Mask(o2::math_utils::TransformType::L2G));
```
iravasen, line 116: move these two lines inside the `if(mTimestamp == -1)`.

iravasen: In addition here you are using the full geometry. I guess a light-weight version of it should be fine. Please copy what was done for the ClusterTask here: https://github.com/AliceO2Group/QualityControl/pull/2164/files#diff-41e35aa0ad16f0bc0150242f13cb67d78b78f53b020364e33f18cce3fcf44de4R125 

Note also the removal from the json of the `grpGeomRequest`

Modules/ITS/itsTrack.json
```diff
@@ -45,9 +45,24 @@
...
-          "InvMasses" : "0"
+          "InvMasses" : "0",
+          "UseDefaultMomResPar" : "0",
+          "MomResParMEAS1": "45, 45, 45, 55, 55, 55, 55",
+          "MomResParMEAS2": "40, 40, 40, 40, 40, 40, 40",
+          "MomResParMSC1": "30, 30, 30, 110, 110, 110, 110",
+          "MomResParMSC2": "25, 25, 25, 75, 75, 75, 75",
+          "ResidualMonitorTrackMinPt" : "0.00"
+        },
+	"grpGeomRequest" : {
```
iravasen, line 56: `grpGeomRequest` can be removed if you use the light weight version of the geometry. See the other comment.

Modules/ITS/src/ITSTrackTask.cxx
```diff
@@ -628,6 +723,21 @@ void ITSTrackTask::createAllHistos()
   addObject(hTrackPtVsPhi);
   formatAxes(hTrackPtVsPhi, "#it{p}_{T} (GeV/#it{c})", "#phi", 1, 1.10);
   hTrackPtVsPhi->SetStats(0);
+
+  for (int l = 0; l < NLayer; l++) {
+    //sensor
+    hdxySensor[l] = std::make_unique<TH2D>(Form("hdxy%dSensor",l), Form("ChipID vs dxy, Layer %d",l), 
+                                           1000, -0.05, 0.05, ChipBoundary[l+1] - ChipBoundary[l], ChipBoundary[l], ChipBoundary[l+1]);
```
iravasen, line 730: do you really need 1000 bins on the X axis? Same comment for `hdz` histo. Bin width is 1 um like this, maybe 5 um are enough (i.e. comparable with the ALPIDE spatial reso).

Modules/ITS/src/ITSTrackTask.cxx
```diff
@@ -648,4 +758,455 @@ void ITSTrackTask::publishHistos()
...
+        initR.push_back(TVector3(CenterX,CenterY,temp_R));
+        if(i2<2) {
+           hit_mid=true; // mid hit is successfully used. Do not find inner or outer hits for initial radius searching
+        }
+        cntR[0]++;
+      }  
+    }
+  }
+  
+  /*
```
iravasen, line 864: remove this long comment.

Modules/ITS/src/ITSTrackTask.cxx
iravasen: Make the y axis bin with the chipID in the center: 
`ChipBoudary[l+1] - ChipBoundary[l] + 1, ChipBoundary[l] - 0.5, ChipBoundary[l+1] - 0.5`

Note the -0.5 also for the upper limit because for example chip 108 belongs already to L1 (and not to L0) if we start counting chipID from 0. Maybe make again the calculation but it should be correct.

Modules/ITS/include/ITS/ITSTrackTask.h
```diff
@@ -118,6 +125,147 @@ class ITSTrackTask : public TaskInterface
...
+  ROOT::Fit::Fitter fitterB;
+  Int_t mDefaultMomResPar = 0;
+  float mResMonTrackMinPt = 0;
+
+  TH1D* hResidualRealTimePerTrackwFinerBin;
+  TH1D* hResidualCPUTimePerTrackwFinerBin;
+  std::array<std::unique_ptr<TH2D>, NLayer> hdxySensor{};//[NLayer];
+  std::array<std::unique_ptr<TH2D>, NLayer> hdzSensor{};//[NLayer];
+
+  void circlefit_XY(double* input, double* par, double &MSEvalue, std::vector<bool> hitUpdate, int step = 0);
```
iravasen, line 154: use the camel case for all functions: `circlefit_XY` becomes `circleFitXY`. This is the convention in QC.

iravasen (changes requested):

iravasen:
I will apply the clang format at the end once all comments have been implemented

Modules/ITS/itsTrack.json
```diff
@@ -45,9 +45,24 @@
           "nBCbins": "103",
           "dicttimestamp" : "0",
           "doNorm" : "1",
-          "InvMasses" : "0"
+          "InvMasses" : "0",
+          "UseDefaultMomResPar" : "0",
```
iravasen, line 49: Can you please add an option to enable/disable your new plots? Default is 1 (i.e. enabled).

iravasen (changes requested):

delico18:
2024 09 19 update

There seems to be tiny errors in getting layer of cluster, and I corrected the lines.

Options...
to activate alignment monitoring(itsTrack.json) is added.
to request the minimum number of clusters / track for alignment monitoring(itsTrack.json) is added.

I updated the simple version of documentation that describes the fitting & residual calculation procedures.

Modules/ITS/src/ITSTrackTask.cxx
```diff
@@ -628,6 +727,21 @@ void ITSTrackTask::createAllHistos()
   addObject(hTrackPtVsPhi);
   formatAxes(hTrackPtVsPhi, "#it{p}_{T} (GeV/#it{c})", "#phi", 1, 1.10);
   hTrackPtVsPhi->SetStats(0);
+
+  for (int l = 0; l < NLayer; l++) {
+    //sensor
```
iravasen, line 732: Initialize histograms only `if (mAlignmentMonitor == 1)`

iravasen (changes requested):

Modules/ITS/src/ITSTrackTask.cxx
```diff
@@ -441,6 +535,11 @@ void ITSTrackTask::reset()
 
   hTrackPtVsEta->Reset();
   hTrackPtVsPhi->Reset();
+
+  for (int l = 0; l < NLayer; l++) {
+    hdxySensor[l]->Reset();
```
iravasen, line 540: Reset histo only `if mAlignmentMonitor == 1`

iravasen (changes requested):

Modules/ITS/include/ITS/ITSTrackTask.h
```diff
@@ -118,6 +125,149 @@ class ITSTrackTask : public TaskInterface
...
   o2::itsmft::TopologyDictionary* mDict;
+
+ private:
+  //analysis for its-only residual
+  o2::its::GeometryTGeo* mGeom;
+
+  double FitStepSize0 = 0.3;
+  double FitStepSize1 = 1.0e-5;
+  double FitStepSize2 = 1.0e-5;
+  double FitStepSize3 = 1.0e-5;
```
IsakovAD, line 136: Could allocate a vector for that?

Modules/ITS/include/ITS/ITSTrackTask.h
```diff
@@ -118,6 +125,149 @@ class ITSTrackTask : public TaskInterface
...
+  ROOT::Fit::Fitter fitterB;
+  Int_t mAlignmentMonitor = 0;
+  Int_t mDefaultMomResPar = 0;
+  Int_t mResMonNclMin = 0;
+  float mResMonTrackMinPt = 0;
+
+  TH1D* hResidualRealTimePerTrackwFinerBin;
+  TH1D* hResidualCPUTimePerTrackwFinerBin;
+  std::array<std::unique_ptr<TH2D>, NLayer> hdxySensor{};//[NLayer];
+  std::array<std::unique_ptr<TH2D>, NLayer> hdzSensor{};//[NLayer];
```
IsakovAD, line 154: Do you that that we need TH2D in this case? The general issue which we are dealing with is the size of objects. Could you try configuration with TH2S or TH2I. In my understanding, you have very fine binning and we do not expect to have so many entries in each bin

Modules/ITS/include/ITS/ITSTrackTask.h
```diff
@@ -118,6 +125,149 @@ class ITSTrackTask : public TaskInterface
...
+      return sum;
+    };
+
+  };
+
+  se_circlefitXY fitfuncXY;
+
+  void lineFitDZ(double* zIn, double* betaIn, double* parz, double Radius, bool vertex, std::vector<bool> hitUpdate);
+
+  double sigma_meas[2][NLayer] = {{45,45,45,55,55,55,55},
```
IsakovAD, line 254: Is it a duplicate on of line 165?

Modules/ITS/src/ITSTrackTask.cxx
```diff
@@ -100,6 +121,11 @@ void ITSTrackTask::monitorData(o2::framework::ProcessingContext& ctx)
     std::map<std::string, std::string> metadata;
     mDict = TaskInterface::retrieveConditionAny<o2::itsmft::TopologyDictionary>("ITS/Calib/ClusterDictionary", metadata, ts);
     ILOG(Debug, Devel) << "Dictionary size: " << mDict->getSize() << ENDM;
+
+    o2::its::GeometryTGeo::adopt(TaskInterface::retrieveConditionAny<o2::its::GeometryTGeo>("ITS/Config/Geometry", metadata, ts));
```
IsakovAD, line 125: If we do not use geometry in the code besides alignment monitoring, could you make geometry downloading optional?

IsakovAD:
Hello, 

Nice new feature! General comments: 
1) Size of new objects: I assume that TH2D is too much. Could you please check the size of objects in case of TH2D and TH2S/TH2I. I think that with such nice binning double is a quite overkill 
2) Try to minimize influence of your code when AlignmentMonitoring set to FALSE: do not initialize objects/do not download geometry etc. Please, check that task still working if user will set monitoring to OFF.

 Otherwise,  looking forward to results of this new feature

Modules/ITS/src/ITSTrackTask.cxx
delico18: It is implemented in the code and failure case(throw) to download ccdb object for ITS geometry is added.

Modules/ITS/include/ITS/ITSTrackTask.h
delico18: No, one is data member of ITSTrackTask, and another one is data member of struct "se_circlefitXY".

I renamed the former case as "mSigmaMeas, mSigmaMsc" to have similar patterns for the normal data member in ITSTrackTask.

Modules/ITS/include/ITS/ITSTrackTask.h
delico18: I changed this to TH2I, but I think TH2F can be more generally used in case we just take care of the size of objects.
Because the size of TH2I and TH2F is the same, but the maximum bin content for TH2F is more large.

If there is no problem using TH2F, I'd like to choose TH2F. Please leave your comments.

delico18:
The monitoring objects for residuals are only activated when mAlignmentMonitor is true.
There is no more influence related to this object when this is set to false.

Please review this PR.(including TH2F vs TH2I)

IsakovAD (approved):

iravasen (approved):

PR #2420, opened 2024-09-17
DAQ: properly reset and un-publish mSumRDHSizesInTF

aferrero2707:
The `mSumRDHSizesInRDH` plot was reset and un-published twice. In the second of each calls `mSumRDHSizesInTF` should appear instead.


aferrero2707:
@knopers8 @justonedev1 a minor fix to the DaqTask.

knopers8 (approved):
thank you!

PR #2422, opened 2024-09-19
Minor fixes for kafka dependency

knopers8:
For some reason I cannot build QC without these fixes on my setup. This includes:
- fixing a typo RDFKAFKA_INCLUDE -> RDKAFKA_INCLUDE
- include modern cpp kafka headers
- create `protos` directory in build directory when it's not there


Barthelemy:
I had the same errors when building with CLion but not building with the terminal. 

Unfortunately it seems that it broke the build system. 
I am now trying on my setup.

Barthelemy (approved):

PR #2423, opened 2024-09-19
QC-1233 Allow to select trending point timestamp in (Slice)TrendingTasks

knopers8:
This also deprecates the behaviour for old-style validity objects which to my knowledge are not trended anymore and repo-cleaner usually updates them to contain a more realistic validity intervals. Also, if the aggregated quality validity is invalided, it is now replaced with last timestamp of the latest object instead of using the old-style validity, so we will not get the 10-years validity there anymore.


Barthelemy (approved):
thank you

PR #2425, opened 2024-09-19
AFIT-101: adding min bias to plots(FT0)

afurs:
Uses functionality from https://github.com/AliceO2Group/AliceO2/pull/13514


Barthelemy (approved):

Barthelemy:
force merging because Andreas is absent

PR #2426, opened 2024-09-20
CTP: PbPb run preparation

lhusova:
@lietava, please have a look whether everything is included

### DIFF
--- a/Modules/CTP/src/RawDataQcTask.cxx
+++ b/Modules/CTP/src/RawDataQcTask.cxx
@@ -126,7 +126,7 @@ void CTPRawDataReaderTask::startOfActivity(const Activity& activity)
     indexMB1 = 3; // 3 is the MTVX index
   }
   if (indexMB2 == -1) {
-    indexMB2 = 1; // 3 is the MT0A index
+    indexMB2 = 5; // 5 is the MTCE index
   }
   for (int i = 0; i < nclasses; i++) {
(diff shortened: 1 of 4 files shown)


Modules/CTP/src/RawDataQcTask.cxx
```diff
@@ -116,13 +117,16 @@ void CTPRawDataReaderTask::startOfActivity(const Activity& activity)
...
+  indexMB1 = o2::ctp::CTPInputsConfiguration::getInputIndexFromName(nameInput1);
+  indexMB2 = o2::ctp::CTPInputsConfiguration::getInputIndexFromName(nameInput2);
+  if (indexMB1 == -1) {
+    indexMB1 = 3; // 3 is the MTVX index
   }
-  indexTvx = o2::ctp::CTPInputsConfiguration::getInputIndexFromName(nameInput);
-  if (indexTvx == -1) {
-    indexTvx = 3; // 3 is the MTVX index
+  if (indexMB2 == -1) {
+    indexMB2 = 1; // 3 is the MT0A index
```
lietava, line 129: why not tu put default3 which is MTVX ?

lietava, line 129: For MB2 I would add deafult 5 which is TCE so we have good default for Pb.

knopers8 (approved):

PR #2428, opened 2024-09-22
THRatio: fix again division of histograms with different labels

aferrero2707:
The ROOT `Divide()` function requires that bin labels, if present, are the same in the numerator and the denominator, otherwise it prints a warning message.

The changes in this commit fix the copy of bin labels from numerator to denominator in the THxRatio `update()` functions, that was broken by recent updates.

This should mute the frequent warning messages for the QC merger nodes, similar to those below:
```
Info in <o2::quality_control_modules::common::TH1Ratio<TH1F>::Divide>: Histograms have different labels task="alio2-cr1-hv-gw01.cern.ch:/opt/git/ControlWorkflows/tasks/jit-dab50f7f440359627025f1426c693f34bed54029-ITS-MERGER-ITSDECODING1l-0@4ca2b184b590c6547212a91c5f30499310ba0cd3#2pxoSbRF2EY"
Info in <o2::quality_control_modules::common::TH2Ratio<TH2F>::Divide>: Histograms have different labels task="alio2-cr1-hv-gw01.cern.ch:/opt/git/ControlWorkflows/tasks/jit-69a39e02ed4aa6dafaebcbe990508b140b68d691-MCH-MERGER-Decoding1l-0_t0@4ca2b184b590c6547212a91c5f30499310ba0cd3#2pxoSsbhBcp"
```

### DIFF
--- a/Modules/Common/include/Common/TH1Ratio.inl
+++ b/Modules/Common/include/Common/TH1Ratio.inl
@@ -19,6 +19,28 @@
 namespace o2::quality_control_modules::common
 {
 
+namespace th1ratio_internal
+{
+
+template <class T>
+bool copyBinLabels(const T* srcHist, T* destHist)
+{
+  if (srcHist->GetXaxis()->GetNbins() != destHist->GetXaxis()->GetNbins()) {
+    return false;
+  }
+
+  bool result = false;
+  if (srcHist->GetXaxis()->GetLabels())  {
+    for (int bin = 1; bin <= srcHist->GetXaxis()->GetNbins(); bin++) {
+      destHist->GetXaxis()->SetBinLabel(bin, srcHist->GetXaxis()->GetBinLabel(bin));
+    }
+    result = true;
+  }
+  return result;
+}
+
+}
+
 template<class T>
 TH1Ratio<T>::TH1Ratio()
   : T(),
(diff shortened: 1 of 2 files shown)


aferrero2707:
@knopers8 @Barthelemy we had again complaints from ROOT about different bin labels in the THxRatio plots, this PR should fix it.

knopers8:
Thank you @aferrero2707 ! Just to understand, do you know why these labels become missing or mismatching in the first place?

aferrero2707:
@knopers8 I think that my previous code was simply wrong. When labels were set on the THRatio object (like [here](https://github.com/AliceO2Group/QualityControl/blob/255f9d3e5c1f17a92d87a5eb5e9ef7647969418e/Modules/MUON/MCH/src/DecodingTask.cxx#L60)), they were copied from the THRatio to the denominator (see [here](https://github.com/AliceO2Group/QualityControl/blob/4d32c59a698c377ff149e50843b00d79d792e4d0/Modules/Common/include/Common/TH1Ratio.inl#L164)) but not to the numerator, and this was triggering the ROOT warning.

The new code in this PR propagates the labels to all the plots. In addition to avoiding the ROOT warning message, it also correctly handles the case were bin labels are set in the numerator plot instead of the THRatio itself.

knopers8 (approved):

knopers8:
thanks for the explanation!

PR #2429, opened 2024-09-22
Common: reference comparator check improvements

aferrero2707:
Several improvements to the reference comparator check:
* optional X and Y ranges can be specified in the configuration to restrict the comparison to a given histogram area
* the comparators now access directly the custom parameters, allowing to have parameters specific to each comparator
* added a `ObjectComparatorBinByBinDeviation` class that checks the relative deviation between each bin of the current and reference histograms
    * A `maxAllowedBadBins` parameter allows to specify the maximum number of bins above threshold for which the quality is still considered Good. 
    * If X and/or Y ranges are specified, only the bins inside the ranges are considered for the comparison.
* The "average deviation" and "Chi2 test"  comparison methods are improved to take into account the X and/or Y ranges in the comparison.
* For 1-D histograms, the checker shows the X range (if specified) on the ratio plot as a double-arrow horizontal line

### DIFF
--- a/Modules/Common/CMakeLists.txt
+++ b/Modules/Common/CMakeLists.txt
@@ -24,6 +24,7 @@ target_sources(O2QcCommon
                        src/CcdbInspectorCheck.cxx
                        src/ObjectComparatorInterface.cxx
                        src/ObjectComparatorDeviation.cxx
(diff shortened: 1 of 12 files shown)


Barthelemy:
thank you, 
I will review it in the coming days.

PR #2432, opened 2024-09-23
TPC QC: Update utility for timestamp extraction to optimize it 

lauraser:
+ switches for IDC task what to plot


knopers8 (approved):

wiechula (changes requested):

lauraser:
@wiechula I implemented all your suggestions. Regarding the call .list() for last two weeks, I also additionally added checks if at least one object version was found. If not, then files are not retreived (otherwise there is seg fault). Please, let me know if you want any additional changes

Barthelemy:
@wiechula @lauraser We need to release QC this morning and this is the last release before the freeze. We can do adhoc patches afterward yet we should aim at merging this PR this morning. 
Could you approve and merge ?
thank you

wiechula (approved):

wiechula:
@Barthelemy , done.

PR #2433, opened 2024-09-23
Common: improve layout of 1-D reference plot comparisons

aferrero2707:
The new layout puts the ratio plot above the two superimposed histograms, and avoids the presence of a transparent pad above the histograms. This restores the possibility to see the bin values by hovering the mouse pointer over the histogram, as well as of zooming the horizontal axis with the mouse.

Here is an example of the new plots arrangement, for different aspect ratios of the canvas:
Screenshot 2024-09-23 at 22 41 56


aferrero2707:
@Barthelemy @knopers8 this is a visual improvement that was requested after some detectors started to use the reference comparison task.

Barthelemy (approved):

PR #2437, opened 2024-09-24
[QC-1232] Improve log message when kafka broker url is missing

justonedev1:
better error handling when kafkaBrokers are missing

### DIFF
--- a/Framework/src/KafkaPoller.cxx
+++ b/Framework/src/KafkaPoller.cxx
@@ -6,6 +6,7 @@
 #include "proto/events.pb.h"
 #include <chrono>
 #include <ranges>
+#include <stdexcept>
 
 namespace o2::quality_control::core
 {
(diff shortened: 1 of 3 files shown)


Framework/src/KafkaPoller.cxx
```diff
@@ -112,10 +113,21 @@ bool end_of_run::isValid(const events::Event& event, const std::string& environm
 
 kafka::Properties createProperties(const std::string& brokers, const std::string& groupId)
 {
-  return { { { "bootstrap.servers", { brokers } },
-             { "group.id", { groupId } },
-             { "enable.auto.commit", { "true" } },
-             { "auto.offset.reset", { "latest" } } } };
+  if (brokers.empty()) {
+    constexpr std::string_view message{ "You are trying to start KafkaPoller without any brokers" };
```
Barthelemy, line 117: why use a `string_view` here ?

Framework/src/Triggers.cxx
```diff
@@ -68,8 +68,25 @@ std::string createKafkaGroupId(std::string_view prefix, std::string_view detecto
   return groupId;
 }
 
+bool checkKafkaParams(const std::string& kafkaBrokers, const std::string& topic, const std::string_view triggerTypeLogId)
+{
```
Barthelemy, line 72: genuine question: why don't you use a string_view for the first two arguments ? 
I understand that the parameters you will give later in SOR are strings, but from the point of view of the interface, wouldn't it be better to have something uniform ?

Framework/src/Triggers.cxx
```diff
@@ -68,8 +68,25 @@ std::string createKafkaGroupId(std::string_view prefix, std::string_view detecto
...
+  if (kafkaBrokers.empty()) {
+    ILOG(Error, Support) << "You are tring to create " << triggerTypeLogId << " trigger using Kafka without any brokers, fill config value 'kafkaBrokersUrl'" << ENDM;
+    return false;
+  }
+  if (topic.empty()) {
+    ILOG(Error, Support) << "You are tring to consume empty Kafka topic from " << triggerTypeLogId << " trigger, fill config value 'kafkaTopic'" << ENDM;
```
knopers8, line 78:
```suggestion
    ILOG(Error, Support) << "You are tring to consume empty Kafka topic from '" << triggerTypeLogId << "' trigger, fill config value 'kafkaTopic'" << ENDM;
```
Please use single quotes around `triggerTypeLogId`, so if it's empty, it's easier to notice it in the log.

Framework/src/Triggers.cxx
justonedev1: you are right, all of these should be the same. Thanks for the remark
I generally prefer `std::string_view` over `const std::string&` because it is non owning and non allocating and read only class, so it 1) implies the usage, 2) allows you to use the function (generally speaking) allocating something/creating temporary string, 3) string_view can be used easily with wider range of classes: `std::string`, `const char*`, `std::vector<char>`, ... But in this case the choice does not really matter...

Barthelemy: ok !

Barthelemy (approved):

PR #2438, opened 2024-09-25
Common: add support for variable bins in THxRatio

aferrero2707:
The added support for variable bins requires two changes:
* implementation of the corresponding constructors
* the histogram axes need to be updated using the `TAxis::Copy()` function


aferrero2707:
@Barthelemy @knopers8 @justonedev1 this follows some requests from detector experts, and did not require substantial changes to the code.

knopers8 (approved):
thank you

PR #2440, opened 2024-09-25
Common: support varibale bins in reference comparator output

aferrero2707:
The changes allow to properly display histograms with variable bin sizes in the output of the reference comparator task.

Here is an example of variable bin size histogram compared to reference:
image


aferrero2707:
@Barthelemy @knopers8 @justonedev1 again a requests from detector experts, only requiring rather trivial changes.

knopers8 (approved):

aferrero2707:
@knopers8 I had to change a bit the code, because the use of the `Clone()` function was giving empty plots when the source histograms were of THRatio type. Not clear why, but anyhow the updated code is also quite trivial.

PR #2441, opened 2024-09-25
ZDC: Add summary and correlection plot

cpuggion84:
Add summary plot about;
amplitude TDC
Peak 1n
Peak 1p
fix correletion plot zp zem


Barthelemy (approved):
I checked that there were no obvious mistakes but I couldn't make sure whether the logic is sound as it is quite convoluted.

PR #2443, opened 2024-09-30
[QC-1201] Replace the merged object with the input if it has higher run number

justonedev1:
is this the way how to do the ticket goal correctly? I have yet to add tests, so I am making draft first, so you can tell me if this is not it

### DIFF
--- a/Framework/src/MonitorObjectCollection.cxx
+++ b/Framework/src/MonitorObjectCollection.cxx
@@ -46,16 +46,32 @@ void MonitorObjectCollection::merge(mergers::MergeInterface* const other)
       if (!otherMO || !targetMO) {
         throw std::runtime_error("The target object or the other object could not be casted to MonitorObject.");
       }
-      if (!reportedMismatchingRunNumbers && targetMO->getActivity().mId != otherMO->getActivity().mId) {
+
+      if (otherMO->getActivity().mId > targetMO->getActivity().mId) {
+        ILOG(Error, Ops) << "The run number of the input object '" << otherMO->GetName() << "' ("
+                         << otherMO->getActivity().mId << ") "
+                         << "is higher than the one of the target object '"
+                         << targetMO->GetName() << "' (" << targetMO->getActivity().mId
+                         << "). Replacing the merged object with input, "
+                         << "but THIS SHOULD BE IMMEDIATELY ADDRESSED IN PRODUCTION. "
+                         << "QC objects from other setups are reaching this one."
+                         << ENDM;
+        otherMO->Copy(*targetMO);
+        continue;
+      }
+
+      if (!reportedMismatchingRunNumbers && otherMO->getActivity().mId < targetMO->getActivity().mId) {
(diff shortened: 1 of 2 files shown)


Framework/src/MonitorObjectCollection.cxx
```diff
@@ -46,16 +45,19 @@ void MonitorObjectCollection::merge(mergers::MergeInterface* const other)
...
-                         << "does not match the run number of the target object '"
+                         << "is higher than the one of the target object '"
                          << targetMO->GetName() << "' (" << targetMO->getActivity().mId
-                         << "). Trying to continue, but THIS SHOULD BE IMMEDIATELY ADDRESSED IN PRODUCTION. "
-                         << "QC objects from other setups are reaching this one. Will not report more mismatches in this collection."
+                         << "). I am switching these objects, but THIS SHOULD BE IMMEDIATELY ADDRESSED IN PRODUCTION. "
+                         << "QC objects from other setups are reaching this one."
                          << ENDM;
-        reportedMismatchingRunNumbers = true;
+        otherMO->Copy(*targetMO);
```
knopers8, line 57: regarding the `Copy`, yes, i think this is the way.

Framework/src/MonitorObjectCollection.cxx
```diff
@@ -46,16 +45,19 @@ void MonitorObjectCollection::merge(mergers::MergeInterface* const other)
       if (!otherMO || !targetMO) {
         throw std::runtime_error("The target object or the other object could not be casted to MonitorObject.");
       }
-      if (!reportedMismatchingRunNumbers && targetMO->getActivity().mId != otherMO->getActivity().mId) {
+
+      if (otherMO->getActivity().mId > targetMO->getActivity().mId) {
```
knopers8, line 49: yes, but we should still produce a warning and ignore an input object if it's run number is lower, and as before, avoid flooding the IL for the rest of `otherCollection`. In this version of the code, you would simply merge such input anyway.

Framework/src/MonitorObjectCollection.cxx
```diff
@@ -46,16 +45,19 @@ void MonitorObjectCollection::merge(mergers::MergeInterface* const other)
...
         ILOG(Error, Ops) << "The run number of the input object '" << otherMO->GetName() << "' ("
                          << otherMO->getActivity().mId << ") "
-                         << "does not match the run number of the target object '"
+                         << "is higher than the one of the target object '"
                          << targetMO->GetName() << "' (" << targetMO->getActivity().mId
-                         << "). Trying to continue, but THIS SHOULD BE IMMEDIATELY ADDRESSED IN PRODUCTION. "
-                         << "QC objects from other setups are reaching this one. Will not report more mismatches in this collection."
+                         << "). I am switching these objects, but THIS SHOULD BE IMMEDIATELY ADDRESSED IN PRODUCTION. "
```
knopers8, line 54: "I am switching these objects" is not clear to me, how about "Replacing the merged object with input"?

PR #2444, opened 2024-10-01
Common: correctly handle integer histograms in reference comparator

aferrero2707:
The reference comparator code is extended to handle histograms with integer values, by internally converting them to floating-point such that the normalization and ratios are computed correctly


aferrero2707:
@knopers8 this is the extension we discussed for the ZDC plots.

knopers8 (approved):
Thank you!

PR #2449, opened 2024-10-04
THRatio: fix again update of histograms with bin labels

aferrero2707:
A direct copy of the axis objects from the numerator to the ratio histograms, that was introduced in a [recent commit](https://github.com/AliceO2Group/QualityControl/pull/2438/commits/42a5a592b790c1fbe86c13cabdba151df57b096e), breaks the preservation of the bin labels that are eventually set in the ratio histogram.

The code in this commit replaces the copy with an explicit update of the number and limits of the bins, such that it works also if the numerator is filled with `TH1::kCanRebin` or `TH1::SetCanExtend()` set, and therefore the axis range can change after the histogram creation.

### DIFF
--- a/Modules/Common/include/Common/TH1Ratio.inl
+++ b/Modules/Common/include/Common/TH1Ratio.inl
@@ -218,7 +218,11 @@ void TH1Ratio<T>::update()
   }
 
   T::Reset();
-  mHistoNum->GetXaxis()->Copy(*T::GetXaxis());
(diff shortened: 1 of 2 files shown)


aferrero2707:
@Barthelemy @knopers8 @justonedev1 this PR will be followed by another one with additional unit tests that will catch this kind of regressions, and hopefully much more.

Barthelemy (approved):

PR #2450, opened 2024-10-07
refactor MonitorObject and adding Copy capabilities

justonedev1:
when working on QC-1201 and implementing Copy function/ctor/operator I found that Monitor Object is weirdly written, moreover there are possiblities, where the class can leak memory. So I refactored it a bit while adding Copy function, ctor and operator.

### DIFF
--- a/Framework/include/QualityControl/MonitorObject.h
+++ b/Framework/include/QualityControl/MonitorObject.h
@@ -146,7 +146,7 @@ class MonitorObject : public TObject
   void releaseObject();
   void cloneAndSetObject(const MonitorObject&);
 
-  ClassDefOverride(MonitorObject, 12);
+  ClassDefOverride(MonitorObject, 13);
 };
 
 } // namespace o2::quality_control::core
--- a/Framework/include/QualityControl/MonitorObject.h
+++ b/Framework/include/QualityControl/MonitorObject.h
@@ -50,12 +50,12 @@ class MonitorObject : public TObject
   /// Destructor
   ~MonitorObject() override;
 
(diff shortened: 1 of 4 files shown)


Framework/include/QualityControl/MonitorObject.h
```diff
@@ -141,6 +143,9 @@ class MonitorObject : public TObject
   // tells Merger to create an object with data from the last cycle only on the side of the complete object
   bool mCreateMovingWindow = false;
 
+  void releaseObject();
+  void cloneAndSetObject(const MonitorObject&);
+
   ClassDefOverride(MonitorObject, 12);
```
Barthelemy, line 149: you must increase the class version otherwise ROOT can't do the schema evolution.
not super important now that we don't store MO in the QCDB but still...

Barthelemy (commented):
Your hands must be dirty and dusty after touching such an antique piece of code !

I agree with all this. 
Not sure if we should merge it now or wait post-HI.

justonedev1:
the thing is that I need this PR for [QC-1201](https://its.cern.ch/jira/browse/QC-1201). But I don't know how necessary that ticket is... probably not much for HI. So I am 100% fine to just wait after HI...

knopers8 (approved):
As I was talking to you the other day, I am not very happy about having a unique_ptr which does not always own the pointer, but I don't see a better way to do this. And I support the intention of the changes. Thank you for the initiative!

justonedev1:
@ktf Can you please take a look at the reason why the build checks are failing?

PR #2461, opened 2024-10-28
[QC-1227] Get beam type from PDP parameter

Barthelemy:


justonedev1 (commented):
Are the string values from `pdp_beam_type` defined somewhere? I saw them mentioned by Andrea but without any link

Barthelemy:
short answer: I don't know. 
@aferrero2707 do you know where they are defined ? 
or perhaps @ktf knows ?

ktf:
No idea. @shahor02 maybe?

shahor02:
`pdp_beam_type` are defined in the ECS interface and converted by the ODC to BEAMTYPE env.var, but I don't know its source code.

Barthelemy:
Ok, thank you for the info. I conclude that the format of these values is defined somewhere in ODC.

Framework/src/runnerUtils.cxx
```diff
@@ -91,6 +91,20 @@ std::string computeStringActivityField(framework::ServiceRegistryRef services, c
...
+{
+  // convert the beam type received from pdp into the format we use in flp/ecs
+  if (pdpBeamType == "pp") {
+    return "PROTON-PROTON";
+  } else if (pdpBeamType == "PbPb") {
+    return "Pb-Pb";
+  } else if (pdpBeamType == "pPb") {
+    return "Pb-PROTON";
+  } else {
+    return "";
```
justonedev1, line 104: Maybe add some warning, that we met value that we did not expect?

Barthelemy: yep

Barthelemy: done, I let you have a look

justonedev1 (approved):

PR #2465, opened 2024-10-30
Fix CompletionPolicy for BookkeepingQualitySink

knopers8:
A label for matching the correct CompletionPolicy in BookkeepingQualitySink was missing. This was causing it to wait indefinitely for matching inputs for the same timeslice if QOs from different Data Processors were requested. The change allows BookkeepingQualitySink to consume the inputs immediatelly and pass them to any other process which may also need them.


justonedev1 (approved):
thank you

knopers8:
The command below with the provided configuration file reproduces the issue observed by Robert. With this fix, the reproducer behaves OK.

```
o2-qc-run-producer | o2-qc --config json://home/pkonopka/alice/QualityControl/Framework/basic.json
```

[basic.json](https://github.com/user-attachments/files/17570260/basic.json)

PR #2469, opened 2024-11-05
Fixes for QC flags in async QC

knopers8:


knopers8:
One more to come tomorrow, thus WIP.

knopers8:
Well, I think it will be actually better to review them separately, so this PR is ready for review @Barthelemy

Barthelemy (approved):

PR #2470, opened 2024-11-06
ITS Track Check returns Null Quality by default

knopers8:
Initializing Quality result with 0 had an effect of producing a Quality without a name and with level 0, which is treated as "not worse than Good". This does not seem like an intentional behaviour, so I am proposing to use the default Null Quality instead. Precautions will be taken in the framework, so that it cannot be done anymore.


Barthelemy (approved):

knopers8:
@iravasen @IsakovAD is this OK with you?

IsakovAD (approved):

PR #2472, opened 2024-11-06
QC-1246 Explicit Quality constructor to avoid implicit casts from int

knopers8:


knopers8:
It will be green once https://github.com/AliceO2Group/QualityControl/pull/2470 is merged.

Barthelemy (approved):

PR #2481, opened 2024-11-20
QC-1248 Document QO->Flag conversion

knopers8:


knopers8:
FYI @JianLIUhep . These are basically my slides about QC flag conversion in form of markdown documentation. Please let me know if there is anything unclear in this doc from your point of view and feel free to refer detector experts to it once the PR is merged.

justonedev1 (commented):
I have a small issue with the images. As a casual Dark theme enjoyer, I really cannot see description and axis on the images, because you created the images with transparent background. I think, that I might not be the only one who might encounter this problem.
image

knopers8:
Yes, I was also not happy about it and I investigated what could be done. In Github you can have two versions of images for light and dark mode, but that requires generating the alternative ones and I wasn't motivated enough. I also could not manage to enforce a background in SVG through libre office. Also, github does not have a way to enforce a white background for transparent images.

So I might still motivate myself to generate dark mode alternatives, seeing that I am not the only one who would suffer from this.

justonedev1:
Let @Barthelemy decide :)

knopers8:
> Let @Barthelemy decide :)

Never!

Barthelemy:
Hi, 

Is the first image ok for you @justonedev1 ? I added a white background via the `style` attribute of the `svg` tag. `

justonedev1:
All of them, I just did printscreen of the first one

knopers8:
> Hi,
> 
> Is the first image ok for you @justonedev1 ? I added a white background via the `style` attribute of the `svg` tag. `

It works for me (I also use dark mode).

Barthelemy:
ok, I will do the others

Barthelemy:
oh you did it, thank you !

PR #2482, opened 2024-11-21
[ITS] Anomalous clusters plots

nicolovalle:
This PR adds two sets of 2D plots published by the cluster task:

- Number of long clusters vs chip ID (3 plots, for the IB layers)
- Number of hits as sum of the clusters size for those events with at least one long cluster, vs chip ID (3 plots, for the IB layers)

Tested locally

### DIFF
--- a/Modules/ITS/src/ITSClusterTask.cxx
+++ b/Modules/ITS/src/ITSClusterTask.cxx
@@ -155,8 +155,8 @@ void ITSClusterTask::monitorData(o2::framework::ProcessingContext& ctx)
     const auto& ROF = clusRofArr[iROF];
     const auto bcdata = ROF.getBCData();
     int nClustersForBunchCrossing = 0;
-    int nLongClusters[ChipBoundary[3]] = {};
-    int nHitsFromClusters[ChipBoundary[3]] = {}; // only IB is implemented at the moment
+    int nLongClusters[ChipBoundary[NLayerIB]] = {};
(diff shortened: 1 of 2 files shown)


Modules/ITS/src/ITSClusterTask.cxx
```diff
@@ -153,34 +155,44 @@ void ITSClusterTask::monitorData(o2::framework::ProcessingContext& ctx)
...
+
     for (int icl = ROF.getFirstEntry(); icl < ROF.getFirstEntry() + ROF.getNEntries(); icl++) {
 
       auto& cluster = clusArr[icl];
       auto ChipID = cluster.getSensorID();
       int ClusterID = cluster.getPatternID(); // used for normal (frequent) cluster shapes
       int lay, sta, ssta, mod, chip, lane;
 
-      if (ChipID != ChipIDprev) {
+      if (ChipID != -1) { // TODO: is this needed?
```
IsakovAD, line 168: Idea of this thing was to avoid re-calculation of the mod/lane/ChipId each new cluster. If I am not wrong,  request to mGeom could take quite time

IsakovAD (commented):
Hello @nicolovalle, 

Thank you very much for such prompt implementation.  I think that it is better to keep old way how we calculated information about chip, or I am missing something?

nicolovalle:
> Hello @nicolovalle,
> 
> Thank you very much for such prompt implementation. I think that it is better to keep old way how we calculated information about chip, or I am missing something?

Hi, I left a comment inline.. to my understanding that was un unused variable

IsakovAD (approved):

PR #2484, opened 2024-12-04
[ZDC] bug fix and plot addition

stefan-zugravel:
-> FIXED - bug that hard-coded two configuration parameter
-> ADDED - ADC and TDC plots with a cut
-> ADDED - centroid plot with cut
-> ADDED - axis label on SIGNAL plots 
-> ADDED - rate histogram for trending plot


Barthelemy (approved):

Modules/ZDC/src/ZDCRawDataTask.cxx
```diff
@@ -952,12 +980,49 @@ bool ZDCRawDataTask::addNewHisto(std::string type, std::string name, std::string
...
+        fNameHisto.push_back(name);
+        h1d.histo = new TH1F(hname, htit, fNumBinX, fMinBinX, fMaxBinX);
+        h1d.condHisto.push_back(condition);
+        ih = (int)fMatrixHistoCounts_a[mod][ch].size();
+        fMatrixHistoCounts_a[mod][ch].push_back(h1d);
+
+        if (ih < (int)fMatrixHistoCounts_a[mod][ch].size()) {
+          getObjectsManager()->startPublishing(fMatrixHistoCounts_a[mod][ch].at(ih).histo);
+          try {
+            getObjectsManager()->addMetadata(fMatrixHistoCounts_a[mod][ch].at(ih).histo->GetName(), fMatrixHistoCounts_a[mod][ch].at(ih).histo->GetName(), "34");
```
Barthelemy, line 997: what is this ?

stefan-zugravel: It is the same code used in line 965.
We want to use the histograms of type COUTSA to generate a trending plot of the rate during a run.
The difference between COUTS and COUTSA is the the second want is zeroed at each cycle and thus provides an estimate of the  instantaneous rate and not the average one.

stefan-zugravel:
Resolved the clang-format issue and updated the branch.

PR #2485, opened 2024-12-16
ITSThresholdCalibrationTask.h: change Zmax for THR plots

iravasen:
@IsakovAD Trivial change to improve plot visibility for shifters. Example here

thr


IsakovAD (approved):
Thank you for pinging me. I've overlooked that PR :)

PR #2486, opened 2024-12-17
MCH: removed custom code for reference plots

aferrero2707:
* remove the custom MCH code for loading reference plots and comparing with actual data

* store directly histograms instead of TCanvas objects for several plots


Barthelemy (approved):
This is always very pleasant : 
Capture d’écran 2025-01-07 à 10 45 39

PR #2495, opened 2025-01-20
ITS/FEE: improved time granularity of payload plot

IsakovAD:
Changed filling of Payload size plot from each TF to each orbit


IsakovAD:
Hello @knopers8 

In all of my recent PRs, I have an issue with `Git command for package 'TBB' failed.`. Do you have any cue how I can fix that?

knopers8:
> In all of my recent PRs, I have an issue with Git command for package 'TBB' failed.. Do you have any cue how I can fix that?

@singiamtel @ktf it seems like macOS-arm builds are stuck (`Rechecking since Thu 15:28 CET on f05bd128 `), we still have results from before your [fix for TBB](https://github.com/alisw/alidist/pull/5747). Would you have a way to unblock them?

knopers8:
> this builder had some corrupted cache with the old repo. We've cleared it and restarted the jobs, they should be picked up by the builder soon

thanks @singiamtel !

iravasen (approved):

iravasen:
@Barthelemy also this one has a non-related failure

PR #2497, opened 2025-01-24
 ITS/ChipStatus: fix in error message text

IsakovAD:
Added shift in the error message about missing stave


iravasen (approved):

iravasen:
@Barthelemy the failure is not related to this PR, I think we can merge this.

PR #2502, opened 2025-01-28
Drop tautological comparison

ktf:
A pointer to a reference cannot be nullptr. If it is, whatever created the
reference in the first place has already crashed. If this is to be protected,
one should return the pointer and check that.


ktf:
@Barthelemy can this be merged and a new release tagged? It's breaking the CI.

Barthelemy:
@ktf yes

PR #2511, opened 2025-02-21
ITS/FEE: better resource handling for payload

IsakovAD:
- optimized saving of payload size for "precise calculation"
- added option for old-type of payload
- optimization of code: removed not-needed if-case
- new histogram with number of empty payloads per FEE


IsakovAD:
Hello @iravasen 

Could you please have a look at this PR?

iravasen (approved):

PR #2512, opened 2025-02-24
QC-1213 A QCDB restart must not stop the run

justonedev1:
The crash was caused by missing exception handling in CcdbDatabase.cxx when we receive something else than expected json as `boost::property_tree::read_json` throws an exception in case it fails. 

This fix was tested on staging. I also tried to create some unit test that catches this, but I wasn't really succesfull is it is not that easy to make ccdb able to initiate but make unresponsive automatically, which is why it took so long to create PR.

### DIFF
--- a/Framework/src/CcdbDatabase.cxx
+++ b/Framework/src/CcdbDatabase.cxx
@@ -37,9 +37,9 @@
 #include <chrono>
 #include <sstream>
 #include <filesystem>
-#include <unordered_set>
 // boost
 #include <boost/property_tree/json_parser.hpp>
+#include <boost/property_tree/json_parser/error.hpp>
 #include <boost/property_tree/ptree.hpp>
 #include <boost/foreach.hpp>
 #include <utility>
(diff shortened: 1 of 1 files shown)


Framework/src/CcdbDatabase.cxx
```diff
@@ -568,7 +568,10 @@ boost::property_tree::ptree CcdbDatabase::getListingAsPtree(const std::string& p
   std::stringstream listingAsStringStream{ getListingAsString(pathWithMetadata.str(), "application/json", latestOnly) };
 
   boost::property_tree::ptree listingAsTree;
-  boost::property_tree::read_json(listingAsStringStream, listingAsTree);
+  try {
+    boost::property_tree::read_json(listingAsStringStream, listingAsTree);
+  } catch (const boost::property_tree::json_parser::json_parser_error&) {
```
knopers8, line 573: Could you add an error with support level, indicating that the listing did not have correct JSON format due to possibly malformed or missing reply from QCDB? The same comment for the other `catch` in the PR.

knopers8 (approved):

PR #2513, opened 2025-02-25
GLO: Add K0s plots

f3sch:
Adding fitting and trending of K0s for GLO. Other changes will follow but first need modifications in O2.

### DIFF
--- a/Modules/GLO/glo-itstpc-mtch-qcmn-test.json
+++ b/Modules/GLO/glo-itstpc-mtch-qcmn-test.json
@@ -10,11 +10,12 @@
         "maxObjectSize": "20000000"
       },
       "Activity": {
-        "number": "48",
-        "type": "2"
+        "number": "",
+        "type": "",
+        "beamType": "PbPb"
       },
       "monitoring": {
-        "url": "infologger:///debug?qc"
+        "url": "no-op://"
       },
       "consul": {
         "url": ""
@@ -25,6 +26,10 @@
       "infologger": {
         "filterDiscardDebug": "false",
         "filterDiscardLevel": "0"
+      },
+      "postprocessing": {
+        "periodSeconds": 0.01,
+        "matchAnyRunNumber": "true"
       }
     },
     "tasks": {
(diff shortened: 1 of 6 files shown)


Modules/GLO/src/ITSTPCMatchingTask.cxx
```diff
@@ -150,42 +134,86 @@ void ITSTPCMatchingTask::endOfCycle()
...
+
+      mK0sCycle.reset();
+      mK0sCycle.reset(dynamic_cast<TH3F*>(k0s->Clone("mK0sMassVsPtVsOcc_Cycle")));
+      if (!mK0sCycle) {
+        ILOG(Fatal) << "Could not retrieve k0s histogram for current cycle" << ENDM;
       }
-      if (!ratio->getDen()->Add(eff->GetTotalHistogram())) {
-        ILOG(Error) << "Add operation for denominator histogram of " << name << " failed; efficiency will be skewed" << ENDM;
+      if (!mK0sIntegral) {
+        mK0sIntegral.reset(dynamic_cast<TH3F*>(k0s->Clone("mK0sMassVsPtVsOcc_Integral")));
```
knopers8, line 193: ROOT sometimes causes memory issues if there are two histograms with the same name in memory. A workaround is to do what you did a few lines before:
```
      mK0sCycle.reset();
      mK0sCycle.reset(dynamic_cast<TH3F*>(k0s->Clone("mK0sMassVsPtVsOcc_Cycle")));
```
you may want to do the same here.

Modules/GLO/include/GLO/ITSTPCMatchingTask.h
```diff
@@ -50,9 +56,44 @@ class ITSTPCMatchingTask final : public TaskInterface
...
   std::unique_ptr<common::TH1FRatio> mEffPhi;
+
+  bool mDoPtTrending{ false };
+  float mTrendingBinPt{ 1.0 };
+  std::unique_ptr<TGraph> mTrendingPt;
+
+  bool mDoK0s{ false };
+  static constexpr auto mMassK0s{ o2::constants::physics::MassK0Short };
+  std::unique_ptr<TH3F> mK0sCycle;
+  std::unique_ptr<TH3F> mK0sIntegral;
```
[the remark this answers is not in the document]
f3sch: per histogram it should be `(10 + 2) * (100 + 2) * (6 + 2) * sizeof(float) * 2 = 76 kB`

Modules/GLO/src/ITSTPCMatchingTask.cxx
```diff
@@ -150,42 +134,86 @@ void ITSTPCMatchingTask::endOfCycle()
...
+        ratio->setHasBinominalErrors();
+        ratio->update();
+      };
+
+      // Pt
+      makeRatio(mEffPt, mMatchITSTPCQC.getFractionITSTPCmatch(gloqc::MatchITSTPCQC::ITS));
+      getObjectsManager()->startPublishing(mEffPt.get(), PublicationPolicy::Once);
+      getObjectsManager()->setDefaultDrawOptions(mEffPt->GetName(), "logx");
+      if (mDoPtTrending) {
+        mTrendingPt->AddPoint(mNCycle, mEffPt->GetBinContent(mEffPt->FindBin(mTrendingBinPt)));
```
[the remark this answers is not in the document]
f3sch: I am not sure as-well but I did not manage to make the post processing work.

Modules/GLO/src/ITSTPCMatchingTask.cxx
f3sch: but `mK0sIntegral` is only needed to be initialized once, after that I use `Reset()` and `Add(k0s)` to spare some memory churn, so there should never be two histograms with the same name, or am I misunderstanding something?

f3sch:
Thanks @knopers8 for the review btw!

Modules/GLO/include/GLO/ITSTPCMatchingTask.h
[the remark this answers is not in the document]
f3sch: technically right now I could also just not publish them, since nobody will look at a 3D histogram.

Modules/GLO/src/ITSTPCMatchingTask.cxx
knopers8: no, then indeed it should be fine

Modules/GLO/include/GLO/ITSTPCMatchingTask.h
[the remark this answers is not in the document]
knopers8: 76kB is absolutely fine, I was fearing >10MB

f3sch:
needs [#14035](https://github.com/AliceO2Group/AliceO2/pull/14035).

knopers8:
@f3sch is it ready for merging?

Modules/GLO/src/ITSTPCMatchingTask.cxx
```diff
@@ -130,6 +83,7 @@ void ITSTPCMatchingTask::startOfActivity(const Activity& activity)
 {
   ILOG(Debug, Devel) << "startOfActivity " << activity.mId << ENDM;
   mMatchITSTPCQC.reset();
+  mIsPbPb = activity.mBeamType == "Pb-Pb";
```
knopers8, line 86:
```suggestion
  mIsPbPb = activity.mBeamType == "PbPb";
```
Use `PbPb`, as we will move from LHC-provided beam type to the one provided by `pdp_beam_type`. See https://github.com/AliceO2Group/QualityControl/pull/2540

Modules/GLO/src/ITSTPCMatchingTask.cxx
```diff
@@ -150,41 +104,130 @@ void ITSTPCMatchingTask::endOfCycle()
...
+      //   mPVITSCycle->Reset();
+      //   mPVITSCycle->Add(h, mPVITSCycle.get(), 1., -1.);
+      //   mPVITSIntegral->Reset();
+      //   mPVITSIntegral->Add(h);
+
+      //   getObjectsManager()->startPublishing(mPVITSCycle.get(), PublicationPolicy::Once);
+      //   getObjectsManager()->startPublishing(mPVITSIntegral.get(), PublicationPolicy::Once);
+      //   getObjectsManager()->startPublishing(mPVITSCycle->ProfileX(), PublicationPolicy::Once);
+      //   getObjectsManager()->startPublishing(mPVITSIntegral->ProfileX(), PublicationPolicy::Once);
+      // }
```
knopers8, line 175: Are you sure you need this commented out code here?

Modules/GLO/glo-itstpc-mtch-qcmn-test.json
```diff
@@ -33,35 +32,33 @@
         "className": "o2::quality_control_modules::glo::ITSTPCMatchingTask",
         "moduleName": "QcGLO",
         "detectorName": "GLO",
-        "cycleDurationSeconds": "60",
+        "cycleDurationSeconds": "1",
```
knopers8, line 35: FYI, the framework will cap this at `10` anyway. In any case, please do not use less than `60` in production.`

knopers8 (commented):
I have some small suggestions that should be fast to address, otherwise it seems good to go.

Modules/GLO/src/ITSTPCMatchingTask.cxx
f3sch: thanks!
N.B. an enum or some getter `isPbPb` provided by the activity would be nicer than a string comp.

f3sch: would decouple this for users as-well

knopers8: I agree.

PR #2514, opened 2025-02-28
Use EOR timestamp from ECS for object validity end in synchronous QC

knopers8:
QC-1270

With OCTRL-987, we should be able to retrieve EOR timestamp also on EPNs.


Barthelemy (approved):
thanks, that's good

PR #2516, opened 2025-03-04
Script to remove version not matching periods

Barthelemy:
Script to remove all the versions in the given path that don't match the given list of periodNames

### DIFF
--- a/Framework/script/RepoCleaner/qcrepocleaner/o2-qc-repo-delete-versions-not-in-periods
+++ b/Framework/script/RepoCleaner/qcrepocleaner/o2-qc-repo-delete-versions-not-in-periods
@@ -10,7 +10,7 @@ import dryable
 def parse_args():
     """Parse the arguments passed to the script."""
     logging.info("Parsing arguments")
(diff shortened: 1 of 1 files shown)


Framework/script/RepoCleaner/qcrepocleaner/o2-qc-repo-delete-versions-not-in-periods
```diff
@@ -0,0 +1,111 @@
...
+import argparse
+
+from qcrepocleaner import binUtils
+from qcrepocleaner.Ccdb import Ccdb
+import dryable
+
+def parse_args():
+    """Parse the arguments passed to the script."""
+    logging.info("Parsing arguments")
+    parser = argparse.ArgumentParser(description='Remove all the versions in the given path that don\'t match the given list of periodNames ')
```
knopers8, line 13: perhaps you could specify explicitly what will happen to versions without `periodName`.

Barthelemy: yes indeed.

PR #2524, opened 2025-03-14
QC-1161 remove unused code for quality flag support

knopers8:


justonedev1 (approved):
It looks ok, but why don't we need this code? Is it because we are using kafka to propagating the flags?

knopers8:
> It looks ok, but why don't we need this code? Is it because we are using kafka to propagating the flags?

Not even Kafka, but a gRPC call to BKP that you added.

A few years ago we assumed we would keep the flags in QCDB, but we switched to BKP.

Barthelemy (approved):
I love seeing code being removed. Thanks !

PR #2527, opened 2025-03-21
[1272] Add missing beam type (cosmic)

Barthelemy:


Barthelemy:
force merging as we need it for the patch

PR #2530, opened 2025-04-01
ITS/Calibration: better handling of wrong run types and Pixel scans

IsakovAD:
- Added new error messages with QC Task is not properly set up in terms of run type: expected values: THR, ITHR, VCASN, Pixel, TOT
- Pixel scan plots now correctly filled when selected Pixel scan type


iravasen (approved):

iravasen:
Hi Artem, the code seems complete to me but this PR is still in WIP. Do you miss any feature?

iravasen:
I mostly checked what was disussed but not the code in detail

PR #2532, opened 2025-04-04
MCH: make LastCycle and trending plots optional

aferrero2707:
The LastCycle plots have been introduced as a workaround to mimic the moving windows, before they were implemented in the framework.

Since the moving windows are now the proper way to produce plots for a single QC cycle, the processing of LastCycle plots is made optional and disabled by default.

The creation of trending plots is also made optional and disabled by default. It should be enabled for the tasks that process the moving window outputs.


Barthelemy (approved):
thanks !

PR #2535, opened 2025-04-14
Don't use deprecated boost functionality, be compatible to boost 1.87

davidrohr:


Barthelemy (approved):
Thank you, makes sense

PR #2537, opened 2025-04-15
Update ubuntu for clang-format

Barthelemy:


Barthelemy:
Necessary as ubuntu 20.04 is too old.

knopers8 (approved):

PR #2538, opened 2025-04-16
CTP: Change in RawDataQcTask

lhusova:
Addition parameter to turn off the consistency check. @lietava, please have a look.

### DIFF
--- a/Modules/CTP/include/CTP/RawDataQcTask.h
+++ b/Modules/CTP/include/CTP/RawDataQcTask.h
@@ -69,6 +69,7 @@ class CTPRawDataReaderTask final : public TaskInterface
   long int mTimestamp;
   std::string classNames[nclasses];
   int mIndexMBclass = -1; // index for the MB ctp class, which is used as scaling for the ratios
+  bool mConsistCheck = 0;
 };
 
 } // namespace o2::quality_control_modules::ctp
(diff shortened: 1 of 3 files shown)


lietava:
@lhusova ok for me

lietava:
Hi @Barthelemy , @knopers8 , please, can this be merged asap as we need to deploy it in production. Cheers, Roman.

Barthelemy:
@lietava When you say to deploy it in production, do you mean at P2 ? because merging will not do that, we would need to prepare a patch.

Barthelemy (approved):

lietava:
@Barthelemy thanks. Yes, I spoke with pippo - he agrees. Who van do the patch, please ?

Barthelemy:
@lietava I will prepare the patch, but @vascobarroso has to approve the installation. I feel that installing something the day before a long weekend is really calling for trouble. Is it preventing you from running ?

Barthelemy:
https://github.com/AliceO2Group/QualityControl/releases/tag/v1.171.1

lietava:
Hi @Barthelemy , thanks for tag. 
Then to your comment: there is new functionality which seems to have bug which is not understood. This fix should enable to switch it on/off new functionality before bug is understood. Otherwise we would run without ctp qc  which is also case if the bug is here in this PR. So it is for @vascobarroso to decide. I would vote for installation before the first stable beams.

Modules/CTP/include/CTP/RawDataQcTask.h
```diff
@@ -69,6 +69,7 @@ class CTPRawDataReaderTask final : public TaskInterface
   long int mTimestamp;
   std::string classNames[nclasses];
   int mIndexMBclass = -1; // index for the MB ctp class, which is used as scaling for the ratios
+  bool mConsistCheck = 0;
```
lietava, line 72: @lhusova  This variable is not used - can be removed at next modifications.

PR #2540, opened 2025-04-17
[QC-1243]: Fix consistency in beam type names

justonedev1:
This ticket was created to address ticket [QC-1243](https://its.cern.ch/jira/browse/QC-1243). The script is meant to be run on STG and PROD after installation of flp-suite with the changes. 

However please don't merge this, I am not 100% sure that everything is working and I wanted to trigger our CI/CD tests.

### DIFF
--- a/doc/Advanced.md
+++ b/doc/Advanced.md
@@ -1295,10 +1295,10 @@ The new syntax is
               "myOwnKey1": "myOwnValue1b",
               "myOwnKey2": "myOwnValue2b"
             },
-            "PROTON-PROTON": {
+            "pp": {
               "myOwnKey1": "myOwnValue1c"
             },
-            "Pb-Pb": {
+            "PbPb": {
               "myOwnKey1": "myOwnValue1d"
             }
           },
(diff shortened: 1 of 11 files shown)


doc/Advanced.md
```diff
@@ -1310,7 +1310,7 @@ The new syntax is
 ```
 It allows to have variations of the parameters depending on the run and beam types. The proper run types can be found here: [ECSDataAdapters.h](https://github.com/AliceO2Group/AliceO2/blob/dev/DataFormats/Parameters/include/DataFormatsParameters/ECSDataAdapters.h#L54). The `default` can be used 
 to ignore the run or the beam type. 
-The beam type is one of the following: `PROTON-PROTON`, `Pb-Pb`, `Pb-PROTON`, `cosmic`.
+The beam type is one of the following: `pp`, `PbPb`, `pPb`, `cosmic`.
```
knopers8, line 1313:
```suggestion
The beam type comes from the parameter `pdp_beam_type` set by ECS and can be one of the following: `pp`, `PbPb`, `pPb`, `pO`, `OO`, `NeNe`, `cosmic`, `technical`.
See `[readout-dataflow](https://github.com/AliceO2Group/ControlWorkflows/blob/master/workflows/readout-dataflow.yaml)` to verify the possible values.
```
Given that we changed the source of this, we can be more helpful in explaining where the expected values come from.

knopers8 (commented):
Thanks, it looks good. I would only propose a minor improvement to the documentation.

knopers8:
Actually, could you also grep the code of user modules for the affected string uses? I just saw one in the Felix' PR, so I imagine there might be more...

justonedev1:
I did grep over whole repo (hence the other reviewers for this PR than you and Barth). Reason why you see it in his code might be because he is using current master as a base branch where the changes are not present. I can wait until he merges and check again.

knopers8:
I meant other uses, which could be already in master. I understand that his PR is not master. In any case, I'm happy that there are no other uses, then we are good to go, I believe.

knopers8 (approved):

PR #2541, opened 2025-04-28
Allow for using only last trigger validity in postprocessing

knopers8:
Closes QC-1281.


Barthelemy (approved):
all clear thank you for updating the doc too

PR #2550, opened 2025-05-07
[QC-1286] Refactor doc

Barthelemy:


Barthelemy:
Refactored the doc by splitting it into more chapters and dispatching the content accross them. 
Also reviewed the content: removed outdated info, updated stuff.

doc/QCDB.md
```diff
@@ -0,0 +1,198 @@
...
+Each MonitorObject is stored as a TFile in the QCDB.
+It is therefore possible to easily open it with ROOT when loaded with alienv. It also seamlessly supports class schema evolution.
+
+The MonitorObjects are stored at a path which is enforced by the qc framework : `/qc/<detector code>/MO/<task name>/object/name`
+Note that the name of the object can contain slashes (`/`) in order to build a sub-tree visible in the GUI.
+The detector name and the taskname are set in the config file :
+
+```json
+"tasks": {
+  "QcTask": {       <---------- task name
```
knopers8, line 36: technically it's task ID. task name would be at:
```
  "QcTask": {       <---------- task ID
Comment
    "active": "true",
    "taskName": "QcTask", <--------- task name
    ...
```
If task name is missing, task ID is used. The split was done to allow for the same task names for multiple detectors in the same config files.

It is correct that task name is used in the QCDB paths though.

I leave it up to you if you want to go into these technicalities, but it might be worth it, given that this distinction is very much used in async (e.g. https://github.com/AliceO2Group/O2DPG/blob/71b020ddb74e7feb4f485c790921c16650f1418d/DATA/production/qc-async/its.json#L28C30-L28C32 )

doc/QCDB.md
```diff
@@ -0,0 +1,198 @@
...
+  mo->addOrUpdateMetadata(key, value);
+```
+
+## Instructions to move an object in the QCDB
+
+The script `o2-qc-repo-move-objects` lets the user move an object, and thus all the versions attached to it. E.g.:
+
+```
+python3 o2-qc-repo-move-objects --url http://ccdb-test.cern.ch:8080 --path qc/TST/MO/Bob --new-path qc/TST/MO/Bob2 --log-level 10 
+```
```
knopers8, line 70: do you feel confident advertizing this to users? If not, it could be moved to DeveloperTips.md

doc/FLPsuite.md
```diff
@@ -0,0 +1,208 @@
...
+*/10 * * * * /opt/o2/bin/o2-qc-repo-cleaner --config /etc/flp.d/ccdb-sql/repocleaner.yaml --dry-run > /dev/null 2>> /tmp/cron-errors.txt
+```
+
+1. copy the config file /etc/flp.d/ccdb-sql/repocleaner.yaml
+2. modify the config file to suit your needs
+3. run by hand the repo-cleaner to check that the config file is ok
+3. update the cron job to use the modified config file
+4. uncomment the cron job
+
+## Definition of new arguments
```
knopers8, line 117: i think this chapte rhas nothing to do with FLP suite, it seems like DPL documentation. Personally, I would remove it completely - we (developers) know how to do it or can easily find out, users don't need to do it since we have a global `o2-qc` executable. This was probably useful when users had their own QC execs (<2019 probably).

doc/ModulesDevelopment.md
```diff
@@ -3,42 +3,34 @@
...
+   * [Implementation](#implementation)
+   * [Results](#results)
+* [Quality Aggregation](#quality-aggregation)
+   * [Quick try](#quick-try)
+   * [Configuration](#configuration-1)
+   * [Implementation](#implementation-1)
+* [Naming convention](#naming-convention)
+* [Committing code](#committing-code)
+* [Data sources](#data-sources)
+   * [Readout](#readout)
```
knopers8, line 28: i guess we could remove Readout section?

knopers8 (changes requested):
Thanks, the organization is much better! I would have a few minor points, but nothing serious.

doc/QCDB.md
Barthelemy: yes, indeed. 
there was also something about `truncate` that I removed. 😂

doc/FLPsuite.md
Barthelemy: yes, moved to `DevelopersTips.md`

doc/ModulesDevelopment.md
Barthelemy: yes

knopers8 (approved):
thank you!

PR #2553, opened 2025-05-15
CTP:Change in the RawDataQcTask

lhusova:
fix for replay, confing read-out from ctx
@lietava, please have a look

### DIFF
--- a/Modules/CTP/src/RawDataQcTask.cxx
+++ b/Modules/CTP/src/RawDataQcTask.cxx
@@ -61,7 +61,6 @@ void CTPRawDataReaderTask::initialize(o2::framework::InitContext& /*ctx*/)
   getObjectsManager()->startPublishing(mHistoDecodeError.get());
 
   mDecoder.setDoLumi(1);
-  mDecoder.setDecodeInps(1);
   mDecoder.setDoDigits(1);
   for (size_t i = 0; i < nclasses; i++) {
     classNames[i] = "";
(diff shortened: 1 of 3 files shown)


Barthelemy:
@lietava @lhusova 
No worries, we will do it. 
Unless there is a need for an urgent patch at P2, I  group the reviews and make sure I have done them in view of the releases (which usually happen on Tuesdays but can be a bit more random once there is beam). 
In case you need an urgent patch, please ping us.

lietava:
@Barthelemy, thanks. We would like to be in the next release (asap in that sense). But no patch for tomorrow upgrade.

Modules/CTP/src/RawDataQcTask.cxx
```diff
@@ -93,40 +97,42 @@ void CTPRawDataReaderTask::startOfActivity(const Activity& activity)
...
-        mIndexMBclass = ctpcls[i].getIndex() + 1;
-        break;
+  if (!mReadCTPconfigInMonitorData) {
+    bool ok;
+    auto& mgr = o2::ccdb::BasicCCDBManager::instance();
+    mgr.setURL(ccdbName);
+    map<string, string> metadata; // can be empty
+    metadata["runNumber"] = run;
+    auto ctpconfigdb = mgr.getSpecific<o2::ctp::CTPConfiguration>(o2::ctp::CCDBPathCTPConfig, mTimestamp, metadata);
+    if (ctpconfigdb == nullptr) {
```
Barthelemy, line 107: Is there a reason not to use a DPL input ?

lietava: For the moment we keep both options - to read directly and to read from dpl.
There is config parameter `readCTPconfigInMonitorData` which control these options.
The dpl part starts in:[ dpl
](https://github.com/lhusova/QualityControl/blob/27140df1a78f1f16775b4577a5b06b959b190422/Modules/CTP/src/RawDataQcTask.cxx#L235)

Barthelemy (approved):

PR #2557, opened 2025-05-21
TPC QC: update json configs

ariedel-cern:
After recent updates to the Cluster Task, update associated json configs.

### DIFF
--- a/Modules/TPC/run/tpcQCRawDigits_direct.json
+++ b/Modules/TPC/run/tpcQCRawDigits_direct.json
@@ -38,7 +38,8 @@
           "mergeableOutput": "true",
           "NRawDigitsNBins": "100",  "NRawDigitsXMin": "0", "NRawDigitsXMax": "100",
           "QmaxNBins":       "200",  "QmaxXMin":       "0", "QmaxXMax":       "200",
-          "TimeBinNBins":    "1000", "TimeBinXMin":    "0", "TimeBinXMax":    "100000"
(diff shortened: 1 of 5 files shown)


Modules/TPC/run/tpcQCRawDigits_direct.json
```diff
@@ -38,7 +38,8 @@
           "mergeableOutput": "true",
           "NRawDigitsNBins": "100",  "NRawDigitsXMin": "0", "NRawDigitsXMax": "100",
           "QmaxNBins":       "200",  "QmaxXMin":       "0", "QmaxXMax":       "200",
-          "TimeBinNBins":    "1000", "TimeBinXMin":    "0", "TimeBinXMax":    "100000"
+          "TimeBinNBins":    "1000", "TimeBinXMin":    "0", "TimeBinXMax":    "100000",
+          "OccupancyNBins":    "1000", "OccupancyXMin":    "0", "OccupancyXMax":    "100000"
```
wiechula, line 42: Shouldn't occupancy be between 0 and 1?

wiechula (changes requested):

Modules/TPC/run/tpcQCRawDigits_direct.json
ariedel-cern: Yes, I updated also the files to set as max 1

wiechula (approved):

PR #2558, opened 2025-05-30
QC-1289 Possibility to use a token with BKP client

knopers8:
Token is expected as an env var to avoid exposing it in QC configuration files. Still, it's not ideal, but I don't think we have a better option at the moment.


knopers8:
Needs https://github.com/alisw/alidist/pull/5899

Barthelemy (approved):

PR #2561, opened 2025-06-05
[QC-1292] Missing run number in logs from BookkeepingSink

Barthelemy:
- added initialization of IL
- set the run number whenever we get one


knopers8 (changes requested):

knopers8:
> wip because I don't know how to test

add a valid bookkeeping URL to a QC config file and add `"exportToBookkeeping" : "true"` for at least one check.

PR #2569, opened 2025-06-19
Common: added reference run number to ReferenceComparator plots

aferrero2707:
The number of the reference run is reported both in the title of the ratio plot, and in a legend at the top of the superimposed histograms.


aferrero2707:
Here is how the legend and the title look like with the new code:
image

Barthelemy:
@aferrero2707 Sorry for the delay, I am going to review it by tomorrow

Barthelemy (approved):
All good, thanks

PR #2578, opened 2025-07-01
Common: allow access to the reference comparator plots in derived classes

aferrero2707:
The changes in the PR allow to use the ReferenceComparatorTask as a base class, providing access to the reference comparison plots for the derived classes.

A typical use case would be a post-processing task that produces some derived plots, which also need to compared with some reference.

With the current production code, this involves adding a dedicated ReferenceComparatorTask in the workflow, that fetches the plots from the post-processing task from the QCDB.

This commit allows to solve the above problem by deriving the post-processing task from the ReferenceComparatorTask, such that the reference comparison plots can be filled directly by the derived task, without going through the QCDB.

This is achieved with the following two changes:
* addition of a function that exposes the reference comparison plots in the ReferenceComparatorTask interface
* storage of the pointer to the reference histogram in the ReferenceComparatorPlot object, such that it does not need to be passed to the `update()` method

The PR also adds to the ReferenceUtils.h header file the functions that allow to access the current, reference and ratio plots from the TCanvas created by the ReferenceComparartorPlot class.

This is for example needed to allow custom checkers to access the output of the reference comparison.

### DIFF
--- a/Framework/include/QualityControl/ReferenceUtils.h
+++ b/Framework/include/QualityControl/ReferenceUtils.h
@@ -25,6 +25,9 @@
 #include "QualityControl/QcInfoLogger.h"
 #include "QualityControl/RepoPathUtils.h"
 
+#include <TH1.h>
+#include <TCanvas.h>
+
 namespace o2::quality_control::checker
 {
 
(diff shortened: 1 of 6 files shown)


Framework/include/QualityControl/ReferenceUtils.h
```diff
@@ -42,6 +45,70 @@ static std::shared_ptr<quality_control::core::MonitorObject> getReferencePlot(qu
...
+}
+
+//_________________________________________________________________________________________
+//
+// Get the ratio histogram from the container canvas
+
+static TH1* getRatioPlotFromCanvas(TCanvas* canvas)
+{
+  // Get the pad containing the current histogram, as well as the reference one in the case of 1-D plots
+  TPad* padHistRatio = (TPad*)canvas->GetPrimitive(TString::Format("%s_PadHistRatio", canvas->GetName()));
```
Barthelemy, line 97: why not a dynamic_cast ?

Framework/include/QualityControl/ReferenceUtils.h
```diff
@@ -42,6 +45,70 @@ static std::shared_ptr<quality_control::core::MonitorObject> getReferencePlot(qu
...
 
+//_________________________________________________________________________________________
+//
+// Get the current and reference histograms from the container canvas.
+// The two histograms are returned as a std::pair
+
+static std::pair<TH1*, TH1*> getPlotsFromCanvas(TCanvas* canvas, std::string& message)
+{
+  // Get the pad containing the current histogram, as well as the reference one in the case of 1-D plots
+  TPad* padHist = (TPad*)canvas->GetPrimitive(TString::Format("%s_PadHist", canvas->GetName()));
```
Barthelemy, line 56: why not a dynamic_cast ?

Barthelemy (approved):
Good idea !

I have just added a couple of trivial comments.

PR #2579, opened 2025-07-03
[QC-1082] filter empty qos and mos in datasource

justonedev1:
The problem that Andrea had was caused by wrong configuration, where he just placed empty string as a qo to datasource. I did this PR that ignores empty datasources.


Barthelemy (approved):
thanks

PR #2587, opened 2025-07-29
[EMCAL-1039]: Add checker for number of physics triggers

jokonig:
- Number of triggers per TF is a good indication to check if detector is (partially) busy
- New checker checks if last entry in trending histogram is below a settable fraction times the maximum observed number of triggers per TF
- Data quality good or bad is now indicated in the TCanvas

### DIFF
--- a/Modules/EMCAL/src/NumPhysTriggCheck.cxx
+++ b/Modules/EMCAL/src/NumPhysTriggCheck.cxx
@@ -103,7 +103,7 @@ void NumPhysTriggCheck::beautify(std::shared_ptr<MonitorObject> mo, Quality chec
       msg->Draw("same");
       Can->Update();
     } else if (checkResult == Quality::Bad) {
-      ILOG(Debug, Devel) << "Quality::Bad";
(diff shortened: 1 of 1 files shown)


Modules/EMCAL/src/NumPhysTriggCheck.cxx
```diff
@@ -0,0 +1,115 @@
...
+
+    if (checkResult == Quality::Good) {
+      //
+      msg->Clear();
+      msg->AddText("Data quality: GOOD");
+      msg->SetFillColor(kGreen);
+      msg->Draw("same");
+      Can->Update();
+    } else if (checkResult == Quality::Bad) {
+      ILOG(Debug, Devel) << "Quality::Bad";
```
knopers8, line 106:
```suggestion
      ILOG(Debug, Devel) << "Quality::Bad << ENDM";
```
without ENDM it will get attached to a different log.

jokonig: @knopers8 thanks for the catch! I have modified it as you suggested

PR #2595, opened 2025-08-08
Tune the postprocessing example config

knopers8:
- "10 minutes" stop trigger is removed, as I saw cases where it was preserved it in production configurations.
- better SliceTrendingTask multigraph drawing option is used, which will allow to see different colours for trended slices
- "resumeTrend" in SliceTrendingTask is switched off to avoid confusion about where the old points come from and to avoid picking up differently configured trends, which may cause incompatibilities


Barthelemy (approved):
nice, thank you !

PR #2598, opened 2025-08-18
FIT: graph helper

mvishiu11:
Added a graph helper, which allows user to easily register TGraphErrors in a similar manner to HistHelper.


andreasmolander:
@knopers8 Ok cool, I was not sure about the policy

PR #2600, opened 2025-08-19
Common: customize ratio plot draw option in ReferenceComparator

aferrero2707:
By default the histograms are drawn with error bars ("E" option), however there are cases where the user might want to hide the errors and draw the histograms with the "H" option, for example because the error bars are unreliable or unphysical.

This can already be done by passing a custom draw option via the task configuration interface.
However, the customized draw option is only passed to the current and reference histograms, but not to the ratio plot, which is always drawn with the "E" option.
The commit extends the custom draw option also to the ratio histogram.

Here is a practical example:
- ratio plot with default "E" option
image

- ratio plot with "H" option
image


aferrero2707:
@Barthelemy @knopers8 @justonedev1 here is a small customization improvement for the reference comparison plots.

Barthelemy (approved):
thanks !

PR #2612, opened 2025-10-17
MUON: process only best matching candidates

aferrero2707:
The MFT-MCH matching task now stores by default multiple matching candidates for each MCH standalone track.
In order to properly assess the tracking quality and avoid a large combinatorial background, only the best matching candidate for each MCH track must be processed by the QC tracks task.


knopers8 (approved):
Assuming I understood correctly the lifetimes of objects and pointers in the new snippet, it seems safe.

aferrero2707:
> Assuming I understood correctly the lifetimes of objects and pointers in the new snippet, it seems safe.

The map structure that temporarily keeps the indexes of the matching candidates is local to the code block that I have added, and only used to collect and sort the candidates. The rest reuses structures that were already used before.

PR #2648, opened 2026-03-06
User standard user I/O spec helpers across the framework

knopers8:
The first commit extends the helpers to allow for custom subSpec and binding, the second adopts the helpers in the framework.

I tried running o2-qc-run-basic + example postprocessing, as well as starting the production TPC remote workflow - no regressions, it seems.


Barthelemy (approved):
Very good, thank you!
That makes it a lot more readable.

PR #2659, opened 2026-03-31
[CTP] use list of specific inputs instead of list of first inputs

musinsky:


lietava:
@Barthelemy ok on my side. Please, merge.

knopers8 (approved):

PR #2660, opened 2026-04-02
[QC-1346] Scaffolding to share common code among QC data processors

knopers8:
Actor is a template base class for all QC Data Processors. It is supposed to bring their commonalities together, such as: service initialization, Data Processing Layer adoption, retrieving configuration and runtime parameters, interactions with controlling entities (DPL driver, AliECS, ODC).

The design is based on CRTP (see the web for explanation), which allows us to:
- avoid code repetition in implementing aforementioned commonalities
- optionally perform certain actions depending on traits of an Actor specialization. CRTP, in contrast to dynamic inheritance, is also advertised to avoid performance impact due to vtable lookups. It is certainly a nice bonus in our case, but it was not the main motivation for CRTP-based approach.

To allow for compile-time customization of centralized Actor features, we require each concrete Actor to implement an ActorTraits structure with certain parameters which is enforced with ValidActorTraits concept. The traits are separated from the main Actor class to improve readability and allow for shorter compilation times by allowing many helper functions avoid including Actor.h and a corresponding actor specialization. For additional savings on compilation time and clutter in code, we validate ActorTraits specializations with a concept only in Actor, but this could be revisited if proven wrong.

This commit paves the path for refactoring existing QC data processors as Actor specializations/children.

### DIFF
--- a/Framework/include/QualityControl/Actor.h
+++ b/Framework/include/QualityControl/Actor.h
@@ -123,12 +123,7 @@ class Actor
     return actor_helpers::requiresService<ConcreteActor, S>();
   }
 
-  // a trick to prevent bugs like "class TaskRunner : public Actor<CheckRunner>"
-  // by keeping the constructor private, thus allowing only CheckRunner to initialize Actor<CheckRunner>
-  // see https://www.fluentcpp.com/2017/05/12/curiously-recurring-template-pattern/
-  friend ConcreteActor;
-
- private:
+ public:
   explicit Actor(const ServicesConfig& servicesConfig)
     : mServicesConfig{ servicesConfig },
       mActivity{ servicesConfig.activity }
--- a/Framework/include/QualityControl/Actor.h
+++ b/Framework/include/QualityControl/Actor.h
@@ -130,7 +130,6 @@ class Actor
 
  private:
   explicit Actor(const ServicesConfig& servicesConfig)
-    requires(std::derived_from<ConcreteActor, Actor<ConcreteActor>>)
     : mServicesConfig{ servicesConfig },
       mActivity{ servicesConfig.activity }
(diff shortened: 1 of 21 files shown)


Framework/include/QualityControl/Actor.h
```diff
@@ -0,0 +1,369 @@
...
+    });
+  }
+
+ protected:
+  // mandatory methods to be implemented by concrete actor
+  void onInit(framework::InitContext&) = delete;
+  void onProcess(framework::ProcessingContext&) = delete;
+
+  // mandatory methods to be implemented by concrete actor if specific features are enabled
+  bool isCritical() const
```
Barthelemy, line 164: the formatting is a tad weird here.

Framework/include/QualityControl/ServicesConfig.h
```diff
@@ -0,0 +1,29 @@
+//
+// Created by pkonopka on 04/12/2025.
+//
```
Barthelemy, line 7: File header is missing

Barthelemy (approved):
Sorry for the delay. I had started and then forgot to finish. Thus I had to restart as I did not remember!

It is very good. It is well documented and tested and despite the change of paradigm I could understand what is done and how. I learned a lot along the way. 
Thanks

Framework/include/QualityControl/Actor.h
knopers8: Yes i agree, but it's clang-format forcing it. I can't make it stay in one line unless I disable it.

Framework/include/QualityControl/ServicesConfig.h
knopers8: fixed

PR #2663, opened 2026-04-08
Documentation inconsistency fixes

knopers8:
Out of curiosity, I asked an LLM agent to review the documentation against the code and point out inconsistencies.

It also rightfully complained about lack of clarity when to use "dataSource" and "dataSources", but I created a separate ticket for it, since it's a bigger effort to sanitize this. There were some very minor issues which I wouldn't bother, which I skipped.


Barthelemy (approved):
excellent idea!

PR #2664, opened 2026-04-08
[CTP] fix quality result for empty histogram

musinsky:


Modules/CTP/src/RawDataReaderCheck.cxx
```diff
@@ -107,6 +107,9 @@ Quality RawDataReaderCheck::check(std::map<std::string, std::shared_ptr<MonitorO
         }
       }
       result.set(setQualityResult(mVecBadBC, mVecMediumBC));
+      if (h->GetEntries() == 0) {
```
knopers8, line 110: FYI, the fix will only affect `bcMinBias1` and `bcMinBias2`, I am not sure if that is exactly the intention

knopers8 (approved):

PR #2665, opened 2026-04-10
[QC-1088] Late Tasks as post-processing within the message passing framework

knopers8:
Late Tasks can create new MonitorObjects based on existing MOs and QOs in the QC workflow.
I called them "late", because they would be typically executed close to the end of the processing chain.
When running QC workflows distributed over multiple nodes, with Mergers combining the results, they can be used to process the merged results.
The adequate use cases involve:
* creating trends and correlations from other MonitorObjects and QualityObjects (trending a histogram average, trending quality)
* creating plots which can only be constructed from merged objects (ratios, visualizations, ...)
* creating summary canvases of QualityObjects available in the QC workflow

They are meant to slowly replace Post-Processing, at least in the use cases mentioned above.

Needs https://github.com/AliceO2Group/QualityControl/pull/2660 to have [WIP] removed.


doc/Configuration.md
```diff
@@ -54,7 +58,7 @@ This is the global structure of the configuration in QC.
 }
 ```
 
-There are six QC-related components:
+There are following QC-related components:
```
Barthelemy:
```suggestion
The following QC-related components are available:
```

doc/ModulesDevelopment.md
```diff
@@ -37,9 +41,10 @@ Before developing a module, one should have a bare idea of what the QualityContr
 
 ![alt text](images/Architecture.png)
```
Barthelemy: I have to update that and upload it after this is merged.

Framework/src/LateTaskRunner.cxx
```diff
@@ -0,0 +1,175 @@
...
+          } else {
+            delete mo;
+          }
+        } else {
+          delete obj;
+        }
+      }
+    } else if (matchingDataSource->isOneOf(DataSourceType::Check, DataSourceType::Aggregator)) {
+      auto qo = DataRefUtils::as<QualityObject>(ref);
+      auto key = qo->getName();
```
Barthelemy: getCheckName()

Modules/Skeleton/src/SkeletonLateTask.cxx
```diff
@@ -0,0 +1,119 @@
...
+namespace o2::quality_control_modules::skeleton
+{
+
+SkeletonLateTask::~SkeletonLateTask()
+{
+}
+
+void SkeletonLateTask::initialize(o2::framework::InitContext& /*ctx*/)
+{
+  // THUS FUNCTION BODY IS AN EXAMPLE. PLEASE REMOVE EVERYTHING YOU DO NOT NEED.
```
Barthelemy: THIS

Framework/src/LateTaskRunner.cxx
```diff
@@ -0,0 +1,175 @@
...
+      auto moc = DataRefUtils::as<MonitorObjectCollection>(ref);
+      // makes MOs own their TObjects and MOC own MO (the latter we cancel in the next step)
+      moc->postDeserialization();
+      moc->SetOwner(false);
+
+      for (const auto& obj : *moc) {
+        auto mo = dynamic_cast<MonitorObject*>(obj);
+        if (mo != nullptr) {
+          if (matchingDataSource->subInputs.empty() || std::ranges::find(matchingDataSource->subInputs, mo->getName()) != matchingDataSource->subInputs.end()) {
+            taskInputs.insert(mo->getName(), std::shared_ptr<MonitorObject>(mo));
```
Barthelemy: I don't remember if we do that elsewhere, but using the MO name is a bit risky in case there are several tasks publishing an object with the same name (given that we have multiple sources). Or are we sure that we receive only data from one given task ?

Framework/src/InfrastructureSpecReader.cxx
```diff
@@ -417,6 +435,48 @@ PostProcessingTaskSpec
...
+  LateTaskSpec lts;
+
+  lts.taskName = lateTaskTree.get<std::string>("taskName", lateTaskId);
+  lts.active = lateTaskTree.get<bool>("active", lts.active);
+  lts.critical = lateTaskTree.get<bool>("critical", lts.critical);
+  lts.className = lateTaskTree.get<std::string>("className");
+  lts.moduleName = lateTaskTree.get<std::string>("moduleName");
+  lts.detectorName = lateTaskTree.get<std::string>("detectorName", lts.detectorName);
+
+  lts.outputActivityStrategy = outputActivityStrategyFromString.at(lateTaskTree.get<std::string>("outputActivityStrategy", "integrated"));
```
Barthelemy: I guess that we do it like that everywhere, but basically here we will just fail if we put "integration" instead of "integrated". Perhaps at some point we should better handle wrong keywords from the config.

Framework/src/LateTaskRunner.cxx
```diff
@@ -0,0 +1,175 @@
...
+  mTask->startOfActivity(activity);
+  mObjectsManager->setActivity(activity);
+  mObjectActivity = std::nullopt;
+}
+
+void LateTaskRunner::onProcess(ProcessingContext& pCtx)
+{
+  const QCInputs& taskInputs = extractInputs(pCtx);
+  if (taskInputs.size() == 0) {
+    ILOG(Warning, Support) << "Could not prepare inputs for task, will not call process() and publish objects. "
```
Barthelemy: in which case can this happen ?

Framework/include/QualityControl/DataProcessorAdapter.h
```diff
@@ -114,7 +114,7 @@ struct DataProcessorAdapter {
   static std::string dataProcessorName(std::string_view userCodeName, std::string_view detectorName)
   {
     using traits = ActorTraits<ConcreteActor>;
-    return dataProcessorName(detectorName, userCodeName, traits::sActorTypeKebabCase);
+    return dataProcessorName(userCodeName, detectorName, traits::sActorTypeKebabCase);
```
Barthelemy: was there a bug here ?

Framework/include/QualityControl/LateTaskSpec.h
```diff
@@ -0,0 +1,53 @@
...
+  std::vector<DataSourceSpec> dataSources;
+
+  // advanced
+  bool active = true;
+  bool critical = true;
+  OutputActivityStrategy outputActivityStrategy = OutputActivityStrategy::Integrated;
+  CustomParameters customParameters;
+
+  // reco
+  // GRPGeomRequestSpec grpGeomRequestSpec;
```
Barthelemy: leftover ?

Barthelemy (commented):
First of all, let me apologize for this very late review. 

This is a great job! Thank you.
The late tasks have clearly the potential to replace the PostProcessing tasks in many cases. 

I have put a few questions and comments that I let you address.

PR #2670, opened 2026-05-20
FIT: Include FT0/FV0 total charge histograms for Pb-Pb Galuber fits

sahilupadhyaya92:
This PR updates the FT0 and FV0 DigitQcTasks.
There are new histograms added that show the total integrated charge for FT0A, FT0C, FT0A+C and FV0
with selection of Colliding BCs with FT0 Vertex and FV0 OrA triggers.

### DIFF
--- a/Modules/FIT/FV0/src/DigitQcTask.cxx
+++ b/Modules/FIT/FV0/src/DigitQcTask.cxx
@@ -26,6 +26,9 @@
 #include "DataFormatsFV0/LookUpTable.h"
 #include "Common/Utils.h"
 
+#include "DataFormatsParameters/GRPLHCIFData.h"
+#include "CCDB/BasicCCDBManager.h"
+
 #include "FITCommon/HelperHist.h"
 #include "FITCommon/HelperCommon.h"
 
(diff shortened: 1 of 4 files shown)


Modules/FIT/FV0/src/DigitQcTask.cxx
```diff
@@ -433,11 +445,54 @@ void DigitQcTask::startOfCycle()
   mTimeSum = 0.;
 }
 
+void DigitQcTask::loadBcPatternIfNeeded()
```
Barthelemy, line 448: do you really need to duplicate code like that ? 
It should go to a common class and be shared between FT0 and FV0.

Barthelemy (changes requested):
Hi, 

Please fix the formatting, we cannot merge otherwise. I don't know if you have time to do it now, otherwise later, but this code duplication between FV0 and FT0 should be factored out.

Modules/FIT/FV0/src/DigitQcTask.cxx
sahilupadhyaya92: Hi @Barthelemy 
I fully agree, but this would require a larger refactor and this is just a very quick modification to add histograms for the Glauber fits.
The refactoring will be definitely done later.
I hope it is ok for temporary solution ?

Barthelemy: Hi, 
Ok, but please take note of it, perhaps create a JIRA ticket, to remember it.

Barthelemy (approved):

Modules/FIT/FV0/src/DigitQcTask.cxx
sahilupadhyaya92: Thanks @Barthelemy 
There is already a JIRA for that: https://its.cern.ch/jira/browse/AFIT-41
I will update the ticket so that we keep track of this.

Barthelemy: ok, thanks. 

Please also fix the formatting in this PR.

sahilupadhyaya92: Formatting is fixed now.
Thanks.

Barthelemy:
@andreasmolander we need your approval here too

andreasmolander (approved):
Fine if we implement the discussed fixes later

PR #2671, opened 2026-06-30
Use PP task name instead of PP task ID when requesting their outputs 

knopers8:
When requesting PP task output as a data source, it's ID was used for DataDescription. Meanwhile, PostProcessingDevice::getOutputSpecs uses the task name, which is the general convention in QC. The commit fixes the data source to use task name.

I tested the change with the big TPC remote workflow, it works. It seems nothing was relying on the wrong behaviour.

Fixes QC-1336.


Barthelemy (approved):
Thank you

Barthelemy:
I don't know why the WIP check is stuck merging

PR #2675, opened 2026-07-08
Fix undefined symbol on macOS ARM

vkucera:


Barthelemy (approved):
thanks, I noticed it the other day !

vkucera:
Thanks @Barthelemy for merging. I think for the CI we need a new release.

PR #2677, opened 2026-07-20
ITS: plots to check dead dcols or regions at SOR and EOR

iravasen:
Added the following plots: 

- Layers 0,1,2 (Inner Barrel): chipID vs dcols ID at SOR and EOR (2 plots in total) 
- Layers 3,4,5,6 (Outer Barrel): chipID vs region ID at SOR and EOR (8 plots in total) 

In particular, 

- plots at SOR are filled for the first `nCycleStop` (configurable) QC cycles and then never updated again. 
- plots at EOR are filled for `nResetCycles` and then reset after. In this way they contain the picture of the last N (with N<=`nResetCycles`) QC cycles. 

Objective: 
check SEU effects on dcols (IB) and regions (IB and OB).

### DIFF
--- a/Modules/ITS/include/ITS/ITSClusterTask.h
+++ b/Modules/ITS/include/ITS/ITSClusterTask.h
@@ -74,11 +74,11 @@ class ITSClusterTask : public TaskInterface
 
   static constexpr int NLayer = 7;
   static constexpr int NLayerIB = 3;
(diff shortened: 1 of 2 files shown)


AlessandroSturniolo (approved):

Barthelemy:
approved, merging. It seems stuck.

PR #2684, opened 2026-08-21
GLO VTX: add missing protection against accessing MC labels of background events

ehellbar:


ehellbar:
@sawenzel @shahor02 adding the missing protection against trying to access MC labels of background events

ehellbar:
like this, we will fill the data vertex plots, but not the MC plots.

knopers8 (approved):
thank you

knopers8:
For the record, this fixes [O2-7132](https://its.cern.ch/jira/browse/O2-7132)

Modules/GLO/src/VertexingQcTask.cxx
```diff
@@ -215,7 +215,7 @@ void VertexingQcTask::monitorData(o2::framework::ProcessingContext& ctx)
     mTimeUncVsNContrib->Fill(nContr, timeUnc);
     mBeamSpot->Fill(x, y);
 
-    if (mUseMC && mcLbl[i].isSet()) { // make sure the label was set
+    if (mUseMC && mcLbl[i].isSet() && mcLbl[i].getSourceID() == 0) { // make sure the label was set and to use only the underlying event
```
sawenzel, line 218: I have some doubts about this: In an O2DPG simulation when using embedding, we will have both SourceID == 0 (the background) and SourceID == 1  (the signal) ... and both might contribute to the vertexing task.

We could however use mcLbl.isNoise() or mcLbl.isValid() or something. ... or directly exclude QED by sourceID != 99.

ehellbar: we do the same check also twice earlier in the function, when we loop over the labels. So for simulations with embedding, this never worked properly, then, always skipping signal and only filling background?

ehellbar: or, is background == simulated events and signal == embedded signal in this case? Then we might not want the embedded signal in those histogram?

shahor02: This particular task extracts the MCHeader multiplicity as a very crude alias of the occupancy, relevant only for the PbPb collision. Assuming that the underlying PbPb event is always source 0, a few tracks coming from the overlaid signal with source>0 should be irrelevant.

Barthelemy (approved):

Modules/GLO/src/VertexingQcTask.cxx
sawenzel: If it's only used for PbPb then it's probably fine

sawenzel:
I think this can be merged and a new tag published.

PR #2685, opened 2026-08-27
Do not hardcode location of test CCDB

ktf:
The new CI setup has a test instance bound to a local port of the
worker's container.


ktf:
@Barthelemy I need a new release with this in, please.

Barthelemy (approved):

PR #2689, opened 2026-09-14
Fix warnings with new Clang

ktf:


ktf:
@Barthelemy we need a new release with this, I guess...

wiechula (approved):

Barthelemy (approved):

PR #2690, opened 2026-09-15
Properly handle object lifetime

ktf:
Taking ownership of DPL provided objects is wrong.


Barthelemy:
I'll let @knopers8 review it
Then I will make a new minor version.

knopers8 (approved):
Out of curiosity, did it cause trouble or are you just fixing a warning? It looks like the pointer is never accessed outside of this scope.

And speaking of stealing the framework's belongings, we also do this ;) https://github.com/AliceO2Group/QualityControl/blob/3eec58e633fb22e128efb0a5bf41f441d26c60a2/Framework/src/TaskRunner.cxx#L95

vkucera:
@Barthelemy Can you please make a new release so that we can include the latest fixes in the builds?

